PR #88, opened 2019-12-07
*: extract storage package, and update kvproto

### LINKED ISSUE #71: The whole argument of --storage is stored in backupmeta which may include sensitive data
According to the current implementation, the argument for --storage will be stored in backup meta and will be retrieved and sent to TiKV during restore.
In this case, for backup & restore with cloud storage, e.g. AWS S3, keys will also be included in the backup meta, which is insecure and should be avoided.

- Is it necessary to store the argument in the backup meta? The --storage option is required in each backup, checksum and restore command.
- It's also insecure to specify the keys in the argument of --storage, we may set the keys or other sensitive data in environment variables and retrieve them and send them to TiKV in BR.

@kennytm @overvenus @tennix What do you think?

kennytm:
1. Update kvproto which replaces the storage "path" field by a structured "StorageBackend" type.
2. Replaces all use of such "path" by "StorageBackend"
3. Extracted the storage module from `pkg/util` into `pkg/storage`. In the future I'd like to further move them into `pingcap/tidb-tools` so it could be reused by Lightning and Dumpling.
4. Added some simple S3 awareness (not compatible with #66).

Closes #71.


pkg/storage/s3.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+const (
+	s3EndpointOption = "s3.endpoint"
+)
+
+// S3EndpointOption is a BackendOption for changing the endpoint of the storage.
+type S3Endpoint struct {
+	backendOption
+	// Endpoint is the host name of the S3 endpoint.
```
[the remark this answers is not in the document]
kennytm: Yes. But Go's AWS SDK accepts just `host:port` too (and they use another flag to determine whether to use HTTPS).

pkg/storage/s3.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/pingcap/errors"
+	"github.com/pingcap/kvproto/pkg/backup"
+	"github.com/spf13/pflag"
+)
+
+const (
+	s3EndpointOption = "s3.endpoint"
+)
+
+// S3EndpointOption is a BackendOption for changing the endpoint of the storage.
```
DanielZhangQD: This is a little complicated, for each option, we have to define one struct and two functions, suggest making it easier to build options.

pkg/storage/parse.go
```diff
@@ -7,32 +7,30 @@ import (
...
+	if !options.S3.isEmpty() && storageScheme != "s3" {
+		return optionsNotApplicableToStorage("s3")
+	}
+	if err := options.S3.validate(); err != nil {
+		return err
+	}
 
-func (backendOption) applyOnS3(_ *backup.Noop, name string) error {
-	return errors.Errorf("option '%s' is not applicable to s3 storage", name)
+	return nil
```
DanielZhangQD: This is not straightforward, can we make it like below:
```
switch storageScheme{
case s3:
  if options.S3.isEmpty(){
    return error
  }
  validate()
```
If the above is accepted, we may just migrate the logic to the blocks from L48.

pkg/storage/s3.go
```diff
@@ -10,39 +10,40 @@ const (
...
-	s3.Endpoint = o.Endpoint
+func (options *S3BackendOptions) validate() error {
 	return nil
 }
 
-// OptionName returns the name of this option.
-func (o *S3Endpoint) OptionName() string {
-	return s3EndpointOption
+func (options *S3BackendOptions) isEmpty() bool {
+	return options.Endpoint == ""
```
DanielZhangQD: Can we combine these two functions?
If we have 2 options for S3, both of them have default values, how can we determine if it's empty?

pkg/storage/parse.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if !options.S3.isEmpty() && storageScheme != "s3" {
+		return optionsNotApplicableToStorage("s3")
+	}
+	if err := options.S3.validate(); err != nil {
+		return err
+	}
+
+	return nil
```
DanielZhangQD: This is not straightforward, if scheme=="s3", we only need to care about if options.S3 is empty, because even if both options.S3 and options.GCS are not empty, we can still work as expected. 
Can we make it like below:
```
switch storageScheme{
case s3:
  if options.S3.isEmpty(){
    return error
  }
  validate()
```
If the above is accepted, we may just migrate the logic to the blocks from L48.

pkg/storage/s3.go
```diff
@@ -0,0 +1,49 @@
...
+	Endpoint string `json:"endpoint" toml:"endpoint"`
+}
+
+func (options *S3BackendOptions) validate() error {
+	return nil
+}
+
+func (options *S3BackendOptions) isEmpty() bool {
+	return options.Endpoint == ""
+}
```
DanielZhangQD: Can we combine these two functions?
If we have 2 options for S3, both of them have default values, how can we determine if it's empty?

pkg/storage/parse.go
kennytm: The point of the check is that if you have `--storage local://...` and provide an --s3 option it is invalid. So this doesn't matter?

DanielZhangQD: It does matter, I mean we should check the options by scheme instead of checking the scheme by the options, for your example, if --storage local://... and provide both --s3 options and the local options (if some), BR can work with local:// and the local options, right? No need to return an error just because users also provide s3 options.
It should be more clear with this case:
--storage s3://...
Users provide both s3 options and gcs options
I think no need to return an error because BR can just use the s3 options to work.

kennytm: With this argument we should allow `--storage local://… --s3.endpoint … --local.stuff …`, but since local has no options in reality we should just allow `--storage local://… --s3.endpoint …` 

The missing s3 options could cause validation failure if we require either --s3.region or --s3.endpoint must be present with an s3 storage. This is a different issue from using X options on Y storage.

DanielZhangQD: See the first [comment](https://github.com/pingcap/br/pull/88#discussion_r355420437), we only need to check s3 options when the scheme is s3.

pkg/storage/s3.go
kennytm: Removed these two functions.

overvenus:
Linter reports some suggestions, could you address them?

PR #94, opened 2019-12-09
tests: test restore db

overvenus:
Test restore db.


tests/br_db/run.sh
```diff
@@ -0,0 +1,49 @@
...
+# TODO: use backup db, once it is supported.
+# backup full
+echo "backup start..."
+run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB" --ratelimit 5 --concurrency 4 --fastchecksum true
+
+run_sql "DROP DATABASE $DB;"
+
+# restore db
+echo "restore start..."
+run_br restore db --db $DB -s "local://$TEST_DIR/$DB" --pd $PD_ADDR
```
kennytm: it should check whether the restored data is correct?

overvenus: Checksum will check data integrity before finishing restore.

kennytm: Okay.

tests/br_key_locked/workload
```diff
@@ -1,4 +1,4 @@
-recordcount=10000
+recordcount=1000
```
kennytm: Is this change intended?

overvenus: It speeds up br_key_locked test, resolving lock is very slow.

Issue #87, opened 2019-12-06
Online Restore Project


overvenus:
# BR support online restore

## Overview

Currently, BR only supports offline restore, but some users need online restore.
The goal of this project: to support online restore, during which the online
query should not have a significant impact.

## Problem Statement

BR will adjust TiKV to import mode during restore. In this mode, TiKV can
quickly import a large amount of data through RocksDB IngestSST, but this mode
will affect the online query seriously.

During restore, BR will split a large number of small regions and randomly
scatter them to each TiKV. This will affect PD scheduling decisions, which will
have an unpredictable impact on the online query.

After restore is complete, the entire cluster executes full Compaction,
which will affect the IO/CPU greatly.

Before the restore is completed, the restore data is inconsistent, but there is
no restriction to prevent TiDB from operating on the data.

## Proposed Solution

Online import scenarios can be roughly divided into two categories:
1. No new TiKV nodes join the cluster, and restore is performed on the old nodes.
2. A new TiKV node joins the cluster and performs restore on the new node.

For case 1, we can:
- Skip setting import mode and control the impact of import on online query
  through flow control. Specific strategies:
    - Flow control Upload/Download.
    - Flow Control Ingest SST.
- After restore is complete, skip Compaction. In addition to optimizing
  RocksDB data structure, Compaction also writes SST table properties,
  so to skip Compaction, TiKV needs:
    - Use TiKV table properties to process SST files during Upload/Download.

For case 2, we can:
- Let the scatter region support scatter regions on specified TiKV nodes.
- Let PD better handle scheduling in online restore scenarios.
- Follow the offline restore process on TiKV.

In both cases, TiDB needs to support making the restored database/table
invisible to the user.

## Success Criteria

- Achieve online restore.
- No service unavailability during online restore.
- Write stall cannot occur during online restore.
- The user cannot operate the recovered database/table during online restore.

## Difficulty

* Medium

## Score

* 4000

## TODO list

- [x] TiKV skips setting import mode during restore. (score: 200)
- [x] TiKV can flow control Upload/Download SST during restore. (score: 200)
- [x] TiKV can flow control ingest SST during restore. (score: 200)
- [x] Upload/Download handle table properties needed for SST to write to TiKV. (score: 300)
- [x] PD support scatter regions on specified TiKV nodes. (score: 800)
- [ ] PD scheduling should not be affected by the restore scenario. (score: 800)
- [ ] TiDB supports hiding the specified database/table. (score: 1000)
[...]

PR #100, opened 2019-12-10
restore: add rateLimit for flow control

lhy1024:
To make that TiKV can flow control Download SST during restore.It is part of #87..
- [x] add test


pkg/utils/storage.go
```diff
@@ -0,0 +1,10 @@
+package utils
```
overvenus: Could you change the file name to `unit.go`?

pkg/restore/import.go
```diff
@@ -67,11 +68,18 @@ func (ic *importClient) DownloadSST(
 	ctx context.Context,
 	storeID uint64,
 	req *import_sstpb.DownloadRequest,
+	rateLimit uint64,
 ) (*import_sstpb.DownloadResponse, error) {
 	client, err := ic.getImportClient(ctx, storeID)
 	if err != nil {
 		return nil, err
 	}
+	_, err = client.SetDownloadSpeedLimit(ctx, &import_sstpb.SetDownloadSpeedLimitRequest{
```
overvenus: I think it only needs to be set once.

cmd/restore.go
```diff
@@ -44,6 +44,8 @@ func NewRestoreCommand() *cobra.Command {
 
 	command.PersistentFlags().Uint("concurrency", 128,
 		"The size of thread pool that execute the restore task")
+	command.PersistentFlags().Uint64("rateLimit", 128,
```
overvenus:
```suggestion
	command.PersistentFlags().Uint64("ratelimit", 1024,
```

Please improve code coverage by choosing a test and adding a ratelimt flag,

lhy1024: how about setting default value to 0 without ratelimit flag?

overvenus:
PTAL @5kbpers @kennytm

lhy1024:
PTAL @5kbpers @kennytm

cmd/restore.go
```diff
@@ -44,6 +44,8 @@ func NewRestoreCommand() *cobra.Command {
 
 	command.PersistentFlags().Uint("concurrency", 128,
 		"The size of thread pool that execute the restore task")
+	command.PersistentFlags().Uint64("rateLimit", 0,
```
5kbpers: Prefer `rate-limit` or `ratelimit`

lhy1024: done

pkg/restore/import.go
```diff
@@ -159,10 +180,22 @@ func (importer *FileImporter) Import(file *backup.File, rewriteRules *restore_ut
...
+		stores := make(map[uint64]struct{})
 		// Try to download and ingest the file in every region
 		for _, regionInfo := range regionInfos {
 			var downloadMeta *import_sstpb.SSTMeta
 			info := regionInfo
+			// set store speed limit
+			for _, peer := range info.Region.Peers {
+				storeID := peer.StoreId
+				if _, ok := stores[storeID]; !ok {
+					err := importer.setDownloadSpeedLimit(storeID, importer.rateLimit)
```
kennytm: Well this looks too complicated.

The current implementation will call `SetDownloadSpeedLimit` for the associated store *every time* a backup file is processed. So if there are N files this API would be called N times. While the expected design is just call this once per store at start up.

The current code can handle a new store appearing while restoring... but I'm not sure if such scenario is worth supporting.

lhy1024: yes, it looks a little complicated. because I have no idea how to get stores before impoting files.I have moved `stores` outside to avoid call N times.

kennytm: @lhy1024 

> because I have no idea how to get stores before impoting files.

You call `GetAllStores()` from a `pd.Client`.

lhy1024: thanks

pkg/restore/client.go
```diff
@@ -216,6 +223,23 @@ func (rc *Client) CreateTables(
 	return rewriteRules, newTables, nil
 }
 
+func (rc *Client) setSpeedLimit() error {
+	if !rc.hasSpeedLimited {
```
overvenus:
```suggestion
	if !rc.hasSpeedLimited && rc.rateLimit != 0 {
```

Should check whether it needs a ratelimit.

lhy1024: In fact even set 0 is considered unlimited。

Issue #89, opened 2019-12-08
Backup to common cloud storage


tennix:
# BR support common cloud storage

## Overview

Integrate BR with common cloud object storage (S3, GCS, Azure Blob storage etc).

## Problem statement

Currently, BR supports local storage where backup files are stored on local directory. But the backup files need to be collected together and copied to every TiKV node. This is difficult to use in practice, so it's better to mount an NFS like filesystem to every TiKV node and BR node. However, mounting NFS to every node is difficult to set up and error-prone.

Alternatively, object storage is better for this scenario, especially that it's quite common to backup to S3/GCS on public cloud. 

### S3 Support (2400 points)
- [x] Extend current backup protobuf to support S3 compatible (S3, Ceph object storage, Minio, Aliyun OSS) storage *(900 points / medium)* @tennix https://github.com/pingcap/kvproto/pull/507
- [x] Store BR metadata to S3 compatible object storage *(600 points / easy)* @DanielZhangQD https://github.com/pingcap/br/pull/103 
- [x] Store TiKV SST files to S3 compatible object storage *(900 points / medium)* @yiwu-arbug https://github.com/tikv/tikv/pull/6209
### GCS Support (1800 points)
- [x] Extend backup protobuf to support GCS *(300 points / easy)* @SunRunAway https://github.com/pingcap/kvproto/pull/514
- [x] Store BR metadata to GCS *(600 points / easy)* @SunRunAway https://github.com/pingcap/br/pull/102
- [x] Store TiKV SST files to GCS *(900 points / medium)* @yiwu-arbug https://github.com/tikv/tikv/pull/6283
### TiDB Operator integration (900 points) 
- [x] Integrate BR with tidb-operator to allow easy deployment (900 points / medium) @DanielZhangQD https://github.com/pingcap/tidb-operator/pull/1280

### Test (2100 points)
- [ ] Unit test coverage >= 60% *(900 points / easy)*
- [ ] e2e test *(1200 points / medium)* https://github.com/pingcap/br/pull/108

## Mentors

* @tennix 
* @overvenus 
* @kennytm 

## Recommended skills
* Go language
* Rust language
* Familiar with S3/GCS/Azure blob storage

SunRunAway:
@francis0407 and I also have interest in this issue.

gregwebs:
Another task that can be added to this is to ensure proper streaming to object storage without copying data in memory.

tennix:
[...]

PR #102, opened 2019-12-11
storage: support GCS backup and restore

SunRunAway:
part of #89 

### GCS Support
store BR metadata to GCS


pkg/storage/gcs.go
```diff
@@ -0,0 +1,140 @@
...
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func newGCSStorage(gcs *backup.GCS) (*gcsStorage, error) {
+	var clientOps []option.ClientOption
+	clientOps = append(clientOps, option.WithCredentialsJSON([]byte(gcs.GetCredentialsBlob())))
```
gregwebs: It looks like we are being forced into one particular way of doing credentials here. I am actually not sure what this is doing. But on GCE or GKE with the workload identity feature enabled the SDK will already automatically find credentials in the metadata.

pkg/storage/gcs.go
```diff
@@ -0,0 +1,140 @@
...
+	wc.PredefinedACL = s.gcs.PredefinedAcl
+	_, err := wc.Write(data)
+	if err != nil {
+		return err
+	}
+	return wc.Close()
+}
+
+// Read storage file
+func (s *gcsStorage) Read(ctx context.Context, name string) ([]byte, error) {
```
gregwebs: This reads the entire file into memory. However, at least some use cases (for example the checksum command) can stream data, avoiding the need to allocate a large amount of memory and finishing more quickly by starting the checksum calculation as soon as data is available. With reduced memory usage it is safer to parallelize the operation.

For this I think you can just add a function `Reader(Context, string) (io.Reader, error)` to the storage interface. Then the checksum command can use `io.Copy` to stream data through the checksum in blocks. It may be preferable to now have a `Read` function and instead return the Reader and size: this way the caller can read into memory if they prefer.

It is possible that the streaming approach could end up being slower, so it is definitely still worth benchmarking.

pkg/storage/gcs.go
SunRunAway: Since `br` is not sure that all the TiKV instances are deployed on GCE or GKE either, the simplest way is getting a credentials JSON and sending it to other TiKV instances.

Yes, I think it's a cool thing to use service account from environment on GCE or GKE. Maybe in the future and we have a idea to handle the credentials for TiKV instances.

pkg/storage/gcs.go
SunRunAway: The interface is definded using `[]byte` alreadly, see https://github.com/pingcap/br/blob/master/pkg/storage/storage.go#L9-L16
We may change the interface in another PR.

Another consideration is that `br` only backups metadata, which need to be hold in memory completly. So I think both interface in this situation is reasonable.

go.mod
```diff
@@ -3,20 +3,20 @@ module github.com/pingcap/br
...
 	github.com/gogo/protobuf v1.3.1
-	github.com/golang/groupcache v0.0.0-20190702054246-869f871628b6 // indirect
 	github.com/google/btree v1.0.0
 	github.com/google/uuid v1.1.1
 	github.com/onsi/ginkgo v1.10.3 // indirect
 	github.com/onsi/gomega v1.7.1 // indirect
 	github.com/pingcap/check v0.0.0-20190102082844-67f458068fc8
 	github.com/pingcap/errors v0.11.4
-	github.com/pingcap/kvproto v0.0.0-20191210040729-c23886becb54
+	github.com/pingcap/kvproto v0.0.0-20191211072056-705501101234
```
overvenus: Please use kvproto@release-3.1

SunRunAway: Updated.

pkg/storage/gcs.go
```diff
@@ -0,0 +1,152 @@
...
+
+// Write file to storage
+func (s *gcsStorage) Write(ctx context.Context, name string, data []byte) error {
+	object := s.gcs.Prefix + name
+	wc := s.bucket.Object(object).NewWriter(ctx)
+	wc.StorageClass = s.gcs.StorageClass
+	wc.PredefinedACL = s.gcs.PredefinedAcl
+	_, err := wc.Write(data)
+	if err != nil {
+		return err
```
overvenus: Forget `wc.Close()`?

pkg/storage/gcs.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		if err == storage.ErrObjectNotExist {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func newGCSStorage(gcs *backup.GCS) (*gcsStorage, error) {
```
overvenus: What about adding a `Context`? It can be used in Line 141.

SunRunAway: Updated.

pkg/storage/gcs.go
SunRunAway: It seems need not to Close here, see https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/objects/main.go#L92-L110 as example.

pkg/storage/gcs.go
SunRunAway: I've created an issue https://github.com/pingcap/br/issues/106

pkg/storage/gcs.go
SunRunAway: I've created an issue #107, let's disscuss here.

overvenus:
Please fix the confliction for merging.

SunRunAway:
> Please fix the confliction for merging.

rebased.

pkg/storage/gcs.go
gregwebs: It happens automatically if you don't force one particular method like this. So you don't have to do anything to support it (other than not hard-code a particular auth technique).

SunRunAway: If BR is not deployed on google cloud but on someone's IDC, we need this credentials file.

gregwebs: Right, so I think all you need here is an if statement, something like:

``` go
creds := []byte(gcs.GetCredentialsBlob()
if creds != "" {
     	clientOps = append(clientOps, option.WithCredentialsJSON(creds))
}
```

SunRunAway: I'll handle this on #106 because we need to deal with `--send-credentials-to-tikv` option here.

pkg/storage/gcs.go
```diff
@@ -0,0 +1,152 @@
...
+	gcs.CredentialsBlob = string(b)
+	return nil
+}
+
+func defineGCSFlags(flags *pflag.FlagSet) {
+	flags.String(gcsEndpointOption, "", "Set the GCS endpoint URL")
+	flags.String(gcsStorageClassOption, "",
+		"Specify the GCS storage class for objects. "+
+			"If it is not set, objects uploaded are followed by the default storage class of the bucket. "+
+			"See https://cloud.google.com/storage/docs/storage-classes for valid values.")
```
overvenus: It would be better if you can wrap them within 80 chars.

PR #103, opened 2019-12-11
backup: store backupmeta to s3

DanielZhangQD:
Support storing backup meta to s3, and support backup and restore with s3.


pkg/storage/flags.go
```diff
@@ -6,8 +6,15 @@ import (
...
 )
 
+const (
+	// flagSendCredentialOption specify whether to send credentials to tikv
+	flagSendCredentialOption = "send-credentials-to-tikv"
+)
+
 // DefineFlags adds flags to the flag set corresponding to all backend options.
 func DefineFlags(flags *pflag.FlagSet) {
+	flags.BoolP(flagSendCredentialOption, "c", true,
```
overvenus: Should it be moved into storage/s3.go? What do you think @kennytm ？

pkg/storage/parse.go
```diff
@@ -37,7 +38,11 @@ func ParseBackend(rawURL string, options *BackendOptions) (*backup.StorageBacken
 		return &backup.StorageBackend{Backend: &backup.StorageBackend_Noop{Noop: noop}}, nil
 
 	case "s3":
-		s3 := &backup.S3{Bucket: u.Host, Prefix: u.Path}
+		if u.Host == "" {
+			return nil, errors.Errorf("please specify the bucket for s3 in %s", rawURL)
+		}
+		prefix := strings.Trim(u.Path, "/")
```
overvenus: Could you add tests to cover them?

pkg/storage/s3.go
```diff
@@ -1,56 +1,229 @@
...
 
 func getBackendOptionsFromS3Flags(flags *pflag.FlagSet) (options S3BackendOptions, err error) {
+	send, err := flags.GetBool(flagSendCredentialOption)
+	if err != nil {
+		err = errors.Trace(err)
+		return
+	}
+	if send {
+		options.AccessKey = os.Getenv(accessKeyEnv)
+		options.SecretAccessKey = os.Getenv(secretAccessKeyEnv)
```
overvenus: What about `NewEnvCredentials`? https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/#NewEnvCredentials

pkg/storage/flags.go
DanielZhangQD: gcs, azure or other cloud vendors may also need this.

pkg/storage/s3.go
DanielZhangQD: This is populated to send to TiKV if `send` required.
BR itself uses a credential chain.

pkg/storage/s3.go
```diff
@@ -1,56 +1,229 @@
...
 		err = errors.Trace(err)
 		return
 	}
-
+	if options.Endpoint != "" {
+		if !strings.HasPrefix(options.Endpoint, "https://") &&
+			!strings.HasPrefix(options.Endpoint, "http://") {
+			options.Endpoint = "http://" + options.Endpoint
+		}
+	}
```
kennytm: Please put this fixup into `apply()`. So that if we get `S3BackendOptions` from a config file (TOML/JSON), we still have the same fix.

(This also means perhaps Provider should be part of `S3BackendOptions` and the Getenv stuff should appear in `apply()`.)

pkg/storage/parse.go
DanielZhangQD: done

pkg/storage/storage.go
```diff
@@ -12,14 +12,19 @@ type ExternalStorage interface {
...
 }
 
 // Create creates ExternalStorage
 func Create(backend *backup.StorageBackend) (ExternalStorage, error) {
 	switch backend := backend.Backend.(type) {
 	case *backup.StorageBackend_Local:
 		return newLocalStorage(backend.Local.Path)
+	case *backup.StorageBackend_S3:
+		if backend.S3 == nil {
+			return nil, errors.Errorf("no s3 config in %#v", backend)
```
kennytm: Would this expose the secrets in the log?

pkg/storage/s3.go
overvenus: Oh, I mean we can use `NewEnvCredentials` instead of hand-written `os.Getenv`.

DanielZhangQD:
@kennytm @overvenus @tennix PTAL new commit with aws-sdk.

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/storage.go
DanielZhangQD: Since backend.S3 is nil, there should be no secret in the log. Anyway, I will remove the backend.

DanielZhangQD:
@kennytm @overvenus @tennix comments addressed, PTAL.

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+	if options.Provider == "alibaba" || options.Provider == "netease" ||
+		options.UseAccelerateEndpoint {
+		options.ForcePathStyle = false
+	}
+	if options.SendCredential {
```
SunRunAway:
```suggestion
	if options.SendCredential && options.AccessKey == "" && options.SecretAccessKey == ""  {
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+	}
+	cred := credentials.NewChainCredentials(providers)
 
-	// TODO: Add more options here.
+	if qs.Region == "" && qs.Endpoint == "" {
+		qs.Endpoint = "https://s3.amazonaws.com/"
+	}
+	if qs.Region == "" {
+		qs.Region = "us-east-1"
+	}
```
kennytm: Let's move these into `apply()`

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+	awsConfig := aws.NewConfig().
+		WithMaxRetries(maxRetries).
+		WithCredentials(cred).
+		WithS3ForcePathStyle(qs.ForcePathStyle)
+	if qs.Region != "" {
+		awsConfig.WithRegion(qs.Region)
+	}
+	if qs.Endpoint != "" {
+		awsConfig.WithEndpoint(qs.Endpoint)
+	}
```
kennytm: The `!=` conditions are always true

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+		return nil, err
+	}
+	c := s3.New(ses)
+	err = checkS3Bucket(c, qs.Bucket)
+	if err != nil {
+		return nil, errors.Errorf("checkS3Bucket error: %v", err)
+	}
+
+	qs.Prefix = strings.Trim(qs.Prefix, "/")
+	qs.Prefix += "/"
```
kennytm: Let's move these into `apply()`.

pkg/storage/s3.go
DanielZhangQD: This is to populate the credentials to send to TiKV, not used by BR.
BR creates the credential with chains in `newS3Storage`.

pkg/backup/client.go
```diff
@@ -107,7 +107,11 @@ func (bc *Client) SetStorage(backend *backup.StorageBackend) error {
 		return err
 	}
 	// backupmeta already exists
-	if exist := bc.storage.FileExists(utils.MetaFile); exist {
+	exist, err := bc.storage.FileExists(utils.MetaFile)
+	if err != nil {
+		return errors.Errorf("error occurred when check %s file: %v", utils.MetaFile, err)
```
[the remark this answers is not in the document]
DanielZhangQD: done

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+	Region                string `json:"region" toml:"region"`
+	StorageClass          string `json:"storage_class" toml:"storage_class"`
+	SSE                   string `json:"sse" toml:"sse"`
+	ACL                   string `json:"acl" toml:"acl"`
+	AccessKey             string `json:"access_key" toml:"access_key"`
+	SecretAccessKey       string `json:"secret_access_key" toml:"secret_access_key"`
+	Provider              string `json:"provider" toml:"provider"`
+	ForcePathStyle        bool   `json:"force_path_style" toml:"force_path_style"`
+	UseAccelerateEndpoint bool   `json:"use_accelerate_endpoint" toml:"use_accelerate_endpoint"`
+	SendCredential        bool   `json:"send_credential" toml:"send_credential"`
```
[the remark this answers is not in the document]
DanielZhangQD: As you have said, we have to consider the case where we get the options from toml or json config files, then, we'd better add them here so they can also be set in config files.

pkg/storage/s3.go
SunRunAway: If acessKey and secretAccessKey are defined in BR explicitly, we should use them to send to TiKV, but not the environment credentials. Peharps we should do it as follow.
```go
if options.SendCredential {
    if options.AccessKey == "" && options.SecretAccessKey == "" {
        // get EnvCredentials 
    }
    options.AccessKey = v.AccessKeyID
    options.SecretAccessKey = v.SecretAccessKey
} else {
    options.AccessKey = ""
    options.SecretAccessKey = ""
}
```

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/s3.go
DanielZhangQD: Remove the check for region.

pkg/storage/s3.go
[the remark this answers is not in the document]
kennytm: We get the entire `BackendOptions` from the config file, not individual `S3BackendOptions`...

pkg/storage/s3.go
DanielZhangQD: The prefix is not in options struct.

pkg/storage/s3.go
DanielZhangQD: We do not have options for the keys in the command line.

pkg/storage/s3.go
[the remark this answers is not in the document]
DanielZhangQD: OK, I will move `SendCredential` out.

DanielZhangQD:
@kennytm @overvenus @SunRunAway comments addressed, PTAL again.

pkg/storage/s3.go
kennytm: @DanielZhangQD isn't worse if Rust handle it differently? (e.g. BR uploaded stuff to `us-east-1` but Rust uploaded to some other random region)

pkg/storage/s3.go
kennytm: it is in `*backup.S3`

pkg/storage/s3.go
DanielZhangQD: The code block has been moved to `getBackendOptionsFromS3Flags`, so if we get keys from other paths, e.g. config file later, the keys from the config file will be applied directly with `ParseBackend`.

pkg/storage/s3.go
DanielZhangQD: `apply` is to apply the options to backup.S3, it's not reasonable to change the fields that are not in the options.
BTW, I have trimed the "/" before calling `apply`.

pkg/storage/s3.go
DanielZhangQD: What I concern about is that `Region` is required by aws-sdk-go but may not in Rust lib.
If I send the updated region to TiKV, it may cause failure of Rust lib.
What about just move below part to `apply`:
```
if qs.Region == "" && qs.Endpoint == "" {
		qs.Endpoint = "https://s3.amazonaws.com/"
	}
```

pkg/storage/flags.go
```diff
@@ -6,8 +6,15 @@ import (
...
+		"If send credentials to tikv(default true)")
```
tennix:
```suggestion
		"Whether send credentials to tikv")
```
I think the default value will be printed by the `flags` package.

pkg/storage/parse_test.go
```diff
@@ -1,12 +1,11 @@
...
+package storage
 
 import (
 	"testing"
 
 	. "github.com/pingcap/check"
 	"github.com/pingcap/kvproto/pkg/backup"
-
-	"github.com/pingcap/br/pkg/storage"
+	// "github.com/pingcap/br/pkg/storage"
```
tennix: Why not remove this directly?

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/defaults"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/spf13/pflag"
+
```
tennix: Remove this empty line.

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+	GetObject(*s3.GetObjectInput) (*s3.GetObjectOutput, error)
+	PutObject(*s3.PutObjectInput) (*s3.PutObjectOutput, error)
+	HeadBucket(*s3.HeadBucketInput) (*s3.HeadBucketOutput, error)
+	WaitUntilObjectExists(*s3.HeadObjectInput) error
+}
+
+// S3Storage info for s3 storage
+type S3Storage struct {
+	session *session.Session
+	// svc     *s3.S3
```
tennix: Remove it?

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
 
 func (options *S3BackendOptions) apply(s3 *backup.S3) error {
 	if options.Endpoint == "" && options.Region == "" {
 		return errors.New("must provide either 's3.region' or 's3.endpoint'")
 	}
-	// TODO: Verify Region.
+	if options.Endpoint != "" {
+		if !strings.HasPrefix(options.Endpoint, "https://") &&
+			!strings.HasPrefix(options.Endpoint, "http://") {
+			options.Endpoint = "http://" + options.Endpoint
```
tennix: Why add `http` by default? To avoid guessing the HTTP scheme, I suggest making user pass the full endpoint.

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+		}
+	}
+	if options.Provider == "alibaba" || options.Provider == "netease" ||
```
tennix: Could you add a document to this for why this is needed? For example, the documentation link to it.

pkg/storage/s3.go
DanielZhangQD: Will remove the `Trim` here.

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
-	flags.String(s3EndpointOption, "", "Set the AWS S3 endpoint URL")
-	flags.String(s3RegionOption, "", "Set the AWS region")
-	// TODO: Finalize the list of options.
-	_ = flags.MarkHidden(s3EndpointOption)
+	flags.String(s3EndpointOption, "", "Set the S3 endpoint URL")
+	flags.String(s3RegionOption, "", "Set the S3 region")
+	flags.String(s3StorageClassOption, "", "Set the S3 storage class")
+	flags.String(s3SSEOption, "", "Set the S3 server-side encryption algorithm")
+	flags.String(s3ACLOption, "", "Set the S3 canned ACLs")
+	flags.String(s3ProviderOption, "", "Set the S3 provider")
```
tennix: Can you add some examples of the providers? Users may not know what the provider is.

pkg/storage/s3.go
DanielZhangQD: Will remove the whole block to `apply`.

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
 }
 
 func getBackendOptionsFromS3Flags(flags *pflag.FlagSet) (options S3BackendOptions, err error) {
+	send, err := flags.GetBool(flagSendCredentialOption)
+	if err != nil {
+		err = errors.Trace(err)
+		return
+	}
+	if send {
+		options.AccessKey = os.Getenv("AWS_ACCESS_KEY_ID")
```
tennix: Is there any way to get these credentials from the chained credentials? Because this is confusing, if its behavior not matching the AWS sdk.

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+// newS3Storage initialize a new s3 storage for metadata
+func newS3Storage(s3Back *backup.S3) (*S3Storage, error) {
+	qs := *s3Back
+	v := credentials.Value{
+		AccessKeyID:     qs.AccessKey,
+		SecretAccessKey: qs.SecretAccessKey,
+	}
+
+	// low timeout to ec2 metadata service
+	lowTimeoutClient := &http.Client{Timeout: 1 * time.Second}
```
tennix: Define this timeout as constant?

pkg/storage/s3.go
```diff
@@ -1,56 +1,318 @@
...
+	if err != nil {
+		err = errors.Trace(err)
+		return
+	}
+
+	return options, err
+}
+
+// newS3Storage initialize a new s3 storage for metadata
+func newS3Storage(s3Back *backup.S3) (*S3Storage, error) {
```
tennix: The name `s3Back` is weird. What about using `backend`?

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/s3.go
DanielZhangQD: done

DanielZhangQD:
@kennytm @SunRunAway @overvenus @tennix comments addressed, PTAL again.

pkg/storage/flags.go
DanielZhangQD: updated.

pkg/storage/parse_test.go
DanielZhangQD: Removed.

pkg/storage/s3.go
DanielZhangQD: This is to distinguish between the third-party packages and the pingcap packages.

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/s3.go
DanielZhangQD: Updated.

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/s3.go
DanielZhangQD: Updated.

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/s3.go
DanielZhangQD: done

pkg/storage/s3.go
kennytm: We could [call `.Get()`](https://godoc.org/github.com/aws/aws-sdk-go/aws/credentials#Credentials.Get) on any credentials to obtain the [tokens](https://godoc.org/github.com/aws/aws-sdk-go/aws/credentials#Value).

The problem is that non-static credentials can expire, the tokens obtained at the beginning may be useless after a while, breaking the BR process.

I think we may want to extract the tokens fields _out_ of StorageBackend, and recompute the tokens for every gRPC call of Backup/Download, if we want to support sending non-static credentials.

```proto
message BackupRequest {
    ...
    StorageBackend storage_backend = 9;
    AuthSecrets auth_secrets = 10;   // <-- new
}

message S3 {
    ...
    reserved 8 to 9;  // remove keys
}

message AuthSecrets {
    oneof secrets {
        S3AuthSecrets s3 = 3;
        GCSAuthSecrets gcs = 4;
    }
}

message S3AuthSecrets {
    string access_key = 8;
    string secret_access_key = 9;
    string session_token = 10;
}
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,327 @@
...
+	def := defaults.Get()
+	def.Config.HTTPClient = lowTimeoutClient
+	ec2Session, err := session.NewSession()
+	if err != nil {
+		return nil, err
+	}
 
-	return
+	// first provider to supply a credential set "wins"
+	providers := []credentials.Provider{
```
gregwebs: Adding an explicit chain here can easily mess up the default chain behavior. It seems like we want the default chain unless the credentials are already present in program memory. So this can be changed to:

``` go
if qs.AccessKey != "" {
    aws_config.WithCredentials(credentials.NewStaticCredentials(qs.AccessKey, qs.SecretAccessKey))
} else {
   // Nothing to do here!
}

Comment (DanielZhangQD) [reply pkg/storage/s3.go]: I just add the static value provider before the default providers, it should have no impact to default chain behavior.
Anyway, I have updated following your suggestion, PTAL again, thanks!

**Comment by gregwebs** [comment]
LGTM

File: pkg/storage/s3.go [hunk 23]
```diff
@@ -1,56 +1,288 @@
...
+	options *backup.S3
+}
+
+// S3BackendOptions contains options for s3 storage
 type S3BackendOptions struct {
-	Endpoint string `json:"endpoint" toml:"endpoint"`
-	Region   string `json:"region" toml:"region"`
+	Endpoint              string `json:"endpoint" toml:"endpoint"`
+	Region                string `json:"region" toml:"region"`
+	StorageClass          string `json:"storage_class" toml:"storage_class"`
```
kennytm:
```suggestion
	StorageClass          string `json:"storage-class" toml:"storage-class"`
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,288 @@
...
+	SSE                   string `json:"sse" toml:"sse"`
+	ACL                   string `json:"acl" toml:"acl"`
+	AccessKey             string `json:"access_key" toml:"access_key"`
```
kennytm:
```suggestion
	AccessKey             string `json:"access-key" toml:"access-key"`
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,288 @@
...
+	SecretAccessKey       string `json:"secret_access_key" toml:"secret_access_key"`
```
kennytm:
```suggestion
	SecretAccessKey       string `json:"secret-access-key" toml:"secret-access-key"`
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,288 @@
...
+	Provider              string `json:"provider" toml:"provider"`
+	ForcePathStyle        bool   `json:"force_path_style" toml:"force_path_style"`
```
kennytm:
```suggestion
	ForcePathStyle        bool   `json:"force-path-style" toml:"force-path-style"`
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,288 @@
...
+	UseAccelerateEndpoint bool   `json:"use_accelerate_endpoint" toml:"use_accelerate_endpoint"`
```
kennytm:
```suggestion
	UseAccelerateEndpoint bool   `json:"use-accelerate-endpoint" toml:"use-accelerate-endpoint"`
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,288 @@
...
-	flags.String(s3EndpointOption, "", "Set the AWS S3 endpoint URL")
-	flags.String(s3RegionOption, "", "Set the AWS region")
-	// TODO: Finalize the list of options.
-	_ = flags.MarkHidden(s3EndpointOption)
+	flags.String(s3EndpointOption, "", "Set the S3 endpoint URL, please specify the http or https scheme explicitly")
+	flags.String(s3RegionOption, "", "Set the S3 region, e.g. us-east-1")
+	flags.String(s3StorageClassOption, "", "Set the S3 storage class, e.g. STANDARD")
+	flags.String(s3SSEOption, "", "Set the S3 server-side encryption algorithm, e.g. AES256")
+	flags.String(s3ACLOption, "", "Set the S3 canned ACLs, e.g. authenticated-read")
+	flags.String(s3ProviderOption, "", "Set the S3 provider, e.g. aws, alibaba, ceph")
```
kennytm: This setting doesn't seem to have any effect yet besides `force-path-style`. I suggest exposing `force-path-style` instead of this option, until the provider can really change the endpoint etc.

pkg/storage/s3_test.go
```diff
@@ -0,0 +1,457 @@
...
+	testFn := func(test *testcase, c *C) {
+		c.Log(test.name)
+		if test.setEnv {
+			os.Setenv("AWS_ACCESS_KEY_ID", "ab")
+			os.Setenv("AWS_SECRET_ACCESS_KEY", "cd")
+		}
+		u, err := ParseBackend("s3://bucket/prefix/", &BackendOptions{S3: test.options})
+		s3 := u.GetS3()
+		c.Assert(err, IsNil)
+		c.Assert(true, Equals, equalS3(s3, test.s3))
```
kennytm:
```suggestion
		c.Assert(equalS3(s3, test.s3), IsTrue)
```

or maybe

```suggestion
		c.Assert(s3, DeepEquals, test.s3)
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,288 @@
...
+	awsConfig := aws.NewConfig().
+		WithMaxRetries(maxRetries).
+		WithS3ForcePathStyle(qs.ForcePathStyle).
+		WithRegion(qs.Region)
+	if qs.Endpoint != "" {
+		awsConfig.WithEndpoint(qs.Endpoint)
+	}
+	if cred != nil {
+		awsConfig.WithCredentials(cred)
+	}
```
kennytm: Move lines 159–162 here?

```go
if qs.AccessKey != "" && qs.SecretAccessKey != "" {
    awsConfig.WithCredentials(credentials.NewStaticCredentials(qs.AccessKey, qs.SecretAccessKey, ""))
}
```

pkg/storage/s3.go
```diff
@@ -1,56 +1,288 @@
...
+			}
+			backend.AccessKey = v.AccessKeyID
+			backend.SecretAccessKey = v.SecretAccessKey
+		}
+	}
+
+	c := s3.New(ses)
+	err = checkS3Bucket(c, qs.Bucket)
+	if err != nil {
+		return nil, errors.Errorf("checkS3Bucket error: %v", err)
```
kennytm:
```suggestion
		return nil, errors.Errorf("Bucket %s is not accessible: %v", qs.Bucket, err)
```

The error message is visible to the CLI users.

pkg/storage/s3.go
SunRunAway: Backup proto uses `_`, https://github.com/pingcap/kvproto/blob/master/proto/backup.proto#L131-L143

kennytm: @SunRunAway Why unify with kvproto? We are not sending these options through gRPC. We use `-` in all config files.

pkg/storage/s3.go
```diff
@@ -1,56 +1,286 @@
...
+	return &S3Storage{
+		session: ses,
+		svc:     c,
+		options: &qs,
+	}, nil
 }
 
-// TODO: Define S3 storage.
+// checkBucket checks if a bucket exists and creates it if not
+var checkS3Bucket = func(svc *s3.S3, bucket string) error {
```
tennix: Why define a closure instead of normal function?

DanielZhangQD: This makes it easy to inject UT functions.

PR #101, opened 2019-12-11
backup: add raw backup command

3pointer:
support backup raw kv according to #86

Issue #86, opened 2019-12-06
RawKV Backup & Restore Project


3pointer:
# RawKV Backup & Restore Project

# Abstract

In this project we needs focus on [TiKV](https://github.com/tikv/tikv) and [BR](https://github.com/pingcap/br) project. and we need to
- Implement Raw KV backup and restore based on current [BR](https://github.com/pingcap/br) project.

### Problem Statement

For now TiKV is an independent CNCF project, the ability of backup and restore data is become more important, so we need to implement backup and restore data in TiKV with BR.
Raw KV is a concept of TiKV and irrelevant with tidb, current BR already support backup Txn KV from TiDB TiKV cluster and restore Txn KV data to new cluster.
we can reuse above process to implement Raw KV backup and restore, so the main process can be divided into two section.
Backup:
- Use BR Backup Command to send backup request to TiKV cluster.
- TiKV receive backup request then start to backup Raw KV, then put data into specify storage(local storage for now).

Restore:
- Use BR Restore Command to send restore request to TiKV cluster.
- TiKV restore data from specify storage.

# Success Criteria

- With [BR](https://github.com/pingcap/br) program, one can backup Raw KV from exist TiKV cluster, and restore to new TiKV cluster successfully.

# TODO list

- [x] TiKV use RegionIterator to scan Raw KV by given key range. *(500 points / medium)* [detail](https://github.com/tikv/tikv/issues/6213)
- [x] TiKV restore raw kv depends on current TiKV restore logic. *(500 points / medium)* [detail](https://github.com/tikv/tikv/issues/6220)
- [ ] BR support communicate with only TiKV cluster. *(300 points / medium)*
- [ ] BR support restore Raw KV command. *(500 points / medium)*
- [ ] BR support backup Raw KV command. *(500 points / medium)*

## Difficulty

* Easy

## Score

* 2300

## Mentor(s)

* @3pointer  

## Recommended skills

* Go language
* Basic understanding of TiKV

# References

- [BR introduction](https://pingcap.com/docs-cn/dev/how-to/maintain/backup-and-restore/br/#%E5%A4%87%E4%BB%BD%E5%8E%9F%E7%90%86)
- [TiKV Storage Source Code Reading](https://pingcap.com/blog-cn/tikv-source-code-reading-11/)

youjiali1995:
I want to join in this task~

breeswish:
I want to join in this task~

GregoryIan:
@3pointer any documents about it?

PR #104, opened 2019-12-12
Implement Raw Restore

MyonKeminta:
This PR implements Raw Restore.

This is part of work of https://github.com/pingcap/br/issues/86 .
Requires:
- [x] https://github.com/pingcap/kvproto/pull/524
- [x] Some utils included in #101 for parsing keys


cmd/restore.go
```diff
@@ -384,6 +397,101 @@ func newTableRestoreCommand() *cobra.Command {
...
+			if err != nil {
+				return errors.Trace(err)
+			}
+			defer client.Close()
+			err = initRestoreClient(client, cmd.Flags())
+			if err != nil {
+				return errors.Trace(err)
+			}
+
+			if client.IsRawKvMode() {
```
3pointer:
```suggestion
			if !client.IsRawKvMode() {
```

cmd/restore.go
```diff
@@ -384,6 +397,101 @@ func newTableRestoreCommand() *cobra.Command {
...
+				// Split/Scatter + Download/Ingest
+				int64(len(ranges)+len(files)),
+				!HasLogFile())
+
+			err = restore.SplitRanges(ctx, client, ranges, rewriteRules, updateCh)
+			if err != nil {
+				return errors.Trace(err)
+			}
+			pdAddrs := strings.Split(pdAddr, ",")
+			err = client.ResetTS(pdAddrs)
```
5kbpers: It seems not necessary to reset ts in raw mode.

pkg/restore/import.go
```diff
@@ -233,6 +250,19 @@ func (importer *FileImporter) downloadSST(
 		return nil, true, errRewriteRuleNotFound
 	}
 	sstMeta := getSSTMetaFromFile(id, file, regionInfo.Region, regionRule)
+	// For raw kv mode, cut the SST file's range to fit in the restoring range.
+	if importer.isRawKvMode {
+		if bytes.Compare(importer.rawStartKey, sstMeta.Range.GetStart()) > 0 {
```
5kbpers: Why not use the range of the region?

MyonKeminta: It's possible that we are restoring a range that's a subrange of backup-ed data, and we need to filter out keys that are not in the range, and the range's border may be neither region's border nor SST file's border. So cut the range of SSTMeta here.

5kbpers: But we download and ingest SSTs per region, use the range of backup data may cause a region ingests extra data.

cmd/restore.go
```diff
@@ -384,6 +397,101 @@ func newTableRestoreCommand() *cobra.Command {
 	return command
 }
 
+func newRawRestoreCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "raw",
+		Short: "restore a raw kv range",
+		RunE: func(cmd *cobra.Command, _ []string) error {
```
gregwebs: The cobra commands should just handle cli options and then call a function. Then it will be easy to create other interfaces to br or to use as a library.

pkg/restore/import.go
MyonKeminta: No, I only meant to cut the range but never enlarge it. So the only possibility is that the range is made smaller than before. Is my code wrong?

cmd/restore.go
MyonKeminta: Sorry I don't know. I just tried to keep my code consistent with other commands here. @5kbpers Do you have any idea?

MyonKeminta:
@3pointer @overvenus @5kbpers PTAL thanks!

MyonKeminta:
@3pointer @overvenus @5kbpers PTAL thanks!

pkg/task/restore_raw.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if len(files) == 0 {
+		return errors.New("all files are filtered out from the backup archive, nothing to restore")
+	}
+	summary.CollectInt("restore files", len(files))
+
+	// Empty rewrite rules
+	//rewriteRules := &restore.RewriteRules{
+	//	Data: []*import_sstpb.RewriteRule{{}},
+	//}
```
overvenus: Need clean up?

pkg/task/restore_raw.go
```diff
@@ -0,0 +1,133 @@
+package task
```
overvenus: Please add license header.

MyonKeminta: I found that other files in the project have no license headers either.

pkg/task/restore_raw.go
```diff
@@ -0,0 +1,128 @@
...
+	ranges, err := restore.ValidateFileRanges(files, nil)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	// Redirect to log if there is no log file to avoid unreadable output.
+	// TODO: How to show progress?
+	updateCh := utils.StartProgress(
+		ctx,
+		"Table Restore",
```
3pointer: Raw Restore

pkg/restore/import.go
```diff
@@ -328,6 +357,55 @@ func (importer *FileImporter) downloadSST(
...
+	}
+	// Empty rule
+	var rule import_sstpb.RewriteRule
+	sstMeta := getSSTMetaFromFile(id, file, regionInfo.Region, &rule)
+
+	// Cut the SST file's range to fit in the restoring range.
+	if bytes.Compare(importer.rawStartKey, sstMeta.Range.GetStart()) > 0 {
+		sstMeta.Range.Start = importer.rawStartKey
+	}
+	// TODO: importer.RawEndKey is exclusive but sstMeta.Range.End is inclusive. How to exclude importer.RawEndKey?
```
3pointer: I saw the only use of sstMeta in tikv is before [ingest](https://github.com/tikv/tikv/blob/113b3987d63885f602857be7ad6ad906a8811988/components/raftstore/src/store/fsm/apply.rs#L2153),  this TODO seems already fixed by tikv `download` already. @kennytm  PTAL

pkg/task/backup_raw.go
```diff
@@ -26,15 +26,20 @@ const (
...
+type RawKvConfig struct {
 	Config
 
 	StartKey []byte `json:"start-key" toml:"start-key"`
 	EndKey   []byte `json:"end-key" toml:"end-key"`
 	CF       string `json:"cf" toml:"cf"`
 }
 
+// BackupRawConfig is the configuration specific for backup tasks.
+type BackupRawConfig struct {
```
3pointer: Maybe `RawKVConfig` here is enough? `BackupRawConfig` seems nothing special

MyonKeminta: I left it here just trying to make a good semantics

pkg/restore/import.go
kennytm: The Download() in TiKV's sst_importer is expecting an inclusive `sstMeta.Range.End`, and returned range is also inclusive. What do you mean by "fixed"?

3pointer: I mean `exclude importer.RawEndKey` has been done already during `download`. because`RawEndKey` has not been backup.

pkg/task/restore_raw.go
overvenus: License header added in #182 .

MyonKeminta:
@3pointer  @overvenus  PTAL again thanks!

kennytm:
```
[2020-03-16T07:23:45.207Z] pkg/task/restore_raw.go:52:20: not enough arguments in call to newMgr

[2020-03-16T07:23:45.207Z] 	have (context.Context, glue.Glue, []string, TLSConfig)

[2020-03-16T07:23:45.207Z] 	want (context.Context, glue.Glue, []string, TLSConfig, conn.UnexpectedStoreBehavior)
```

PR #109, opened 2019-12-12
Add lastbackupts argument  to support incremental backup/restore

YangKeao:
This PR has two part: backup and restore:

Most modification on backup is :

1. Set `StartVersion` for `BackupRequest`

Most modification on restore is:

1. Filter files with `lastbackupts`. Import only file after `lastbackupts`

Close #85


cmd/validate.go
```diff
@@ -116,9 +118,10 @@ func newCheckSumCommand() *cobra.Command {
 					s := sha256.Sum256(data)
 					if !bytes.Equal(s[:], file.Sha256) {
 						return errors.Errorf(`
-backup data checksum failed: %s may be changed
-calculated sha256 is %s,
-origin sha256 is %s`, file.Name, hex.EncodeToString(s[:]), hex.EncodeToString(file.Sha256))
+								backup data checksum failed: %s may be changed
+								calculated sha256 is %s,
+								origin sha256 is %s`,
```
overvenus: Please remove the leading `\t`s, otherwise, the log looks weird.

```
								backup data checksum failed: %s may be changed
```

YangKeao: Fixed

cmd/restore.go
```diff
@@ -135,8 +131,13 @@ func runRestore(flagSet *flag.FlagSet, cmdName, dbName, tableName string) error
...
 	}
 
+	newTS, err := client.GetTS(ctx)
+	if err != nil {
+		return err
+	}
+
 	summary.CollectInt("restore files", len(files))
-	rewriteRules, newTables, err := client.CreateTables(mgr.GetDomain(), tables)
+	rewriteRules, newTables, err := client.CreateTables(mgr.GetDomain(), tables, newTS)
```
overvenus: We don't need to rewrite ts if it's a full restore.

pkg/restore/client.go
```diff
@@ -276,11 +279,17 @@ func (rc *Client) RestoreTable(
...
 	// We should encode the rewrite rewriteRules before using it to import files
 	encodedRules := encodeRewriteRules(rewriteRules)
+
 	err = rc.setSpeedLimit()
 	if err != nil {
 		return err
 	}
+
 	for _, file := range table.Files {
+		err := rc.setSpeedLimit()
```
overvenus: SpeedLimit should be set only once.

PR #115, opened 2019-12-16
GCS: support --send-credentials-to-tikv and get credentials from GKE or GCE

### LINKED ISSUE #106: GCS: support --send-credentials-to-tikv and get credentials from GKE or GCE
1. We can get credentials from GKE or GCE, see https://cloud.google.com/docs/authentication/production#auth-cloud-compute-engine-go as example
2. We should support --send-credentials-to-tikv when using GCS.

SunRunAway:
Closes #106


pkg/storage/gcs.go
```diff
@@ -139,6 +139,14 @@ func (s *gcsStorage) FileExists(ctx context.Context, name string) (bool, error)
 }
 
 func newGCSStorage(ctx context.Context, gcs *backup.GCS) (*gcsStorage, error) {
+	if gcs.CredentialsBlob == "" {
```
overvenus:
```suggestion
	if gcs.CredentialsBlob == ""  && sendCredential {
```

gregwebs:
LGTM

pkg/storage/gcs.go
SunRunAway: No, it's not related to `sendCredential`.
We should get credentials from environment if user does not provide credentials explicitly.

SunRunAway:
> Could you add some unit tests to improve coverage?

Ok, I'll try.

overvenus:
Could you fix conflicts?

PR #121, opened 2019-12-18
*: use oracle package to manipulate ts and test gc safe point 

overvenus:


pkg/backup/client.go
```diff
@@ -71,31 +72,34 @@ func (bc *Client) GetTS(ctx context.Context, timeAgo string) (uint64, error) {
...
 
-	ts := utils.Timestamp{
-		Physical: p,
-		Logical:  l,
+	// check backup time do not exceed GCSafePoint
+	safePoint, err := GetGCSafePoint(ctx, bc.mgr.GetPDClient())
+	if err != nil {
+		return 0, errors.Trace(err)
+	}
+	if backupTS < safePoint {
```
kennytm:
```suggestion
	if backupTS <= safePoint {
```

Just for consistency, since `pkg/backup/safe_point.go` uses `<=` 🙃 (BTW should it reuse `CheckGCSafepoint`? the only difference is the handing of `GetGCSafePoint` error.)

PR #123, opened 2019-12-19
restore: auto remove/add pd scheduler in restore

3pointer:


cmd/restore.go
```diff
@@ -168,113 +212,18 @@ func newDbRestoreCommand() *cobra.Command {
...
 				return err
 			}
-			defer mgr.Close()
-
-			client, err := restore.NewRestoreClient(
-				ctx, mgr.GetPDClient(), mgr.GetTiKV())
-			if err != nil {
-				return errors.Trace(err)
+			if len(db) == 0 {
+				return errors.Errorf("empty database name is not allowed")
```
kennytm:
```suggestion
				return errors.New("empty database name is not allowed")
```

cmd/restore.go
```diff
@@ -283,122 +232,29 @@ func newTableRestoreCommand() *cobra.Command {
...
-
-			dbName, err := cmd.Flags().GetString("db")
-			if err != nil {
-				return errors.Trace(err)
-			}
-			db := client.GetDatabase(dbName)
-			if db == nil {
-				return errors.New("not exists database")
+			if len(db) == 0 {
+				return errors.Errorf("empty database name is not allowed")
```
kennytm:
```suggestion
				return errors.New("empty database name is not allowed")
```

cmd/restore.go
```diff
@@ -283,122 +232,29 @@ func newTableRestoreCommand() *cobra.Command {
...
-
-			// Checksum
-			updateCh = utils.StartProgress(
-				ctx, "Checksum", int64(len(newTables)), !HasLogFile())
-			err = client.ValidateChecksum(
-				ctx, mgr.GetTiKV().GetClient(), []*utils.Table{table}, newTables, updateCh)
-			if err != nil {
-				return err
+			if len(table) == 0 {
+				return errors.Errorf("empty table name is not allowed")
```
kennytm:
```suggestion
				return errors.New("empty table name is not allowed")
```

cmd/restore.go
```diff
@@ -56,108 +67,141 @@ func NewRestoreCommand() *cobra.Command {
...
-		Use:   "full",
-		Short: "restore all tables",
-		RunE: func(cmd *cobra.Command, _ []string) error {
-			pdAddr, err := cmd.Flags().GetString(FlagPD)
-			if err != nil {
-				return errors.Trace(err)
-			}
-			ctx, cancel := context.WithCancel(GetDefaultContext())
-			defer cancel()
+func runRestore(flagSet *flag.FlagSet, cmdName, dbName, tableName string) error {
```
kennytm: Nice :+1:

pkg/conn/conn.go
```diff
@@ -58,11 +64,15 @@ func pdGet(ctx context.Context, addr string, prefix string, cli *http.Client) ([
...
+	var (
+		req  *http.Request
+		resp *http.Response
+	)
+	req, err = http.NewRequestWithContext(ctx, method, url, body)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	resp, err := cli.Do(req)
+	resp, err = cli.Do(req)
```
kennytm: Why removing the `:=`?

3pointer: okay, it's not necessary, change it back

3pointer:
> Rest LGTM.
> 
> Are the schedulers still removed even in online import mode?

Sure, I  think we should check is online in prepareWork

pkg/restore/client.go
```diff
@@ -387,17 +392,11 @@ func (rc *Client) RestoreAll(
 
 //SwitchToImportModeIfOffline switch tikv cluster to import mode
 func (rc *Client) SwitchToImportModeIfOffline(ctx context.Context) error {
-	if rc.isOnline {
-		return nil
-	}
```
kennytm: Well with this change the function name is inaccurate again 🙃

cmd/restore.go
```diff
@@ -446,3 +302,59 @@ func initRestoreClient(ctx context.Context, client *restore.Client, flagSet *fla
...
+	}
+	err := client.SwitchToNormalMode(ctx)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	return addPDLeaderScheduler(ctx, mgr, removedSchedulers)
+}
+
+func addPDLeaderScheduler(ctx context.Context, mgr *conn.Mgr, removedSchedulers []string) error {
+	for _, scheduler := range removedSchedulers {
```
kennytm: Question. Since map is randomly ordered, the schedulers are added in random order as well. Is this OK for PD?

overvenus: AFAIK, yes, there is no dependency problem between each scheduler.

PR #128, opened 2019-12-24
validate: debug backupmeta

3pointer:
#80 is broken after BR project open source

### DIFF
--- a/tests/br_debug_meta/run.sh
+++ b/tests/br_debug_meta/run.sh
@@ -1,0 +1,55 @@
+#!/bin/sh
+#
+# Copyright 2019 PingCAP, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


tests/br_debug_meta/run.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ ! -f "$TEST_DIR/$DB/backupmeta.json" ]; then
+    echo "TEST: [$TEST_NAME] failed!"
+    exit 1
+fi
+
+# Test validate encode
+run_br validate encode -s "local://$TEST_DIR/$DB"
+
+# should generate backupmeta_from_json
+if [ ! -f "$TEST_DIR/$DB/backupmeta_from_json" ]; then
```
kennytm: we should compare `backupmeta` and `backupmeta_from_json` are the same file (`diff` them)?

3pointer:
> Rest LGTM.
> 
> Could we do something for the coverage?

let my try

PR #130, opened 2019-12-30
update release 3.1

overvenus:
Merge recent commits into release-3.1


tests/br_incremental/run.sh
```diff
@@ -0,0 +1,75 @@
...
+# incremental backup
+echo "incremental backup start..."
+run_br --pd $PD_ADDR backup table -s "local://$TEST_DIR/$DB" --db $DB -t $TABLE --ratelimit 5 --concurrency 4 --lastbackupts $last_backup_ts
+
+start_ts=$(br validate decode --field="start-version" -s "local://$TEST_DIR/$DB" | tail -n1)
+end_ts=$(br validate decode --field="end-version" -s "local://$TEST_DIR/$DB" | tail -n1)
+
+echo "start version: $start_ts, end version: $end_ts"
+
+# incremental restore
```
3pointer: It seems we did not cleanup data after Line43, so it always success even though restore failed?

pkg/restore/import.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/pingcap/kvproto/pkg/kvrpcpb"
 	"github.com/pingcap/log"
 	restore_util "github.com/pingcap/tidb-tools/pkg/restore-util"
+	"github.com/pingcap/tidb/util/codec"
```
3pointer: CI failed, seems should use pd [codec](https://github.com/pingcap/br/pull/137/files#diff-eae46bd20abe35cad68c49bd179a3728R14)

3pointer: but the codec in pd and tidb seems no different

overvenus: Turns out I exchange `DecodeBytes` arguments by mistake.

PR #133, opened 2020-01-03
flags: hidden not stable command flags

3pointer:


overvenus:
Could you hidden `validate` subcommand too?

PR #137, opened 2020-01-06
restore: support table partition

5kbpers:
Support table partition:
1. add rewrite rule for table partition
2. fix table partition checksum
3. refactor logic about key rewrite

### DIFF
--- a/pkg/restore/client.go
+++ b/pkg/restore/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"sort"
 	"sync"
 	"time"
 
@@ -17,7 +16,6 @@ import (
 	restore_util "github.com/pingcap/tidb-tools/pkg/restore-util"
 	"github.com/pingcap/tidb/domain"
 	"github.com/pingcap/tidb/kv"
-	"github.com/pingcap/tidb/tablecodec"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
@@ -200,11 +198,6 @@ func (rc *Client) CreateTables(
 		Data:  make([]*import_sstpb.RewriteRule, 0),
 	}
 	newTables := make([]*model.TableInfo, 0, len(tables))
-	// Sort the tables by id for ensuring the new tables has same id ordering as the old tables.
-	// We require this constrain since newTableID of tableID+1 must be not bigger
-	// than newTableID of tableID.
-	sort.Sort(utils.Tables(tables))
-	tableIDMap := make(map[int64]int64)
 	for _, table := range tables {
 		err := rc.db.CreateTable(rc.ctx, table)
 		if err != nil {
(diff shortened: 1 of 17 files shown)


tests/br_table_partition/run.sh
```diff
@@ -0,0 +1,59 @@
...
+    fi
+    echo "database $DB${i} [original] row count: ${row_count_ori[i]}, [after br] row count: ${row_count_new[i]}"
+done
+
+if $fail; then
+    echo "TEST: [$TEST_NAME] failed!"
+    exit 1
+else
+    echo "TEST: [$TEST_NAME] successed!"
+fi
```
overvenus: Should `DROP DATABASE` before finishing the test.

pkg/restore/util.go
```diff
@@ -237,94 +268,42 @@ func ValidateFileRewriteRule(file *backup.File, rewriteRules *restore_util.Rewri
...
-) *import_sstpb.RewriteRule {
-	for _, rule := range rewriteRules.Data {
-		// regions may have the new prefix
-		if bytes.HasPrefix(region.GetStartKey(), rule.GetNewKeyPrefix()) {
-			return rule
-		}
+// Rewrites a raw key and returns a encoded key
+func rewriteRawKey(key []byte, rewriteRules *restore_util.RewriteRules) ([]byte, *import_sstpb.RewriteRule) {
+	if len(key) > 0 {
+		rule := matchOldPrefix(key, rewriteRules)
```
3pointer: should check nil here?

pkg/restore/util.go
```diff
@@ -71,32 +77,48 @@ func GetRewriteRules(
...
-
-	dataRules := make([]*import_sstpb.RewriteRule, 0, len(oldTable.Indices)+1)
-	dataRules = append(dataRules, &import_sstpb.RewriteRule{
-		OldKeyPrefix: append(tablecodec.EncodeTablePrefix(oldTable.ID), recordPrefixSep...),
-		NewKeyPrefix: append(tablecodec.EncodeTablePrefix(newTable.ID), recordPrefixSep...),
-		NewTimestamp: newTimeStamp,
-	})
-
+	tableIDs := make(map[int64]int64)
+	tableIDs[oldTable.ID] = newTable.ID
```
3pointer: if we got `oldTable.Partition != nil` then tableID is useless
should it be?
```
if oldTable.Partition != nil {
	for _, srcPart := range oldTable.Partition.Definitions {
		for _, destPart := range newTable.Partition.Definitions {
			if srcPart.Name == destPart.Name {
				tableIDs[srcPart.ID] = destPart.ID
			}
		}
	}
} else {
	tableIDs[oldTable.ID] = newTable.ID
}
```

5kbpers: We don't have partitioned index so tableID is necessary.
BTW I think extra rewrite rules do not have a bad effect.

pkg/restore/client.go
```diff
@@ -200,11 +198,6 @@ func (rc *Client) CreateTables(
 		Data:  make([]*import_sstpb.RewriteRule, 0),
 	}
 	newTables := make([]*model.TableInfo, 0, len(tables))
-	// Sort the tables by id for ensuring the new tables has same id ordering as the old tables.
-	// We require this constrain since newTableID of tableID+1 must be not bigger
-	// than newTableID of tableID.
-	sort.Sort(utils.Tables(tables))
```
overvenus: Can we remove `utils.Table`?

pkg/restore/client.go
```diff
@@ -215,21 +208,10 @@ func (rc *Client) CreateTables(
 			return nil, nil, err
 		}
 		rules := GetRewriteRules(newTableInfo, table.Schema, newTS)
-		tableIDMap[table.Schema.ID] = newTableInfo.ID
 		rewriteRules.Table = append(rewriteRules.Table, rules.Table...)
 		rewriteRules.Data = append(rewriteRules.Data, rules.Data...)
 		newTables = append(newTables, newTableInfo)
 	}
-	// If tableID + 1 has already exist, then we don't need to add a new rewrite rule for it.
-	for oldID, newID := range tableIDMap {
```
overvenus: We should update validate.go since we change the rewrite process.

pkg/restore/util.go
3pointer: yes, extra rewrite rules seems no bad effect

PR #138, opened 2020-01-08
tests: add a simple test case showing independent VIEWs can be backed up and restored

kennytm:


kennytm:
There is a question about views depending on other tables and views (need to sort by dependency during backup). We could improve the test case about that in the future?

overvenus:
SGTM, I have opened an issue https://github.com/pingcap/br/issues/145.

kennytm:
#142 make BR cannot work with VIEWs 🤷. Closing.

PR #139, opened 2020-01-10
restore: fix alter auto increment id for no-primary-key table

5kbpers:

Previously we rely on CREATE TABLE SQL to recover AutoIncID, but the API TiDB supplied does not set AutoIncID in CREATE TABLE when a table does not have a primary key.
Now we change to execute an `Alter AutoIncrementID` SQL after creating a table.

### DIFF
--- a/pkg/restore/db.go
+++ b/pkg/restore/db.go
@@ -92,7 +92,10 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 			zap.Error(err))
 		return errors.Trace(err)
 	}
-	alterAutoIncIDSQL := fmt.Sprintf("alter table %s auto_increment = %d", schema.Name, schema.AutoIncID)
+	alterAutoIncIDSQL := fmt.Sprintf(
+		"alter table %s auto_increment = %d",
+		escapeTableName(schema.Name),
+		schema.AutoIncID)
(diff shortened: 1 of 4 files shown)


kennytm:
LGTM but remember to test INSERT after RESTORE later.

3pointer:
> Signed-off-by: 5kbpers [tangminghua@pingcap.com](mailto:tangminghua@pingcap.com)
> 
> Previously we rely on CREATE TABLE SQL to recover AutoIncID, but the API TiDB supplied does not set AutoIncID in CREATE TABLE when a table does not have a primary key.
> Now we change to execute an `Alter AutoIncrementID` SQL after creating a table.

seems it does not set AutoIncID in CREATE TABLE when a table does not have an auto increment column

5kbpers:
@3pointer right

pkg/restore/db.go
```diff
@@ -92,6 +92,16 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 			zap.Error(err))
 		return errors.Trace(err)
 	}
+	alterAutoIncIDSQL := fmt.Sprintf("alter table %s auto_increment = %d", schema.Name, schema.AutoIncID)
```
overvenus: Do you need `EncloseName`?

PR #140, opened 2020-01-10
metrics: add grafana scripts

WangXiangUSTC:
cp from https://github.com/pingcap/tidb-ansible/blob/master/scripts/br.json


metrics/grafana/br.json
```diff
@@ -0,0 +1,1690 @@
...
+      "id": "singlestat",
+      "name": "Singlestat",
+      "version": ""
+    }
+  ],
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "${DS_TEST_CLUSTER}",
```
glorv: should be `DS_TEST-CLUSTER`?

WangXiangUSTC: fixed in aaf7cbb

PR #144, opened 2020-01-15
cmd: convert version command to flags

overvenus:
Convert version command to flags.


cmd/cmd.go
```diff
@@ -52,12 +53,18 @@ const (
...
 
-	flagTable = "table"
+	flagVersion      = "version"
+	flagVersionShort = "V"
 )
 
 // AddFlags adds flags to the given cmd.
 func AddFlags(cmd *cobra.Command) {
+	cmd.Version = utils.BRInfo()
+	cmd.Flags().BoolP(flagVersion, flagVersionShort, false, "Display information about BR")
```
kennytm:
```suggestion
	cmd.Flags().BoolP(flagVersion, flagVersionShort, false, "Display version information about BR")
```

pkg/utils/version.go
```diff
@@ -18,23 +21,30 @@ var (
 	BRGitBranch      = "None"
 )
 
-// LogBRInfo prints the BR version information.
+//go:linkname goVersion runtime.buildVersion
+var goVersion string
```
kennytm: why not just use [`runtime.Version()`](https://golang.org/pkg/runtime/#Version)? They are [equivalent](https://github.com/golang/go/blob/cfe3cd903f018dec3cb5997d53b1744df4e53909/src/runtime/proc.go#L14). (perhaps update TiDB too)

PR #151, opened 2020-02-04
backup: add backupts file as lock

3pointer:
1.save backup ts to file as a file lock before backup begins
2.output backup ts after backup

add save backup ts logic

### Check List 

Tests 

 - Manual test check the generate file after backup it outputs like this
```
**************************************
Started backup at: 2020-02-04 15:37:51
BackupTS: 414397725835264003
**************************************
```

Related changes

### DIFF
--- a/cmd/backup.go
+++ b/cmd/backup.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
@@ -108,6 +109,11 @@ func runBackup(flagSet *pflag.FlagSet, cmdName, db, table string) error {
 		return err
 	}
 
+	err = client.SaveBackupTS(ctx, backupTS)
+	if err != nil {
+		return err
+	}
+
 	defer summary.Summary(cmdName)
 
 	ranges, backupSchemas, err := backup.BuildBackupRangeAndSchema(
@@ -170,6 +176,12 @@ func runBackup(flagSet *pflag.FlagSet, cmdName, db, table string) error {
 	if err != nil {
 		return err
 	}
+
+	content, err := client.GetBackupTS(ctx)
+	if err != nil {
+		return err
+	}
+	fmt.Println(string(content))
 	return nil
 }
 
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -126,8 +126,8 @@ func (bc *Client) SaveBackupTS(ctx context.Context, backupTS uint64) error {
 	backendURL := storage.FormatBackendURL(bc.backend)
 	log.Info("save backup ts", zap.Stringer("path", &backendURL))
 	fileString := fmt.Sprintf(
-		"**************************************" +
-			"\nStarted backup at: %v\nBackupTS: %d\n" +
+		"**************************************"+
+			"\nStarted backup at: %v\nBackupTS: %d\n"+
 			"**************************************",
 		time.Now().Format("2006-01-02 15:04:05"), backupTS)
 	return bc.storage.Write(ctx, utils.TSFile, []byte(fileString))
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io"
 	"sync"
 	"time"
@@ -118,6 +119,23 @@ func (bc *Client) SetStorage(ctx context.Context, backend *backup.StorageBackend
 	}
 	bc.backend = backend
 	return nil
+}
+
+// SaveBackupTS saves the backup ts at the given path.
+func (bc *Client) SaveBackupTS(ctx context.Context, backupTS uint64) error {
+	backendURL := storage.FormatBackendURL(bc.backend)
+	log.Info("save backup ts", zap.Stringer("path", &backendURL))
+	fileString := fmt.Sprintf(
+		"**************************************" +
+			"\nStarted backup at: %v\nBackupTS: %d\n" +
+			"**************************************",
+		time.Now().Format("2006-01-02 15:04:05"), backupTS)
+	return bc.storage.Write(ctx, utils.TSFile, []byte(fileString))
+}
+
+// GetBackupTS get the backup ts from given path.
+func (bc *Client) GetBackupTS(ctx context.Context) ([]byte, error) {
+	return bc.storage.Read(ctx, utils.TSFile)
 }
 
 // SaveBackupMeta saves the current backup meta at the given path.
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
 	"io"
 	"sync"
 	"time"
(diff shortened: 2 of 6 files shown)


kennytm:
This is the same as the backupmeta's `EndVersion` field?

3pointer:
> This is the same as the backupmeta's `EndVersion` field?

yes, it is easy for user to get this field and it write to file to prevent multiple backup at same directory before backup start.

pkg/backup/client.go
```diff
@@ -120,6 +121,23 @@ func (bc *Client) SetStorage(ctx context.Context, backend *backup.StorageBackend
...
 }
 
+// SaveBackupTS saves the backup ts at the given path.
+func (bc *Client) SaveBackupTS(ctx context.Context, backupTS uint64) error {
+	backendURL := storage.FormatBackendURL(bc.backend)
+	log.Info("save backup ts", zap.Stringer("path", &backendURL))
+	fileString := fmt.Sprintf(
+		"**************************************"+
+			"\nStarted backup at: %v\nBackupTS: %d\n"+
+			"**************************************",
```
overvenus: Can it be pretty-printed json?

kennytm:
As I understand it,

1. The `backupts` is only a human-readable representation of the `EndVersion` field in the `backupmeta`.
2. It is saved *before* backup starts, and before `backupmeta` file is written.
3. The `backupts` file is read at the end of backup, but only used to log the output, so there's no semantic value about the content of this file.
4. The file is not used in restore.
5. Backup doesn't check for existence of `backupts` (unlike `backupmeta`).
6. There is no check whether `backupts` and `backupmeta`'s `EndVersion` are consistent.

So I'm not quite sure what is the exact purpose of `backupts`, compared with `br validate decode`.

3pointer:
> Where is the lock? I do not see the file work as a lock.

Not strictly lock, I mean if check the file exists before backup starts, then other backup progress will exits before they start backup on tikv

pkg/backup/client.go
```diff
@@ -120,6 +120,35 @@ func (bc *Client) SetStorage(ctx context.Context, backend *backup.StorageBackend
...
+	f := &fileContent{
+		StartedAt: time.Now().Format("2006-01-02 15:04:05"),
+		BackupTS:  backupTS,
+	}
+	fileBytes, err := json.MarshalIndent(f, "", "\t")
+	if err != nil {
+		return err
+	}
+
+	return bc.storage.Write(ctx, utils.TSFile, fileBytes)
```
overvenus: Does it fail if there is TSFile already?

pkg/backup/client.go
```diff
@@ -120,6 +120,35 @@ func (bc *Client) SetStorage(ctx context.Context, backend *backup.StorageBackend
 	return nil
 }
 
+// SaveBackupTS saves the backup ts at the given path.
+func (bc *Client) SaveBackupTS(ctx context.Context, backupTS uint64) error {
```
overvenus: We should add a test.

overvenus:
@kennytm The file not only record backup ts, but also works like a LOCK, it tries its best to prevent multiple concurrent backup writing to the same location which could lead to data corruption.

kennytm:
@overvenus At the time I've written [the comment](https://github.com/pingcap/br/pull/151#issuecomment-582272594) it did not function as a lock (the fix was added in 6dd158f). 

This PR fixes TET-277 but this connection is not mentioned at all in the PR, and PR description also did not explain "what **problem** does this PR solve", it only describes how this PR works. So reviewer unfamiliar with the Incremental Backup QA Tests cannot understand why we need to write a second file which has the subset of information of `backupmeta`, and can only reverse engineer the rationale from the code (which had a bug before 6dd158f and skew my understanding). 

This is something we should beware of when building an open community. The summary of TET-277 and https://github.com/pingcap/br/pull/151#issuecomment-581831798 should replace the "What problem does this PR solve?" session of the PR.

----

Some alternatives brainstormed:

* We may also consider adding a flag `--lastbackupmeta` to Incremental Restore which directly parses a `backupmeta` file and extracts the `EndVersion`.
* We may format `backupmeta` as JSON.
* We may construct an empty `backupmeta` when we start to backup to function as the lock.

cmd/backup.go
```diff
@@ -170,6 +176,12 @@ func runBackup(flagSet *pflag.FlagSet, cmdName, db, table string) error {
 	if err != nil {
 		return err
 	}
+
+	content, err := client.GetBackupTS(ctx)
+	if err != nil {
+		return err
+	}
+	fmt.Println(string(content))
```
kennytm: I don't think we still need to log the backup TS if we write it to a human-readable file.

pkg/utils/schema.go
```diff
@@ -17,6 +17,8 @@ import (
 const (
 	// MetaFile represents file name
 	MetaFile = "backupmeta"
+	// TSFile represents backup ts file, it generated before backup
+	TSFile = "backupts"
```
kennytm: `backupts.json`?

pkg/backup/client.go
```diff
@@ -120,6 +120,35 @@ func (bc *Client) SetStorage(ctx context.Context, backend *backup.StorageBackend
...
+	}
+	backendURL := storage.FormatBackendURL(bc.backend)
+
+	log.Info("save backup ts", zap.Stringer("path", &backendURL))
+	type fileContent struct {
+		StartedAt string `json:"started_at"`
+		BackupTS  uint64 `json:"backup_ts"`
+	}
+	f := &fileContent{
+		StartedAt: time.Now().Format("2006-01-02 15:04:05"),
```
overvenus: Could you log the time in backupTS?

cmd/backup.go
overvenus:
```suggestion
	cmd.Println(string(content))
```

3pointer:
> @overvenus At the time I've written [the comment](https://github.com/pingcap/br/pull/151#issuecomment-582272594) it did not function as a lock (the fix was added in [6dd158f](https://github.com/pingcap/br/commit/6dd158f62ff004d02e454e6199d3c2d12b891472)).
> 
> This PR fixes TET-277 but this connection is not mentioned at all in the PR, and PR description also did not explain "what **problem** does this PR solve", it only describes how this PR works. So reviewer unfamiliar with the Incremental Backup QA Tests cannot understand why we need to write a second file which has the subset of information of `backupmeta`, and can only reverse engineer the rationale from the code (which had a bug before [6dd158f](https://github.com/pingcap/br/commit/6dd158f62ff004d02e454e6199d3c2d12b891472) and skew my understanding).
> 
> This is something we should beware of when building an open community. The summary of TET-277 and [#151 (comment)](https://github.com/pingcap/br/pull/151#issuecomment-581831798) should replace the "What problem does this PR solve?" session of the PR.
> 
> Some alternatives brainstormed:
> 
> * We may also consider adding a flag `--lastbackupmeta` to Incremental Restore which directly parses a `backupmeta` file and extracts the `EndVersion`.
> * We may format `backupmeta` as JSON.
> * We may construct an empty `backupmeta` when we start to backup to function as the lock.

@kennytm It's my fault that I didn't make the purpose of this PR clearly in description, Sorry for that and I agree with that make purpose more clearly is important when building an open community. 
also
> * We may format `backupmeta` as JSON.
> * We may construct an empty `backupmeta` when we start to backup to function as the lock.
seems clearly can solve this problem, @overvenus what do you think?

overvenus:
> We may format backupmeta as JSON.
 
I don't think it's a good idea to format backupmeta as JSON, eg. uint64 is not strictly defined in JSON, and may cause some unknown compatibility issues in the future. 

> We may construct an empty backupmeta when we start to backup to function as the lock.

LGTM.

kennytm:
Practically both Go's `encoding/json` and Rust's `serde_json` supports loss-less parsing of i64 so it is not an issue. Unless you wanna parse that `backupmeta` in Javascript 🙃.  But if i64 is the only blocking issue we could also consider other human-readable format such as TOML or YAML 🙃.

overvenus:
Let's stay protobuf for now, change it to JSON sounds a breaking change.

PR #152, opened 2020-02-04
restore: enhance error handling

5kbpers:

This PR enhances the error handling of restore:
1. stop and return when receiving an error which cannot retry
2. update region info and retry to ingest SST only

Add `Backoffer` interface to unify the retry logic.

### Check List 

Tests 

 - Unit test
 - Integration test (**TODO**)

Code changes


pkg/restore/import.go
```diff
@@ -333,13 +296,29 @@ func (importer *FileImporter) ingestSST(
...
+	err := errGrpc
+	switch {
+	case strings.Contains(e.Error(), "bad format"):
+		err = errBadFormat
+	case strings.Contains(e.Error(), "wrong prefix"):
+		err = errWrongKeyPrefix
+	case strings.Contains(e.Error(), "corrupted"):
+		err = errFileCorrupted
+	case strings.Contains(e.Error(), "Cannot read"):
+		err = errCannotRead
```
overvenus: Could we define these errors in protobuf? @kennytm

kennytm: Frankly it's ridiculous that we return a structured error in Rust and it got serialized into *debug* representation. What if we change https://github.com/tikv/tikv/blob/3ba72ab98ed8f3155bca4bd8fed13d1ddbf1a02a/components/sst_importer/src/service.rs#L7-L10 and return a structured (JSON) message?

overvenus: I guess that is because we did not define the structured error in protobuf.

In our use case, I think we should choose "Richer error model" https://grpc.io/docs/guides/error/

> and return a structured (JSON) message?

IMO, we should use one data format to avoid confusion, that is protobuf.

kennytm: @overvenus Thanks for the pointer. The Rust `grpcio` package we're using [doesn't have that `repeated Any details = 3` field](https://github.com/tikv/grpc-rs/blob/35564e56f0becb3f014dee0c023717958e768f4f/src/call/mod.rs#L133-L141) described in https://cloud.google.com/apis/design/errors#error_model though.  So we need to first update `grpcio` before using the rich model.

overvenus: Another way is including error message in response.

Anyway, it should be future work, let's focus on enhance error handling.

pkg/restore/backoff.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/pingcap/br/pkg/utils"
+)
+
+var (
+	errNotLeader           = errors.New("not leader")
+	errKeyNotInRegion      = errors.New("key not in region")
+	errResp                = errors.New("response error")
+	errRewriteRuleNotFound = errors.New("rewrite rule not found")
+	errRangeIsEmpty        = errors.New("range is empty")
+	errGrpc                = errors.New("gRPC error")
```
kennytm: These will record the stack of `restore.init()`, making the stack trace useless. Unless you replace all calls of `errors.Trace()` with `errors.WithStack()`, please construct these errors using a stack-less constructor such as the [standard `errors.New`](https://pkg.go.dev/errors?tab=doc#New).

pkg/utils/retry.go
```diff
@@ -0,0 +1,32 @@
...
+func WithRetry(
+	retryableFunc RetryableFunc,
+	backoffer Backoffer,
+) error {
+	var lastErr error
+	for backoffer.Attempt() > 0 {
+		err := retryableFunc()
+		if err != nil {
+			lastErr = err
+			time.Sleep(backoffer.NextBackoff(err))
```
kennytm: Make this function accept a `Context` so if user does a Ctrl+C we don't need to block on this sleep.

pkg/utils/retry.go
```diff
@@ -23,7 +27,12 @@ func WithRetry(
 		err := retryableFunc()
 		if err != nil {
 			lastErr = err
-			time.Sleep(backoffer.NextBackoff(err))
+			select {
+			case <-ctx.Done():
+				return lastErr
+			default:
+				time.Sleep(backoffer.NextBackoff(err))
+			}
```
kennytm: This select is ineffective. Use `case <-time.After` instead.

kennytm:
/run-integration-test

(Previous failure for record: https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/br_ghpr_unit_and_integration_test/runs/782/nodes/45/log/?start=0)

pkg/restore/import.go
```diff
@@ -185,63 +170,56 @@ func (importer *FileImporter) Import(file *backup.File, rewriteRules *RewriteRul
...
 				}
-				return err
+				log.Error("download file failed",
+					zap.Stringer("file", file),
+					zap.Stringer("region", info.Region),
+					zap.Binary("startKey", startKey),
+					zap.Binary("endKey", endKey),
+					zap.Error(err1))
+				return err1
+			}
```
kennytm:
```go
switch errors.Cause(err1) {
case nil:
        // proceed to ingest
case errRewriteRuleNotFound, errRangeIsEmpty:
        // skip this region
        continue
default:
        log.Error(...)
        return err1
}
```

pkg/restore/import.go
```diff
@@ -185,63 +170,56 @@ func (importer *FileImporter) Import(file *backup.File, rewriteRules *RewriteRul
...
+			err1 = importer.ingestSST(downloadMeta, info)
+			// If error is `NotLeader`, update the region info and retry
+			for err1 == errNotLeader {
```
kennytm:
```suggestion
			for errors.Cause(err1) == errNotLeader {
```

PR #158, opened 2020-02-10
Reduce TiDB dependencies

### LINKED ISSUE #154: Integrate BR into TiDB - Refactor BR to abstract out TiDB to avoid import cycle

BR is currently a CLI program. To some DBAs, this interface is not as straightforward as managing via SQL directly.

We would like to invoke BR directly from TiDB via SQL. This involves several parts:

* `pingcap/parser` need to recognize the BR-related SQL statements.
* `pingcap/tidb` need to forward the requests to the BR library.
* `pingcap/br` (this repository) need to refactor a bit so TiDB can import it.

This issue mainly focuses on the last item.

BR itself imports various packages from TiDB. If such TiDB packages depend on BR again, we get an "import cycle". We need to modify BR to avoid those packages. (The cycle is checked at package level, not module level, so `go.mod` is hopefully irrelevant.)

kennytm:

Fix #154. Removes all transitive `tidb/session` and `tidb/executor` dependencies reachable from `br/pkg/task`, so that `br/pkg/task` itself can be used from `tidb/executor`.

1. Extract functions and types involving `tidb/session` and `tidb/executor` into a `br/pkg/glue.Glue` interface.
2. Removes the no-longer-used `ResultSetToStringSlice()` function (which itself uses `tidb/session`)
3. Keep the `mock_cluster` file test-only (which again uses `tidb/session`).

(Note: `go.mod` doesn't need to be changed. Cyclic dependency between modules is fine, e.g. `tidb` depends on `tidb-tools` which again depends on `tidb` which doesn't cause any failure.)

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


kennytm:
PTAL @3pointer

3pointer:
need resolve conflicts

PR #167, opened 2020-02-23
Batch restore

overvenus:

Scatter region does not work properly when restoring terabyte backup. This leads to an imbalanced cluster.

image

Split restore tasks into small batches (128 by default).

image

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes


pkg/task/common.go
```diff
@@ -78,9 +78,14 @@ func DefineCommonFlags(flags *pflag.FlagSet) {
...
 
 	flags.Uint64(flagRateLimit, 0, "The rate limit of the task, MB/s per node")
-	flags.Uint32(flagConcurrency, 4, "The size of thread pool on each node that executes the task")
 	flags.Bool(flagChecksum, true, "Run checksum at end of task")
 
+	// Default concurrency is different for backup and restore.
+	// Leave it 0 and let them adjust the value.
+	flags.Uint32(flagConcurrency, 0, "The size of thread pool on each node that executes the task")
+	// It may confuse users , so just hide it.
+	_ = flags.MarkHidden(flagConcurrency)
```
kennytm: 🤔 but i don't find this confusing

pkg/summary/collector_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	c.Assert(len(fields), Equals, 3)
+	assertContains := func(field zap.Field) {
+		for _, f := range fields {
+			if f.Key == field.Key {
+				c.Assert(f, DeepEquals, field)
+				return
+			}
+		}
+		c.Assert(fields, DeepEquals, field) // Must fail and prints detail
```
kennytm: ??? use [`c.Error`](https://godoc.org/github.com/pingcap/check#C.Error) if this must fail

pkg/summary/collector.go
```diff
@@ -97,19 +104,30 @@ func (tc *logCollector) CollectFailureUnit(name string, reason error) {
 func (tc *logCollector) CollectDuration(name string, t time.Duration) {
 	tc.mu.Lock()
 	defer tc.mu.Unlock()
-	tc.fields = append(tc.fields, zap.Duration(name, t))
+	val, ok := tc.durations[name]
+	if ok {
+		tc.durations[name] = val + t
+		return
+	}
+	tc.durations[name] = t
```
kennytm: This entire block can be simplified to

```go
tc.durations[name] += t
```

Demo: https://play.golang.org/p/5s8j5x7GSpl

pkg/summary/collector.go
```diff
@@ -97,19 +104,30 @@ func (tc *logCollector) CollectFailureUnit(name string, reason error) {
...
 func (tc *logCollector) CollectInt(name string, t int) {
 	tc.mu.Lock()
 	defer tc.mu.Unlock()
-	tc.fields = append(tc.fields, zap.Int(name, t))
+	val, ok := tc.ints[name]
+	if ok {
+		tc.ints[name] = val + t
+		return
+	}
+	tc.ints[name] = t
```
kennytm: (ditto)

pkg/backup/range_test.go
```diff
@@ -0,0 +1,148 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backup
```
kennytm: This test should be in the `rtree` package, not here.

pkg/backup/range.go
```diff
@@ -0,0 +1,73 @@
...
+	"encoding/hex"
+
+	"github.com/google/btree"
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+
+	"github.com/pingcap/br/pkg/utils/rtree"
+)
+
+func getIncompleteRange(
```
kennytm: this function seems more suitable to be in the `rtree` package.

pkg/task/restore.go
```diff
@@ -131,33 +147,55 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
...
 			log.Error("reset pd TS failed", zap.Error(err))
 			return err
 		}
 	}
 
-	removedSchedulers, err := restorePreWork(ctx, client, mgr)
-	if err != nil {
-		return err
+	// Restore sst files in batch.
+	batchSize := int(math.Min(maxRestoreBatchSizeLimit /* 256 */, float64(cfg.Concurrency)))
```
```suggestion
	batchSize := cfg.Concurrency
	if batchSize > maxRestoreBatchSizeLimit {
		batchSize = maxRestoreBatchSizeLimit // 256
	}
```

or use `mathutil.Min(cfg.Concurrency, maxRestoreBatchSizeLimit)` if you must need this to be single-line.

pkg/task/restore.go
```diff
@@ -131,33 +147,55 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
...
+	batches := make([][]rtree.Range, 0, (len(ranges)+batchSize-1)/batchSize)
+	for batchSize < len(ranges) {
+		ranges, batches = ranges[batchSize:], append(batches, ranges[0:batchSize:batchSize])
 	}
-	err = client.RestoreFiles(files, rewriteRules, updateCh)
-	// always run the post-work even on error, so we don't stuck in the import mode or emptied schedulers
-	postErr := restorePostWork(ctx, client, mgr, removedSchedulers)
+	batches = append(batches, ranges)
 
+	for _, rangeBatch := range batches {
```
kennytm: seems constructing the `batches` array is unnecessary? the two `for` loops can be merged?

pkg/task/common.go
overvenus: `--help` prints "concurrency .... (default 0)", I find this is confusing. :joy:

kennytm: `(default 0)` should not appear, otherwise the `--rate-limit` would also show `(default 0)`.

5kbpers:
Do you know why scatter region does not work? Is it because the timeout is too short?

overvenus:
> Do you know why scatter region does not work? Is it because the timeout is too short?

In short, yes.

In the test, we are scattering ~40k region at once, if a cluster is configured 5 nodes 3 replicas, it means PD needs to scatter ~48k replicas in a single scatter operator.

Based on my test, PD takes 15s to scatter 156 replicas, it is clearly scattering 48k replicas  will be timeout.

pkg/rtree/check.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"encoding/hex"
+
+	"github.com/google/btree"
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+)
+
+// CheckDupFiles checks if there are any files are duplicated.
+func CheckDupFiles(rangeTree *RangeTree) {
```
kennytm: IMO this function is irrelevant to RangeTree and should be placed in its original package. Not a big deal anyway.

Issue #170, opened 2020-02-27
Support timestamp parameter on backup


3pointer:
### Describe your feature request related problem:
Sometimes user need one certain timestamp in different backup, such as
* two different backup want the same timestamp snapshot

current [timeago](#9 ) can't solve above problem.
### Describe the feature you'd like:
support input timestamp on backup

kennytm:
Already implemented in #172. Closing.

PR #172, opened 2020-02-28
support backupts

3pointer:
finish issue #170 

* use same concept as [tidb snapshot](https://pingcap.com/docs/stable/how-to/get-started/read-historical-data/) to set certain backupts to br
* update storage help message: default `local:` -> `s3`

### Check List 

Tests 

 - Unit test

Related changes

### DIFF
--- a/pkg/task/backup.go
+++ b/pkg/task/backup.go
@@ -53,7 +53,7 @@ func DefineBackupFlags(flags *pflag.FlagSet) {
 	flags.Uint64(flagLastBackupTS, 0, "(experimental) the last time backup ts,"+
 		" use for incremental backup, support TSO only")
 	flags.String(flagBackupTS, "", "the backup ts support TSO or datetime,"+
-		" e.g. '400036290571534337', '2018-05-11 01:42:23' ")
+		" e.g. '400036290571534337', '2018-05-11 01:42:23'")
 }
 
 // ParseFromFlags parses the backup-related flags from the flag set.
--- a/pkg/task/backup.go
+++ b/pkg/task/backup.go
@@ -50,7 +50,8 @@ func DefineBackupFlags(flags *pflag.FlagSet) {
 		"The history version of the backup task, e.g. 1m, 1h. Do not exceed GCSafePoint")
 
 	// TODO: remove experimental tag if it's stable
-	flags.Uint64(flagLastBackupTS, 0, "(experimental) the last time backup ts, use for incremental backup, support TSO only")
+	flags.Uint64(flagLastBackupTS, 0, "(experimental) the last time backup ts,"+
+		" use for incremental backup, support TSO only")
 	flags.String(flagBackupTS, "", "the backup ts support TSO or datetime,"+
 		" e.g. '400036290571534337', '2018-05-11 01:42:23' ")
 }
--- a/pkg/task/backup.go
+++ b/pkg/task/backup.go
@@ -4,13 +4,18 @@ package task
 
 import (
 	"context"
+	"strconv"
 	"time"
 
 	"github.com/pingcap/errors"
 	kvproto "github.com/pingcap/kvproto/pkg/backup"
 	"github.com/pingcap/log"
 	"github.com/pingcap/parser/model"
+	"github.com/pingcap/parser/mysql"
 	"github.com/pingcap/tidb-tools/pkg/filter"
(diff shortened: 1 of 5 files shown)


pkg/task/backup.go
```diff
@@ -179,3 +197,25 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
...
+	}
+	t, err := types.ParseTime(sc, ts, mysql.TypeTimestamp, types.MaxFsp)
+	if err != nil {
+		return 0, errors.Trace(err)
+	}
+	t1, err := t.GoTime(time.Local)
+	if err != nil {
+		return 0, errors.Trace(err)
+	}
+	return variable.GoTimeToTS(t1), nil
```
kennytm: this looks very heavyweight to support the additional timestamp format. wdyt if we only allow TSO?

pkg/task/common.go
```diff
@@ -93,7 +93,7 @@ type Config struct {
 // DefineCommonFlags defines the flags common to all BRIE commands.
 func DefineCommonFlags(flags *pflag.FlagSet) {
 	flags.BoolP(flagSendCreds, "c", true, "Whether send credentials to tikv")
-	flags.StringP(flagStorage, "s", "", `specify the url where backup storage, eg, "local:///path/to/save"`)
+	flags.StringP(flagStorage, "s", "", `specify the url where backup storage, eg, "s3:///path/to/save"`)
```
kennytm:
```suggestion
	flags.StringP(flagStorage, "s", "", `specify the url where backup storage, eg, "s3:///bucket/prefix"`)
```

pkg/task/backup.go
3pointer: yes it's heavyweight , but I think unified the usage with tidb `set tidb_snapshot` is better, TSO is not easy for general users to get, @overvenus what's your opinion?

overvenus: LGTM.

kennytm: which looks good? 😅

overvenus: `variable.GoTimeToTS` is LGTM.

pkg/task/backup.go
```diff
@@ -37,7 +44,9 @@ func DefineBackupFlags(flags *pflag.FlagSet) {
 		flagBackupTimeago, 0,
 		"The history version of the backup task, e.g. 1m, 1h. Do not exceed GCSafePoint")
 
-	flags.Uint64(flagLastBackupTS, 0, "the last time backup ts")
+	flags.String(flagBackupTS, "", "the backup ts support TSO or datetime,"+
+		" e.g. '400036290571534337', '2018-05-11 01:42:23' ")
```
overvenus:
```suggestion
		" e.g. '400036290571534337', '2018-05-11 01:42:23'")
```

pkg/task/backup_test.go
```diff
@@ -0,0 +1,34 @@
...
+	c.Assert(err, IsNil)
+	c.Assert(int(ts), Equals, 0)
+
+	ts, err = parseTSString("400036290571534337")
+	c.Assert(err, IsNil)
+	c.Assert(int(ts), Equals, 400036290571534337)
+
+	ts, err = parseTSString("2018-05-11 01:42:23")
+	c.Assert(err, IsNil)
+	c.Assert(int(ts), Equals, 400024965742592000)
```
kennytm: This test will fail when `time.Local` is not UTC+8 🤔

pkg/task/backup_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	ts, err = parseTSString("400036290571534337")
+	c.Assert(err, IsNil)
+	c.Assert(int(ts), Equals, 400036290571534337)
+
+	utcTS := 400024965742563200
+	_, offset := time.Now().Local().Zone()
+	ts, err = parseTSString("2018-05-11 01:42:23")
+	c.Assert(err, IsNil)
+	c.Assert(int(ts), Equals, utcTS+offset)
```
kennytm: This looks wrong. The `offset` is the number of seconds after UTC (for CST this is +28800). The TSO is `(timestamp in milliseconds) << 18`. Adding `offset` directly to the TSO should not work.

The TSO of 2018-05-11 01:42:23 when Local = UTC is 400032515489792000, BTW.

PR #176, opened 2020-03-03
*: update tidb dependency build with go1.14

3pointer:
update tidb(https://github.com/pingcap/tidb/pull/14972) to make go build success in go1.14
also need fix this https://github.com/pingcap/tidb/issues/15093 first
add `AllocSeqCache` function in `idAlloctor` to adapt new tidb `Allocator` interface

### Check List 

Tests 

 - Unit test

Code changes

Related changes


pkg/restore/util.go
```diff
@@ -36,6 +36,10 @@ func (alloc *idAllocator) Alloc(tableID int64, n uint64, increment, offset int64
 	return alloc.id, alloc.id, nil
 }
 
+func (alloc *idAllocator) AllocSeqCache(sequenceID int64) (min int64, max int64, round int64, err error) {
+	return 0, 0, 0, nil
```
overvenus: Is it safe to return zero if we restore Sequence?

3pointer: do we support backup sequence?

kennytm: i don't think backup sequence need to go through this function

kennytm:
Is this PR still WIP?

3pointer:
> Is this PR still WIP?

yes, need fix https://github.com/pingcap/tidb/issues/15093 first

Issue #193, opened 2020-03-16
Support restoring to a cluster containing TiFlash


kennytm:
### Describe your feature request related problem:

Currently we'll reject BR restore if the cluster contains TiFlash.

### Describe the feature you'd like:

We should be able to skip TiFlash nodes during restore.

kennytm:
Current proposed solution 

1. Full restore: 

    * Before restoring, use `ALTER TABLE xxxx SET TIFLASH REPLICA 0` to prevent TiFlash from learning the Raft logs
    * Wait until all learner peers and all PD placement rules related to TiFlash are removed (how?)
    * Perform normal restore (without any other placement rules?)
    * Re-enable TiFlash. TiFlash should be able to catch up the diff at speed ≥30 GB/h/node.

2. Incremental restore:

    * Same as full restore

PTAL @ilovesoup @overvenus 

(In the future TiFlash may fully support the sst_importer service, thus rendering these workaround eventually obsolete.)

PR #194, opened 2020-03-17
restore: remove tiflash replica before restore

5kbpers:

Support restoring to a cluster containing TiFlash, see https://github.com/pingcap/br/issues/193

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Increased code complexity


pkg/restore/db.go
```diff
@@ -129,6 +129,21 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 			zap.Stringer("table", table.Info.Name),
 			zap.Error(err))
 	}
+
+	// TODO: remove this after tiflash supports restore
+	removeTiFlashSQL := fmt.Sprintf(
+		"alter table %s set tiflash replica 0",
+		utils.EncloseName(tableInfo.Name.O),
+	)
+	err = db.se.Execute(ctx, removeTiFlashSQL)
```
3pointer: what if ignore this error when tidb not support set tiflash replica

overvenus:
Should change `GetAllTiKVStore` on restore, it fails if there is a tiflash node.

pkg/restore/db.go
5kbpers: Would fail, do we need to support the versions of tidb which don't support tiflash?

3pointer:
Do we need add replica back after restore?

pkg/restore/client.go
```diff
@@ -305,6 +305,71 @@ func (rc *Client) CreateTables(
...
+		err = rc.db.se.Execute(rc.ctx, removeTiFlashSQL)
+		if err != nil {
+			log.Error("remove tiflash replica failed",
+				zap.String("query", removeTiFlashSQL),
+				zap.Stringer("db", table.Db.Name),
+				zap.Stringer("table", table.Info.Name),
+				zap.Error(err))
+			return err
+		} else if replicaCount > 0 {
+			log.Warn("remove tiflash replica done, please recover it manually later",
```
3pointer: what if print one executable DDL `alter table xxx set tiflash replica replicaCount`?

pkg/restore/client.go
```diff
@@ -305,6 +305,71 @@ func (rc *Client) CreateTables(
...
+				zap.Stringer("db", table.Db.Name),
+				zap.Stringer("table", table.Info.Name),
+				zap.Int("originalReplicaCount", replicaCount))
+		}
+	}
+	return nil
+}
+
+// GetTiFlashStores returns an id list of tiflash stores.
+func (rc *Client) GetTiFlashStores() ([]uint64, error) {
```
overvenus: Could you add an option to `GatAllTiKVStores` to return TiFlash store only?

pkg/restore/split.go
```diff
@@ -192,6 +215,30 @@ func (rs *RegionSplitter) isScatterRegionFinished(ctx context.Context, regionID
...
+	}
+	for _, peer := range regionInfo.Region.GetPeers() {
+		if rejectStores[peer.GetStoreId()] {
+			return true, nil
+		}
+	}
+	retryTimes := ctx.Value(retryTimes).(int)
+	if retryTimes > 10 {
+		log.Warn("get region info", zap.Stringer("region", regionInfo.Region))
+	}
```
overvenus: Could you handle the retry outside of the method?

pkg/restore/client.go
```diff
@@ -305,6 +305,71 @@ func (rc *Client) CreateTables(
 	return rewriteRules, newTables, nil
 }
 
+// RemoveTiFlashReplica removes all the tiflash replica of a table
+func (rc *Client) RemoveTiFlashReplica(tables []*utils.Table) error {
+	stores, err := rc.GetTiFlashStores()
+	if err != nil {
+		return errors.Trace(err)
+	}
```
3pointer: if len(stores) > 0  need return immediately

go.mod
```diff
@@ -20,9 +20,10 @@ require (
 	github.com/onsi/gomega v1.8.1 // indirect
 	github.com/pingcap/check v0.0.0-20200212061837-5e12011dc712
 	github.com/pingcap/errors v0.11.5-0.20190809092503-95897b64e011
-	github.com/pingcap/kvproto v0.0.0-20200228095611-2cf9a243b8d5
+	github.com/pingcap/kvproto v0.0.0-20200317112120-78042b285b75
 	github.com/pingcap/log v0.0.0-20200117041106-d28c14d3b1cd
 	github.com/pingcap/parser v0.0.0-20200305120128-bde9faa0df84
+	github.com/pingcap/pd v2.1.19+incompatible
```
3pointer: should remove this?

pkg/restore/client.go
```diff
@@ -305,6 +334,87 @@ func (rc *Client) CreateTables(
...
+	backendURL := storage.FormatBackendURL(rc.backend)
+	log.Info("update backup meta", zap.Stringer("path", &backendURL))
+	err = rc.storage.Write(rc.ctx, utils.MetaFile, backupMetaData)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	for _, table := range tables {
```
3pointer: should we skip table doesn't need set replica?

pkg/restore/split.go
5kbpers: No, the region info would be printed here.

3pointer:
/run-integration-test pd=release-4.0 tidb=release-4.0 tikv=release-4.0

pkg/conn/conn.go
```diff
@@ -103,6 +103,9 @@ const (
 	// SkipTiFlash causes GetAllTiKVStores to skip the store when it is found to
 	// be a TiFlash node.
 	SkipTiFlash UnexpectedStoreBehavior = 1
+	// TiFlashOnly caused GetAllTiKVStores to skip the store which is not a
+	// TiFlash node.
+	TiFlashOnly UnexpectedStoreBehavior = 2
```
overvenus: Could you rename `UnexpectedStoreBehavior` to `StoreBehavior`?

pkg/restore/client.go
```diff
@@ -305,6 +334,70 @@ func (rc *Client) CreateTables(
...
+			Crc64Xor:        table.Crc64Xor,
+			TotalKvs:        table.TotalKvs,
+			TotalBytes:      table.TotalBytes,
+			TiflashReplicas: uint32(table.TiFlashReplicas),
+		})
+	}
+
+	// Update backup meta
+	rc.backupMeta.Schemas = schemas
+	backupMetaData, err := proto.Marshal(rc.backupMeta)
```
3pointer: the origin backupmeta schemas has changed to new-created schemas, but it seems has no affect on next restore. will it affect on our current debug procedure?

cmd/restore.go
```diff
@@ -30,6 +38,59 @@ func runRestoreRawCommand(command *cobra.Command, cmdName string) error {
...
+	defer store.Close()
+	db, err := restore.NewDB(tidbGlue, store)
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	for _, d := range dbs {
+		for _, t := range d.Tables {
+			log.Info("get table", zap.Stringer("name", t.Info.Name), zap.Int("replica", t.TiFlashReplicas))
```
overvenus: Could you `cmd.Println` the result?

pkg/restore/client.go
```diff
@@ -305,6 +334,70 @@ func (rc *Client) CreateTables(
...
+	if err != nil {
+		return errors.Trace(err)
+	}
+	backendURL := storage.FormatBackendURL(rc.backend)
+	log.Info("update backup meta", zap.Stringer("path", &backendURL))
+	err = rc.storage.Write(rc.ctx, utils.MetaFile, backupMetaData)
```
overvenus: Could you write to another file? Backup files are supposed to be immutable.

pkg/restore/client.go
```diff
@@ -305,6 +334,70 @@ func (rc *Client) CreateTables(
...
+			Table:           tableData,
+			Crc64Xor:        table.Crc64Xor,
+			TotalKvs:        table.TotalKvs,
+			TotalBytes:      table.TotalBytes,
+			TiflashReplicas: uint32(table.TiFlashReplicas),
+		})
+	}
+
+	// Update backup meta
+	rc.backupMeta.Schemas = schemas
```
overvenus: Can we skip write to storage if there is no TiFlash node?

pkg/utils/pd.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	defer resp.Body.Close()
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(resp.Body)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if resp.StatusCode == 412 {
```
overvenus: 412 feels like magic.

cmd/restore.go
5kbpers: Added a progress bar.

pkg/task/restore.go
```diff
@@ -349,3 +369,51 @@ func splitPostWork(ctx context.Context, client *restore.Client, tables []*model.
...
+	if err != nil {
+		return err
+	}
+
+	tables := make([]*utils.Table, 0)
+	for _, db := range dbs {
+		tables = append(tables, db.Tables...)
+	}
+	updateCh := utils.StartProgress(
+		ctx, "Checksum", int64(len(tables)), !cfg.LogProgress)
```
3pointer:
```suggestion
		ctx, "RestoreTiflashReplica", int64(len(tables)), !cfg.LogProgress)
```

5kbpers:
/run-integration-test pd=release-4.0 tidb=release-4.0 tikv=release-4.0

PR #198, opened 2020-03-19
restore: adjust PD config to speed up restore

overvenus:

Tweak a few PD configurations to speed up restore.

### Check List 

Tests 

 - Integration test

Side effects

 - Increased code complexity

Related changes


kennytm:
/run-integration-test tikv=release-4.0 pd=release-4.0 tidb=release-4.0

pkg/conn/conn.go
```diff
@@ -433,6 +434,53 @@ func (mgr *Mgr) listSchedulersWith(ctx context.Context, get pdHTTPRequest) ([]st
 	return nil, err
 }
 
+// GetPDScheduleConfig returns PD schedule config value associated with the key.
+// It returns nil if there is no such config item.
+func (mgr *Mgr) GetPDScheduleConfig(
+	ctx context.Context, configKey string,
```
kennytm: better replace `configKey string` by `configKeys []string`, or even better, `cfgPointer interface{}`, so that fetching N config values doesn't require performing N `pdRequest`s.

pkg/conn/conn.go
```diff
@@ -433,6 +434,53 @@ func (mgr *Mgr) listSchedulersWith(ctx context.Context, get pdHTTPRequest) ([]st
...
+			return nil, nil
+		}
+		return value, nil
+	}
+	return nil, err
+}
+
+// UpdatePDScheduleConfig updates PD schedule config value associated with the key.
+func (mgr *Mgr) UpdatePDScheduleConfig(
+	ctx context.Context, configKey string, configValue interface{},
```
kennytm: (ditto)

pkg/task/restore.go
```diff
@@ -287,27 +291,89 @@ func filterRestoreFiles(
...
+		if value == nil {
+			// Ignore non-exist config.
+			continue
+		}
+
+		// Speed update PD scheduler by enlarging scheduling limits.
+		// Multiply limits by store count but no more than 40.
+		// Larger limit may make cluster unstable.
+		limit := int(value.(float64))
+		scheduleLimitCfg[cfgKey] = math.Max(40, float64(limit*len(stores)))
```
```suggestion
		scheduleLimitCfg[cfgKey] = math.Min(40, float64(limit*len(stores)))
```

`math.Max` means "no less than 40".

pkg/task/restore.go
```diff
@@ -211,6 +221,7 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 		err = restore.SplitRanges(ctx, client, rangeBatch, rewriteRules, updateCh)
 		if err != nil {
 			log.Error("split regions failed", zap.Error(err))
+			// If any error happened, return now, don't execute checksum.
 			return err
```
kennytm: no need to `restorePostWork`?

overvenus: It's now `defer`ed.

overvenus:
/run-integration-test pd=release-4.0 tikv=release-4.0 tidb=release-4.0

overvenus:
PTAL, thanks!

pkg/task/restore.go
kennytm: @overvenus  But this means the `restorePostWork` is done _after_ checksum

pkg/task/restore.go
```diff
@@ -203,6 +213,7 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 	if err = splitPrepareWork(ctx, client, newTables); err != nil {
 		return err
 	}
+	splitPostWork(ctx, client, newTables)
```
3pointer, line 216: it seems miss defer?

pkg/task/restore.go
```diff
@@ -229,7 +230,13 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 	}
 	// Always run the post-work even on error, so we don't stuck in the import
 	// mode or emptied schedulers
-	defer restorePostWork(ctx, client, mgr, clusterCfg)
+	restorePostWorkOnce := sync.Once{}
+	restorePostWork := func() {
+		restorePostWorkOnce.Do(func() {
+			restorePostWork(ctx, client, mgr, clusterCfg)
+		})
+	}
```
kennytm, line 238:
```suggestion
	shouldRestorePostWork := true
	restorePostWork := func() {
		if shouldRestorePostWork {
			shouldRestorePostWork = false
			restorePostWork(ctx, client, mgr, clusterCfg)
		}
	}
```

no point using `sync.Once{}` since the whole thing is not shared 🤣

overvenus: Right, thanks!

PR #199, opened 2020-03-19
fix incremental bug in llroad test

3pointer:
1. Fix filter ddl bug. in some case that ddls has same tableName but different DB name
2. incremental restore should execute DDL before create Database.
3. `nothing to do` shouldn't return error, because even if there is no sst files, we still need to execute ddl jobs as soon as possible in restore.

### Check List 

Tests 

 - Manual test

### DIFF
--- a/pkg/restore/db.go
+++ b/pkg/restore/db.go
@@ -210,12 +210,13 @@ func FilterDDLJobs(allDDLJobs []*model.Job, tables []*utils.Table) (ddlJobs []*m
 		tableNames[table.Info.Name.String()] = true
 		for _, job := range allDDLJobs {
 			if job.BinlogInfo.TableInfo != nil {
-				if tableIDs[job.TableID] || tableNames[job.BinlogInfo.TableInfo.Name.String()] {
+				name := job.SchemaName + job.BinlogInfo.TableInfo.Name.String()
+				if tableIDs[job.TableID] || tableNames[name] {
 					ddlJobs = append(ddlJobs, job)
 					tableIDs[job.TableID] = true
 					// For truncate table, the id may be changed
 					tableIDs[job.BinlogInfo.TableInfo.ID] = true
-					tableNames[job.BinlogInfo.TableInfo.Name.String()] = true
+					tableNames[name] = true
 				}
 			}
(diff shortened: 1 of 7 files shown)


pkg/restore/db.go
```diff
@@ -210,12 +210,13 @@ func FilterDDLJobs(allDDLJobs []*model.Job, tables []*utils.Table) (ddlJobs []*m
 		tableNames[table.Info.Name.String()] = true
 		for _, job := range allDDLJobs {
 			if job.BinlogInfo.TableInfo != nil {
-				if tableIDs[job.TableID] || tableNames[job.BinlogInfo.TableInfo.Name.String()] {
+				name := job.SchemaName + job.BinlogInfo.TableInfo.Name.String()
```
kennytm: 🤔 

Can we use a `struct { schemaName string; tableName string }` as the key? The current scheme will still break for `"abc" + "de"` vs `"ab" + "cde"`.

3pointer:
> Could you add a test?

already add this test case in llroad cases, so that we can find this error

kennytm:
> already add this test case in llroad cases, so that we can find this error

llroad is still private, is non-trivial to setup, and is outside of this repository. If possible please still include a minimized test case.

3pointer:
> > already add this test case in llroad cases, so that we can find this error
> 
> llroad is still private, is non-trivial to setup, and is outside of this repository. If possible please still include a minimized test case.

ok, I'll add tests when this PR can merge

pkg/restore/db.go
```diff
@@ -207,15 +207,17 @@ func FilterDDLJobs(allDDLJobs []*model.Job, tables []*utils.Table) (ddlJobs []*m
 		tableIDs := make(map[int64]bool)
 		tableIDs[table.Info.ID] = true
 		tableNames := make(map[string]bool)
-		tableNames[table.Info.Name.String()] = true
+		name := fmt.Sprintf("%s:%s", table.Db.Name.String(), table.Info.Name.String())
```
kennytm: though less likely, there is still problem with `:` in the table/db name.

PR #201, opened 2020-03-21
Fix the issue that rawkv restore's end key is incorrectly inclusive when restoring partial of backed up data

MyonKeminta:

This PR fixes the issue that rawkv restore's end key is incorrectly inclusive when restoring partial of backed up data. See https://github.com/tikv/tikv/issues/7163

This PR adds is_raw_kv field to DownloadRequest and adds end_key_exclusive field to SstMeta. For SST files that contains the restoring end key in its range, it will cut the file's end key to the restoring endKey, and set end_key_exclusive.

- [x] kvproto: https://github.com/pingcap/kvproto/pull/581
- [x] tikv: https://github.com/tikv/tikv/pull/7196

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

  - Added is_raw_kv to kvproto `DownlaodRequest`
  - Added end_key_exclusive to kvproto `SstMeta`

Side effects

\-

Related changes

  - release-3.1 and release-4.0

### Release note

* Fix the issue that in rawkv mode, the end key of restoring range may be included, which is expected to be exclusive.


MyonKeminta:
@3pointer @kennytm  PTAL thanks!

pkg/restore/import.go
```diff
@@ -397,9 +397,9 @@ func (importer *FileImporter) downloadRawKVSST(
 	if bytes.Compare(importer.rawStartKey, sstMeta.Range.GetStart()) > 0 {
 		sstMeta.Range.Start = importer.rawStartKey
 	}
-	// TODO: importer.RawEndKey is exclusive but sstMeta.Range.End is inclusive. How to exclude importer.RawEndKey?
-	if len(importer.rawEndKey) > 0 && bytes.Compare(importer.rawEndKey, sstMeta.Range.GetEnd()) < 0 {
+	if len(importer.rawEndKey) > 0 && bytes.Compare(importer.rawEndKey, sstMeta.Range.GetEnd()) <= 0 {
```
3pointer, line 400: Is it possible `sstMeta.Range.GetEnd()` is empty?

MyonKeminta: It's inclusive so I think will not be empty, but I'm not sure. Maybe I'd better check it here.

MyonKeminta:
@3pointer @kennytm PTAL can we merge it?

PR #211, opened 2020-03-27
add skip create sqls

3pointer:
skip create sqls in restore, this may use when user has already use br create tables and databases but broken during restore, because user may have lots of tables, skip retry creates sql will save time.

1.add `skipCreateSQL` and check whether we can skip before `createDatabases` and `createTables`

### Check List 

Tests 
 - Integration test

Related changes

### DIFF
--- a/pkg/task/restore.go
+++ b/pkg/task/restore.go
@@ -23,8 +23,8 @@ import (
 )
 
 const (
-	flagOnline        = "online"
-	flagSkipCreateSQL = "skip-create-sql"
+	flagOnline   = "online"
+	flagNoSchema = "no-schema"
 )
 
 var schedulers = map[string]struct{}{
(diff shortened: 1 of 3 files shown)


pkg/task/restore.go
```diff
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	flagOnline = "online"
+	flagOnline        = "online"
+	flagSkipCreateSQL = "skip-create-sql"
```
kennytm: If we follow Lightning this flag should be named `--no-schema`

pkg/task/restore.go
```diff
@@ -45,13 +46,18 @@ const (
...
+	Online        bool `json:"online" toml:"online"`
+	SkipCreateSQL bool `json:"skip-create-sql" toml:"skip-create-sql"`
 }
 
 // DefineRestoreFlags defines common flags for the restore command.
 func DefineRestoreFlags(flags *pflag.FlagSet) {
 	// TODO remove experimental tag if it's stable
-	flags.Bool("online", false, "(experimental) Whether online when restore")
+	flags.Bool(flagOnline, false, "(experimental) Whether online when restore")
+	flags.Bool(flagSkipCreateSQL, false, "Only allow when restore cluster has already use br create all schema and tables before")
```
```suggestion
	flags.Bool(flagSkipCreateSQL, false, "skip creating schemas and tables, reuse existing empty ones")
```

pkg/task/restore.go
3pointer: change it to --no-schema

kennytm:
```
[2020-03-27T02:59:32.676Z] TEST: [br_db_skip] restore br_db_skip with no-schema succeed

[2020-03-27T02:59:32.676Z] TEST: [br_db_skip] failed!
```

PR #222, opened 2020-03-31
move waiting reject stores in import file

3pointer:
currently we have issue when deal with tiflash stores:
to reproduce it, we can assume
  1. we have two tables, t1 and t2
        * t1 has 100000 record
        * t2 has 10 record
2. alter table `t1` set tiflash replica 1;
3. br backup table --db xxx --table `t2` -s ... --pd ...
4. br restore table --db xxx --table `t2` -s ... --pd ...

then we may get this log until waiting out of time
```
[2020/03/31 17:57:12.714 +08:00] [INFO] [split.go:150] ["start to wait for removing rejected stores"] [rejectStores="{\"92\":true}"]
```

this because `t2` and `t1` may be in the same region. and the condition of waiting rejected store is whether [region has tiflash store peers](https://github.com/pingcap/br/blob/6268cde45e6a05b5e46d1208b31ea18a1770b8ba/pkg/restore/split.go#L292)

move `wait reject store` logic after import file region scan, when we finish split & scatter, the new scan regions will only belong to one table.

### Check List 

Tests 

 - Manual test

Related changes

### DIFF
--- a/pkg/restore/import.go
+++ b/pkg/restore/import.go
@@ -175,7 +175,8 @@ func (importer *FileImporter) SetRawRange(startKey, endKey []byte) error {
 
 // Import tries to import a file.
 // All rules must contain encoded keys.
-func (importer *FileImporter) Import(file *backup.File,
+func (importer *FileImporter) Import(
+	file *backup.File,
 	rejectStoreMap map[uint64]bool,
 	rewriteRules *RewriteRules,
 ) error {
--- a/pkg/restore/import.go
+++ b/pkg/restore/import.go
@@ -198,10 +198,7 @@ func (importer *FileImporter) Import(
 		zap.Binary("startKey", startKey),
 		zap.Binary("endKey", endKey))
 
-	needReject := false
-	if len(rejectStoreMap) > 0 {
-		needReject = true
-	}
+	needReject := len(rejectStoreMap) > 0
 
 	err = utils.WithRetry(importer.ctx, func() error {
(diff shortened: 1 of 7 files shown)


pkg/restore/import.go
```diff
@@ -175,7 +175,10 @@ func (importer *FileImporter) SetRawRange(startKey, endKey []byte) error {
 
 // Import tries to import a file.
 // All rules must contain encoded keys.
-func (importer *FileImporter) Import(file *backup.File, rewriteRules *RewriteRules) error {
+func (importer *FileImporter) Import(file *backup.File,
+	rejectStoreMap map[uint64]bool,
+	rewriteRules *RewriteRules,
+) error {
```
overvenus, line 181:
```suggestion
func (importer *FileImporter) Import(
	file *backup.File,
	rejectStoreMap map[uint64]bool,
	rewriteRules *RewriteRules,
) error {
```

pkg/restore/import.go
```diff
@@ -193,6 +196,12 @@ func (importer *FileImporter) Import(file *backup.File, rewriteRules *RewriteRul
 		zap.Stringer("file", file),
 		zap.Binary("startKey", startKey),
 		zap.Binary("endKey", endKey))
+
+	needReject := false
+	if len(rejectStoreMap) > 0 {
+		needReject = true
+	}
```
kennytm, line 203:
```suggestion
	needReject := len(rejectStoreMap) > 0
```

PR #223, opened 2020-04-03
--checksum=false won't work.

YuJuncen:
The flag --checksum won't work properly, when you specify `--checksum=false`, you will(probably) receive a error message like "failed to checksum."

I have modified the `RunBackup` and `RunRestore` methods so they won't check checksums when disabled.
But for raw kv, I have not done any change yet.
One more thing, maybe use `--skip-checksum` instead of `--checksum=false` will be better, but this will break backward compatibility. 

### Check List

Tests 

 - Unit test
 - Integration test
Code changes

Related changes


pkg/backup/client.go
```diff
@@ -748,8 +748,23 @@ func SendBackup(
...
 }
 
+// ChecksumMatches tests whether the "local" checksum matches the checksum from TiKV.
+func (bc *Client) ChecksumMatches(local []uint64) bool {
+	if len(local) != len(bc.backupMeta.Schemas) {
+		return false
+	}
+
+	for i, schema := range bc.backupMeta.Schemas {
+		if local[i] != schema.Crc64Xor {
```
kennytm, line 758: The checksum should compare all of crc64xor, total bytes and total kvs.

YuJuncen: Yep, I'm dealing with that now.

pkg/backup/client.go
```diff
@@ -748,8 +755,48 @@ func SendBackup(
 	return nil
 }
 
+// ChecksumMatches tests whether the "local" checksum matches the checksum from TiKV.
+func (bc *Client) ChecksumMatches(local []Checksum) bool {
+	if len(local) != len(bc.backupMeta.Schemas) {
```
3pointer, line 760: Perhaps add a log error here is better

pkg/task/restore.go
```diff
@@ -291,15 +291,18 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
...
 
-	// Checksum
-	updateCh = g.StartProgress(
-		ctx, "Checksum", int64(len(newTables)), !cfg.LogProgress)
-	err = client.ValidateChecksum(
-		ctx, mgr.GetTiKV().GetClient(), tables, newTables, updateCh)
-	if err != nil {
-		return err
+	// Don't do checksum if user specified.
+	if cfg.Checksum {
```
3pointer, line 295: sometimes we cannot do`ValidateChecksum` when there is no checksum data store in `backupmeta`, so I think maybe we can add a check requirement, whether we can do `ValidateChecksum` even user specified

YuJuncen: That would be good. Let me try it.

pkg/backup/client.go
YuJuncen: Yes, firstly I think this error shall never happens. (Since we just checked `json.Unmarshal` in `FastChecksum`) So I just omitted them and print a log, but in fact, add a error return value will be a better slouation.

pkg/task/backup.go
```diff
@@ -195,21 +195,24 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
 	if err != nil {
 		return err
 	}
-
-	if cfg.LastBackupTS == 0 {
-		var valid bool
-		valid, err = client.FastChecksum()
+	checksums, err := client.FastChecksum()
```
3pointer, line 198: It seems we still have checksum progress bar here when `cfg.Checksum`=false

YuJuncen: That was done... Even a little complex, but I think that is acceptable.

3pointer: okay，I think we can move L207-L226 to L184-L198, it will looks like
```
if cfg.LastBackupTS == 0 {
...
return
}

if cfg.Checksum {
...
} else {
...
}

Comment (YuJuncen) [reply pkg/task/backup.go]: OK, that was done.

File: pkg/task/backup.go [hunk 5]
```diff
@@ -180,39 +180,35 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
...
-	if err != nil {
-		return err
-	}
-
-	if cfg.LastBackupTS == 0 {
-		var valid bool
-		valid, err = client.FastChecksum()
+	// Checksum from server, and then fulfill the backup metadata.
+	if cfg.Checksum {
+		backupSchemasConcurrency := backup.DefaultSchemaConcurrency
```
3pointer, line 185: you can use `MinInt` here too

YuJuncen: Ok! Let me do it.

pkg/backup/client.go
```diff
@@ -748,8 +755,59 @@ func SendBackup(
...
+		if err != nil {
+			log.Error("failed in fast checksum, and cannot parse table info.")
+			return false, err
+		}
+		if localChecksum.Crc64Xor != schema.Crc64Xor ||
+			localChecksum.TotalBytes != schema.TotalBytes ||
+			localChecksum.TotalKvs != schema.TotalKvs {
+			log.Error("failed in fast checksum",
+				zap.String("database", dbInfo.Name.String()),
+				zap.String("table", tblInfo.Name.String()),
```
kennytm, line 783:
```suggestion
				zap.Stringer("db", dbInfo.Name),
				zap.Stringer("table", tblInfo.Name),
```

pkg/backup/client.go
```diff
@@ -815,3 +865,14 @@ func (bc *Client) CompleteMeta(backupSchemas *Schemas) error {
 	bc.backupMeta.Schemas = schemas
 	return nil
 }
+
+// CopyMetaFrom copies schema metadata directly from pending backupSchemas, without calculating checksum.
+// use this when user skip the checksum generating.
+func (bc *Client) CopyMetaFrom(backupSchemas *Schemas) {
```
kennytm, line 871: why do we need a deep copy here? i.e. why running

```go
bc.backupMeta.Schemas = backupSchemas.schemas
```

isn't sufficient?

pkg/restore/client.go
```diff
@@ -650,6 +650,14 @@ func (rc *Client) ValidateChecksum(
 			workers.Apply(func() {
 				defer wg.Done()
 
+				if table.NoChecksum() {
+					log.Info("table doesn't has checksum, skipping checksum.",
+						zap.String("database", table.Db.Name.L),
+						zap.String("table name", table.Info.Name.L))
```
kennytm, line 656:
```suggestion
					log.Info("table doesn't have checksum, skipping checksum",
						zap.Stringer("db", table.Db.Name),
						zap.Stringer("table", table.Info.Name))
```

pkg/task/backup.go
```diff
@@ -224,6 +219,30 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
...
+	}
+	if cfg.LastBackupTS == 0 {
+		var matches bool
+		matches, err = client.ChecksumMatches(checksums)
+		if err != nil {
+			return err
+		}
+		if !matches {
+			log.Error("backup FastChecksum mismatch!")
+			return errors.Errorf("mismatched checksum")
```
kennytm, line 237:
```suggestion
			return errors.New("mismatched checksum")
```

pkg/backup/client.go
YuJuncen: I'm not sure whether it's OK to just shallow copy(i.e. will the slice be unexpected changed somewhere?), the original version(`schemas.Start()` -> `schemas.finishTableChecksum()` -> `client.CompleteMeta()`) makes a deep copy of the slice, so I just followed it...

pkg/task/backup.go
YuJuncen: nice, thanks.

pkg/backup/client.go
YuJuncen: BTW, I think it won't be costly since the number of schema is usually small... and this method is invoked just once per backup...

Issue #228, opened 2020-04-03
BR cannot incremental restore when met DDL with auto random DDL


3pointer:
1. What did you do?
incremental restore met DDL with auto random would fail, reproduce it with follow step
   1.   Full backup
   2. `create table ta(`a` int(11) NOT NULL /*T!30100 AUTO_RANDOM(5) */, PRIMARY KEY (`a`))`
   3. Incremental backup
   4. Full restore
   5. Incremental restore will fail
```
Invalid auto random: auto_random is an experimental feature, which can only be used when allow-auto-random is enabled. This can be changed in the configuration.
```
2. What did you expect to see?
incremental restore success

3. What did you see instead?
```
Invalid auto random: auto_random is an experimental feature, which can only be used when allow-auto-random is enabled. This can be changed in the configuration.
```

4. What version of BR and TiDB/TiKV/PD are you using?
br 3.1-rc
br 4.0-rc

kennytm:
We should just enable all experiment features which doesn't affect existing behaviors.

PR #230, opened 2020-04-07
enable tidb config by default

3pointer:
Fix issue #228 

Enable integrated TiDB experimental config by default 

### Check List 

Tests 

 - Manual test 
Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


pkg/task/restore.go
```diff
@@ -467,3 +462,18 @@ func RunRestoreTiflashReplica(c context.Context, g glue.Glue, cmdName string, cf
...
 	return nil
 }
+
+func enableTiDBConfig() {
+	// set max-index-length before execute DDLs and create tables
+	// we set this value to max(3072*4), otherwise we might not restore table
+	// when upstream and downstream both set this value greater than default(3072)
+	conf := config.GetGlobalConfig()
+	conf.MaxIndexLength = config.DefMaxOfMaxIndexLength
+	log.Warn("set max-index-length to max(3072*4) to skip check index length in DDL")
```
kennytm, line 472: Perhaps this warning should also mention the experimental features. But I'm not sure about the point of having such warning at all when it's always emitted @5kbpers.

3pointer:
> Add a test case for #228. Rest LGTM.

okay, I'm working on it

kennytm:
One more thing: can you also enable `alter-primary-key = true`? Without this setting the alter-pk related DDLs cannot be executed during incremental restore 😂.

3pointer:
> One more thing: can you also enable `alter-primary-key = true`? Without this setting the alter-pk related DDLs cannot be executed during incremental restore 😂.

okay, I'll enable it and add related test

3pointer:
> LGTM.
> 
> But I'm a bit worried that DDLs are executed before the CREATE TABLEs, meaning that in a _non_-alter-pk server if the incremental backup contained a CREATE TABLE the rows will be wrong.

we should consider to get restore cluster config to fix it, should I file an issue?

kennytm:
Yes that would be the ideal case.

Alternatively we could special case all CREATE TABLE DDLs, such that if a table is successfully constructed, we will DROP that table before calling CreateDatabase(). 

Let's file an issue, and if this breaks non-alter-pk CREATE TABLE probably don't enable alter-pk in BR yet 😢

3pointer:
> Yes that would be the ideal case.
> 
> Alternatively we could special case all CREATE TABLE DDLs, such that if a table is successfully constructed, we will DROP that table before calling CreateDatabase().
> 
> Let's file an issue, and if this breaks non-alter-pk CREATE TABLE probably don't enable alter-pk in BR yet 😢

Even we can get restore cluster config `alter-pk`, we still can't simply enable alter-pk or not. because this config may change at created time and we didn't exactly know the table format. we need to get table format and save it into `backupmeta`, so I think I need file a tidb issue, to provide this format for BR

kennytm:
There is a TableInfo field in the DDL Job which we can use in the next PR.

PR #231, opened 2020-04-08
Use table info

3pointer:
Fix https://github.com/pingcap/br/pull/230#issuecomment-610798266

1. use `ddljob.TableInfo` and `ddljob.DBInfo` to create tables and schema in incremental restore.
2. set `config.AlterPrimaryKey` to true so we can execute drop/add pk DDL in incremental restore.

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/pkg/restore/db.go
+++ b/pkg/restore/db.go
@@ -45,7 +45,24 @@ func NewDB(g glue.Glue, store kv.Storage) (*DB, error) {
 // ExecDDL executes the query of a ddl job.
 func (db *DB) ExecDDL(ctx context.Context, ddlJob *model.Job) error {
 	var err error
-	if ddlJob.BinlogInfo.TableInfo != nil {
+	tableInfo := ddlJob.BinlogInfo.TableInfo
+	dbInfo := ddlJob.BinlogInfo.DBInfo
+	switch ddlJob.Type {
+	case model.ActionCreateSchema:
+		err = db.se.CreateDatabase(ctx, dbInfo)
+		if err != nil {
+			log.Error("create database failed", zap.Stringer("db", dbInfo.Name), zap.Error(err))
+		}
+		return errors.Trace(err)
+	case model.ActionCreateTable:
+		err = db.se.CreateTable(ctx, model.NewCIStr(ddlJob.SchemaName), tableInfo)
+		if err != nil {
+			log.Error("create database failed", zap.Stringer("db", dbInfo.Name), zap.Error(err))
+		}
+		return errors.Trace(err)
+	}
+
+	if tableInfo != nil {
 		switchDbSQL := fmt.Sprintf("use %s;", utils.EncloseName(ddlJob.SchemaName))
 		err = db.se.Execute(ctx, switchDbSQL)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/restore/db.go
```diff
@@ -45,7 +45,27 @@ func NewDB(g glue.Glue, store kv.Storage) (*DB, error) {
...
+	dbInfo := ddlJob.BinlogInfo.DBInfo
+	switch ddlJob.Type {
+	case model.ActionCreateSchema:
+		err = db.se.CreateDatabase(ctx, dbInfo)
+		if err != nil {
+			log.Error("create database failed", zap.Stringer("db", dbInfo.Name), zap.Error(err))
+		}
+		return errors.Trace(err)
+	case model.ActionCreateTable:
+		err = db.se.CreateTable(ctx, model.NewCIStr(ddlJob.SchemaName), tableInfo)
```
kennytm, line 58: we should expose the "IF NOT EXISTS" parameter

tests/br_incompatible_tidb_config/run.sh
```diff
@@ -29,17 +29,30 @@ run_sql "create schema $DB;"
...
+INCREMENTAL_TABLE="t1inc"
 
 run_sql "create table $DB.$TABLE (a int primary key, b int unique);"
 run_sql "insert into $DB.$TABLE values (42, 42);"
 
 # backup
 run_br --pd $PD_ADDR backup db --db "$DB" -s "local://$TEST_DIR/$DB$TABLE"
 
+run_sql "create table $DB.$INCREMENTAL_TABLE (a int primary key, b int unique);"
+run_sql "insert into $DB.$INCREMENTAL_TABLE values (42, 42);"
```
kennytm, line 41: please include a test case for [`CREATE TABLE … LIKE`](https://dev.mysql.com/doc/refman/8.0/en/create-table-like.html).

pkg/restore/db.go
3pointer: I didn't get it, table shouldn't exist if we met create table sql here

kennytm: Ok.

tests/br_incompatible_tidb_config/run.sh
3pointer: okay, add test case below

tests/br_incompatible_tidb_config/run.sh
```diff
@@ -29,17 +29,33 @@ run_sql "create schema $DB;"
...
+
+# drop pk
+run_sql "alter table $DB.$INCREMENTAL_TABLE drop primary key"
+run_sql "drop table $DB.$INCREMENTAL_TABLE"
+run_sql "create table $DB.$INCREMENTAL_TABLE like $DB.$TABLE"
+run_sql "insert into $DB.$INCREMENTAL_TABLE values (42, 42);"
+
+# incremental backup
+run_br --pd $PD_ADDR backup db --db "$DB" -s "local://$TEST_DIR/$DB$INCREMENTAL_TABLE"
```
kennytm, line 50: this doesn't look incremental.

3pointer:
run-integration-test

PR #236, opened 2020-04-16
cmd: don't use ':' in the default log file name

kennytm:

`:` is not a valid filename character for NTFS. The default log file path contains a `:` and thus invalid.

Replace the `:` in the filename by `.`.

### Check List 

Tests 

    * run `bin/br --help` and check the output.

Code changes

Side effects

Related changes


YuJuncen:
LGTM

PR #240, opened 2020-04-20
Restore with lower case

3pointer:
fix the issue that BR not support restore table with Upper table name.

add integration test for Upper table name.

### Check List 

Tests 

 - Integration test

Related changes


GregoryIan:
LGTM

PR #216, opened 2020-03-28
glue: create schema/table directly with info

kennytm:
### What problem does this PR solve? 

Fix #215.

### What is changed and how it works?

Use the new CreateTableWithInfo API from pingcap/tidb#15806 which do not care about the alter-pk setting.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

PR #242, opened 2020-04-22
Support backup & restore views and sequences

### LINKED ISSUE #145: Support backup/restore VIEW
VIEW is an important feature of TiDB. For now, BR only support for independent VIEWs. We need to support more complex VIEWs(depending on other tables and views).

kennytm:

Fix #145, and also support backup/restore sequences.

Thanks to #216, views and sequences can be naturally restored.

The only problems left are

1. Views don't have auto IDs.  This PR skips the auto-ID rebase when the table is a view.
2. Value of sequence needs to be backed up. This PR stores the current sequence value into the `AutoIncID` field, and performs `setval()` on restore.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

    * BR now supports views and sequences.


tests/br_views_and_sequences/run.sh
```diff
@@ -0,0 +1,47 @@
...
+run_sql "create view $DB.view_2 as select * from $DB.view_1;"
+run_sql "create sequence $DB.seq_1 nocache cycle maxvalue 40;"
+run_sql "create table $DB.table_1 (m int primary key default next value for $DB.seq_1, b int);"
+run_sql "insert into $DB.table_1 (b) values (8), (12), (16), (20);"
+run_sql "create sequence $DB.seq_2;"
+run_sql "create table $DB.table_2 (a int default next value for $DB.seq_1, b int default next value for $DB.seq_2, c int);"
+run_sql "insert into $DB.table_2 (c) values (24), (28), (32);"
+run_sql "create view $DB.view_3 as select m from $DB.table_1 union select a * b as m from $DB.table_2 union select m from $DB.view_2;"
+run_sql "drop view $DB.view_1;"
+run_sql "create view $DB.view_1 as select 133 as m;"
```
3pointer, line 30: should we put all views at last during restore, because some view may depend on un-restored table.

kennytm: this doesn't matter for TableInfo-based restore, because the TableInfo already encoded the view's structure (the main reason for the dependency). plus it is fine to drop a view's dependency after it is created.

kennytm:
Sequences actually have 2 associated metadata: the sequence value (`SID:n`) and sequence cycle (`SequenceCycle:n`). There is no exposed SQL to configure the latter value. Is it fine if the number of rounds is wrong (zero) for the cyclic sequence when restored? cc @AilinKid

AilinKid:
> Sequences actually have 2 associated metadata: the sequence value (`SID:n`) and sequence cycle (`SequenceCycle:n`). There is no exposed SQL to configure the latter value. Is it fine if the number of rounds is wrong (zero) for the cyclic sequence when restored? cc @AilinKid

To address this problem, the `cycle` value will influence the nextval computation formula.

there some way manually
if a sequence isn't cycle-type or it hasn't been in the cycle, it is ok to let it be 0.
if a cycle-typed sequence is in the cycle, you can just `setval(MAX_VALUE)`, then use `nextval(seq)` to trigger cycle, then setval(target_num) to set value in cycle round.

3pointer:
> > Sequences actually have 2 associated metadata: the sequence value (`SID:n`) and sequence cycle (`SequenceCycle:n`). There is no exposed SQL to configure the latter value. Is it fine if the number of rounds is wrong (zero) for the cyclic sequence when restored? cc @AilinKid
> 
> To address this problem, the `cycle` value will influence the nextval computation formula.
> 
> there some way manually
> if a sequence isn't cycle-type or it hasn't been in the cycle, it is ok to let it be 0.
> if a cycle-typed sequence is in the cycle, you can just `setval(MAX_VALUE) / setval(MIN_VALUE) if negative`, then use `nextval(seq)` to trigger cycle, then `setval(target_num)` to set value in cycle round.
> 
> now we only just `cycle>0` to know it's in cycle, so there is no difference in 1,2,3... if you don't care the real cycle round times.

I got two questions for this.
1. how to know this sequence value is in cycle round > 0 when backup?

2. I feel like the nextval of this sequence is same no matter how many rounds in cycle, did I miss something?

kennytm:
After discussion with the designers of the sequence function, we conclude that 

* TiDB needs to add two internal APIs for getting and setting the cycle value.
* TiDB's result of `nextval()` after `setval()` depends on the cycle count, which is observable with this:

    ```sql
    create sequence seq start 2 minvalue 1 maxvalue 999999 increment 2 cycle;
    do setval(seq, 10);
    select nextval(seq);
    ```

    In the first cycle, the "base" of the sequence is `start`, so `nextval()` will always return an even number.

    In the future cycles, the "base" of the sequence is `minvalue`, so `nextval()` will always return an odd number.


----


`setval()` is not an SQL Standard function. (ISO/IEC 9075-2 only specified `ALTER SEQUENCE seq RESTART WITH 12345;` with DDL behavior and is blocking, but `setval()` is not blocking). Thus, the behavior of `setval()` is incompatible between different SQL dialects.

| System | setval behavior after cycle | get cycle count | set cycle count |
|--------|-----------------------------|-----------------|-----------------|
| TiDB | Success; rebase on `minvalue` rather than `start` | unimplemented | unimplemented |
| MariaDB | Becomes no-op, unless explicitly setting [the `round` argument](https://mariadb.com/kb/en/setval/) to the cycle count: `setval(seq, 12345, true, 67)` | `select cycle_count from seq;` | `do setval(seq, value, true, @cycle_count)` |
| PostgreSQL | No concept of cycle count, and `setval` will succeed even if the value is decreased | N/A | N/A |
| MS SQL Server / IBM Informix / SQL standard | No `setval()` function; `ALTER SEQUENCE RESTART WITH` is supported and can decrease values | N/A | N/A |
| Oracle | Need to drop and create the sequence to change its value. Boo. | N/A | N/A |

The TiDB behavior is designed to keep the same pace among all nodes within the same cluster. In any case, both MariaDB and TiDB's `setval` disallow decreasing the sequence number, and thus the cycle count must be recorded.

pkg/restore/db.go
```diff
@@ -106,16 +106,52 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
...
-		utils.EncloseName(table.Info.Name.O),
-		table.Info.AutoIncID)
 
-	err = db.se.Execute(ctx, alterAutoIncIDSQL)
+	var restoreMetaSQL string
+	if table.Info.IsSequence() {
+		setValFormat := fmt.Sprintf("do setval(%s.%s, %%d);",
+			utils.EncloseName(table.Db.Name.O),
+			utils.EncloseName(table.Info.Name.O))
+		if table.Info.Sequence.Cycle {
```
AilinKid, line 115: LGTM

3pointer:
manual test success.
```sql
create sequence seq start 2 minvalue 1 maxvalue 999999 increment 2 cycle;
do setval(seq, 10);
select nextval(seq);
```

before backup we got sequence with even number:
```
mysql> select nextval(seq);
+--------------+
| nextval(seq) |
+--------------+
|           12 |
+--------------+
```

after restore we got sequence with odd number:
```
mysql> select nextval(seq);
+--------------+
| nextval(seq) |
+--------------+
|         2013 |
+--------------+
```

PR #243, opened 2020-04-22
pass sse_kms_key_id to S3

yiwu-arbug:
Pass sse_kms_key_id when backup to S3. This is to support using user owned KMS key for server-side encryption.

Pass sse_kms_key_id when backup to S3 to both tikv and when storing backupmeta in S3.

### Check List 

Tests 
Manual test with BR. Try:
* specify kms key id on backup. restore success.
* disable the kms key and try restore again. restore failed.

Related changes

depends on kvproto change: https://github.com/pingcap/kvproto/pull/607 and tikv change: https://github.com/tikv/tikv/pull/7627


yiwu-arbug:
ready for review. not sure why integration test fail before.

PR #246, opened 2020-04-23
storage: support placing the S3/GCS options into the storage URL

kennytm:

pingcap/tidb#15274 requires a function to parse the URL parameters into actual options. For compatibility, we'd like to introduce this feature into BR as well, so you could write e.g.

```sh
./br backup -s 's3://bucket/prefix?region=us-west-2'
```

We introduced a new function `storage.ExtractQueryParameters(url, &options)` to be shared by both TiDB and BR itself for moving query parameters from the URL into the S3/GCS options. We used reflection to avoid repeating the keys. Parameter keys follow the JSON keys in the struct tags.

BTW, added `gs://` as alias of `gcs://`.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

    * Mention that there are now two ways to provide the parameters: `-s 's3://bucket/prefix?region=us-west-2'` vs `-s 's3://bucket/prefix' --s3.region us-west-2`.

### DIFF
--- a/pkg/storage/parse.go
+++ b/pkg/storage/parse.go
@@ -4,6 +4,8 @@ package storage
 
 import (
 	"net/url"
+	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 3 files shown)


YuJuncen:
LGTM

Issue #95, opened 2019-12-09
BR should provide the total size of one backup


DanielZhangQD:
BR should provide the total size of one backup considering following cases:

- Users want to know the size of each backup.
- The platforms that integrate with BR, e.g. TEP, DBaaS, etc. need to show the size of each backup.

Maybe TiKV can return the size of the files it created for each backup to BR and BR calculates the total size from all the TiKVs and writes the size to backup meta.

@kennytm @overvenus @tennix

kennytm:
Include this as part of checksum calculation?

DanielZhangQD:
You mean platforms integrating with BR want to include this as part of checksum? Not necessarily.
They may just need to know the total size of each backup.

overvenus:
> BR should provide the total size of one backup considering following cases:
> ...

That's a good feature request! And it would be easy to implement. What we need to do is,

1. Add a new field `uint64 size = 10` to `backup.File` https://github.com/pingcap/kvproto/blob/5dbce7e7b868a1a19be3a368f9ea6db120ea8926/proto/backup.proto#L42-L54
2. Adjust TiKV backup writer, attach file size to the `File`. https://github.com/tikv/tikv/blob/134c5663340c8e868f23bc2f3273b84168aed862/components/backup/src/writer.rs#L60-L86
3. Sum up all files at BR.

> Include this as part of checksum calculation?

The `total_bytes` in checksum calculation is a logical size, and the actual file content is compressed, so it does not match the file size.

fredchenbj:
I would like to try it ~ If I have some questions, who can I ask for help ? @overvenus

overvenus:
That's so cool! You can always ask me! @fredchenbj

fredchenbj:
ok, this week I would do some work about this.

kennytm:
The total size is already computed in #249. It just needs to printed somewhere to close this issue.

kennytm:
@liangyuanpeng that "total size" is the total uncompressed KV size, not the total (compressed) archive size.

PR #249, opened 2020-04-25
glue,task: let the tasks inform the glue about the BackupTS and Size.

kennytm:

Added some methods to the glue for recording the total backup size and BackupTS for the summary in TiDB.

### Check List 

Tests 

    * Verified with computation result from TiDB.

Code changes

    * New method in Glue.

Side effects

Related changes


overvenus:
It seems resolves https://github.com/pingcap/br/issues/95?

kennytm:
> It seems resolve #95?

Well we could change the gluetikv implementation to log the input if you like 🙂

PR #270, opened 2020-05-09
summary: fix restore files count miss

3pointer:
restore summary didn't count files correctly

add Unitcount for restore unit.

### Check List 

Tests 

 - No code

Related changes

### Release Note

 -No Release note

### DIFF
--- a/pkg/summary/collector.go
+++ b/pkg/summary/collector.go
@@ -103,18 +103,10 @@ func (tc *logCollector) CollectSuccessUnit(name string, unitCount int, arg inter
 
 	switch v := arg.(type) {
 	case time.Duration:
-		if _, ok := tc.successCosts[name]; !ok {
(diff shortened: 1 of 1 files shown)


pkg/summary/collector.go
```diff
@@ -108,6 +108,7 @@ func (tc *logCollector) CollectSuccessUnit(name string, unitCount int, arg inter
 			tc.successUnitCount += unitCount
 		} else {
 			tc.successCosts[name] += v
+			tc.successUnitCount += unitCount
```
kennytm, line 111: so I wonder why can't we just replace the entire lines 106 to 112 by

```go
tc.successCosts[name] += v
tc.successUnitCount += unitCount
```

similar for lines 114 to 118:

```go
tc.successData[name] += v
```

Issue #269, opened 2020-05-09
make concurrency setting more clear


3pointer:
### Describe your feature request related problem:

user may confuse about concurrency setting, because Restore concurrency default value is 128, but Backup concurrency default value is 4.

### Describe the feature you'd like:

1. give a warning when user didn't set proper value of concurrency.

### Describe alternatives you've considered:

1. update document to make it more clearly
> * remove concurrency introduction in document
> * remove rate-limit introduction in document

kennytm:
Everything causing this issue has been fixed. Closing.

PR #276, opened 2020-05-13
restore: set lower bound of batch size

3pointer:
Solve https://github.com/pingcap/br/issues/269

Add warning when user set unproper concurrency.

### Check List 

Tests 

Related changes

### Release Note

 - No Release note

### DIFF
--- a/pkg/task/restore.go
+++ b/pkg/task/restore.go
@@ -4,9 +4,7 @@ package task
 
 import (
 	"context"
-	"fmt"
 	"math"
-	"os"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/kvproto/pkg/backup"
@@ -91,13 +89,6 @@ func (cfg *RestoreConfig) ParseFromFlags(flags *pflag.FlagSet) error {
 
 	if cfg.Config.Concurrency == 0 {
 		cfg.Config.Concurrency = defaultRestoreConcurrency
-	}
-	if cfg.Config.Concurrency < defaultRestoreConcurrency {
-		fmt.Fprintf(os.Stdout,
-			"[warn] concurrency is too low for restore cluster,"+
-				"set concurrency >= %d will get suitable performance,"+
-				"if you want to reduce impact on restore cluster, please consider --rate-limit\n",
-			defaultRestoreConcurrency)
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/task/restore.go
```diff
@@ -86,9 +88,17 @@ func (cfg *RestoreConfig) ParseFromFlags(flags *pflag.FlagSet) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
+
 	if cfg.Config.Concurrency == 0 {
 		cfg.Config.Concurrency = defaultRestoreConcurrency
 	}
+	if cfg.Config.Concurrency < defaultRestoreConcurrency {
```
kennytm, line 95: This sounds too strict. The program will warn you if the concurrency is less than 128, while what this PR prevents is setting the concurrency to 4. The warning limit could be somewhere between 4 and 128.

pkg/task/restore.go
```diff
@@ -86,9 +88,17 @@ func (cfg *RestoreConfig) ParseFromFlags(flags *pflag.FlagSet) error {
...
+		fmt.Fprintf(os.Stdout,
```
kennytm, line 96: why print to stdout?

pkg/task/restore.go
3pointer: I didn't get the point why this PR prevents setting the concurrency to 4?
I think the concurrency lower than `defaultRestoreConcurrency`(128) is not suitable.

pkg/task/restore.go
3pointer: I didn't print it to log, because I think not all user will check log after restore, and when user set unsuitable config to start command, I think we should let them know immediately

pkg/task/restore.go
kennytm: The question is not about "preventing" the concurrency to be 4, it's having the warning when I set concurrency to 40. 

The problem is we're using the same concurrency flag to both BatchSplit and the workerPool for file.Import.

pkg/task/restore.go
kennytm: Stdout is completely hidden when using SQL though.

pkg/task/restore.go
3pointer: what about just set lower bound of concurrency to 128?

3pointer:
> we should update the document

Do you mean remove concurrency introduction in BR documents?
we already have PR merged.https://github.com/pingcap/docs-cn/pull/3065

GregoryIan:
> > we should update the document
> 
> Do you mean remove concurrency introduction in BR documents?
> we already have PR merged.[pingcap/docs-cn#3065](https://github.com/pingcap/docs-cn/pull/3065)

ok, please fix the title of PR, and then merge it

PR #294, opened 2020-05-18
change auto_random columns to bigint.

YuJuncen:
Because of [tidb#17119](https://github.com/pingcap/tidb/pull/17119), current version of `br_incompatible_tidb_config` will fail: 
```
[2020-05-18T02:14:18.885Z] ERROR 8216 (HY000) at line 1: Invalid auto random: auto_random option must be defined on `bigint` column, but not on `int` column
```

Because we use a int column with auto_random...
https://github.com/pingcap/br/blob/c2f193cd09da31f811eae192bf375eda42016ecd/tests/br_incompatible_tidb_config/run.sh#L92

I changed type of columns with auto_random in the test case to bigint.

### Check List 

Tests 

 - Integration test
 - No code

### Release Note

 - No release note

### DIFF
--- a/tests/br_incompatible_tidb_config/run.sh
+++ b/tests/br_incompatible_tidb_config/run.sh
@@ -89,13 +89,13 @@ start_services "$cur"
 TABLE="t3"
 INCREMENTAL_TABLE="t3inc"
 run_sql "create schema $DB;"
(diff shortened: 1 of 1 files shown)


kennytm:
Cheery pick to 3.1 only? No 4.0?

YuJuncen:
> Cheery pick to 3.1 only? No 4.0?

tidb#17119 says it just need to cherry-pick to 3.1, so this follow it. But I think we can pick it to 4.0 too.

PR #311, opened 2020-05-26
*: add version check before start 

### LINKED ISSUE #301: BR add version check at beginning.

Current not all BR version is compatible with all TiDB version, For example, BR 3.1 is not compatible with TiDB 4.0.

A version check when start BR.

### Describe alternatives you've considered:

Use PD to get backup/restore cluster version, if not compatible, then warning in log.

3pointer:
Resolve #301 

Add cluster version check when br start
### Check List 

Tests 

 - Unit test

Related changes

### Release Note

 - Feature: Add version check at beginning in order to avoid version mismatch.


kennytm:
Please include a flag to disable version check (e.g. `--check-requirements=false`).

pkg/utils/version.go
```diff
@@ -43,3 +48,65 @@ func BRInfo() string {
...
 }
+
+var minTiKVVersion *semver.Version = semver.New("3.1.0-beta.2")
+var incompatibleTiKVMajor3 *semver.Version = semver.New("3.1.0")
+var incompatibleTiKVMajor4 *semver.Version = semver.New("4.0.0-rc.1")
+
+func removeV(v string) string {
+	if v == "" {
+		return v
+	}
```
kennytm, line 59:
```suggestion
```

Trimming the `v` from an empty string still results in an empty string.

pkg/utils/version.go
```diff
@@ -43,3 +48,65 @@ func BRInfo() string {
...
+	for _, s := range stores {
+		tikvVersion, err := semver.NewVersion(removeV(s.Version))
+		if err != nil {
+			return err
+		}
+
+		if tikvVersion.Compare(*minTiKVVersion) < 0 {
+			return errors.Errorf("TiKV %s don't support BR, please upgrade cluster to %s",
+				removeV(s.Version), BRReleaseVersion)
```
kennytm, line 81: consider including the IP of the store too.

YuJuncen:
```
[2020-05-26T12:08:45.673Z] Error: None is not in dotted-tri format
```

Maybe PD returns `"None"` at Version Field?

3pointer:
> ```
> [2020-05-26T12:08:45.673Z] Error: None is not in dotted-tri format
> ```

I found it's BRReleaseVersion is None, but CI has build BR with ldflags, after a few attempt to adjust CI, It didn't work well, So I decide to add [skip](https://github.com/pingcap/br/pull/311/files#diff-8eecc8924ff86cb2554e93a8274932d3R62) to work around.

pkg/task/common.go
```diff
@@ -116,6 +118,9 @@ func DefineCommonFlags(flags *pflag.FlagSet) {
 	flags.Uint64(flagRateLimitUnit, utils.MB, "The unit of rate limit")
 	_ = flags.MarkHidden(flagRateLimitUnit)
 
+	flags.Bool(flagCheckRequirement, true,
```
YuJuncen, line 121: Boolean flags seems must use the form `--some=false` to disable(see [here](https://golang.org/pkg/flag/#hdr-Command_line_flag_syntax)), otherwise when it appears, it would be set to `true`.
Maybe `--skip-check-reqirements` here would be better? Since we must use `--check-requirements=false` to skip check requirements if we let the default value be `true` on this flag.

kennytm: IMO this is fine since we've got `--checksum=false` already. 

`--skip-check-requirements` would be a "negative flag".

overvenus:
> > ```
> > [2020-05-26T12:08:45.673Z] Error: None is not in dotted-tri format
> > ```
> 
> I found it's BRReleaseVersion is None, but CI has build BR with ldflags, after a few attempt to adjust CI, It didn't work well, So I decide to add [skip](https://github.com/pingcap/br/pull/311/files#diff-8eecc8924ff86cb2554e93a8274932d3R62) to work around.

You can add ` -ldflags '$(LDFLAGS)'` to the `build_for_integration_test` target in Makefile.

overvenus:
Please remove .DS_Store files https://github.com/pingcap/br/blob/1a825f01f7e61fc0ac6b3d2408de7b75116d89f3/.DS_Store

Rest LGTM

PR #315, opened 2020-05-27
Fix row data lost

3pointer:
Try fix row data lost.

row id bigger than 9151314442816847872 after encoded may exceed than default max range end [`0xff`]( https://github.com/pingcap/br/blob/943e502cc14da6bff81d15b78d2468500526a44b/pkg/restore/util.go#L106), change it `10*[0xff]`
### Check List 

Tests 

 - Integration test
 - manual test

### Release Note

 - Fix the row data lost when row_id bigger than 9151314442816847872

### DIFF
--- a/pkg/restore/util.go
+++ b/pkg/restore/util.go
@@ -103,8 +103,13 @@ func GetSSTMetaFromFile(
 	if bytes.Compare(rangeStart, region.GetStartKey()) < 0 {
 		rangeStart = region.GetStartKey()
 	}
-	suffix := []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}
-	rangeEnd := append(append([]byte{}, regionRule.GetNewKeyPrefix()...), suffix...)
+	rangeEnd := append([]byte{}, regionRule.GetNewKeyPrefix()...)
+	for i := len(rangeEnd)-1; i >= 0; i -- {
+		if rangeEnd[i] != 0xff {
+			rangeEnd[i] += 1
+			break
+		}
+	}
 	// rangeEnd = min(rangeEnd, region.EndKey)
 	if len(region.GetEndKey()) > 0 && bytes.Compare(rangeEnd, region.GetEndKey()) > 0 {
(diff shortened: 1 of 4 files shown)


GregoryIan:
I have two questions @3pointer 

* This bug will only restore less data, not corrupt other data, right?
* Does append 10x 0xff have any other effects, such as for index key which is not a fixed-length

pkg/restore/util.go
```diff
@@ -103,11 +103,21 @@ func GetSSTMetaFromFile(
 	if bytes.Compare(rangeStart, region.GetStartKey()) < 0 {
 		rangeStart = region.GetStartKey()
 	}
-	rangeEnd := append(append([]byte{}, regionRule.GetNewKeyPrefix()...), 0xff)
+
+	// Append 10 * 0xff to make sure rangeEnd cover all file key
+	// If choose to regionRule.NewKeyPrefix + 1, it may cause WrongPrefix here
+	// https://github.com/tikv/tikv/blob/master/components/sst_importer/src/sst_importer.rs#L221
```
overvenus, line 109:
```suggestion
	// https://github.com/tikv/tikv/blob/970a9bf2a9ea782a455ae579ad237aaf6cb1daec/components/sst_importer/src/sst_importer.rs#L221
```

pkg/restore/util.go
```diff
@@ -103,11 +103,21 @@ func GetSSTMetaFromFile(
...
+	suffix := []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}
+	rangeEnd := append(append([]byte{}, regionRule.GetNewKeyPrefix()...), suffix...)
 	// rangeEnd = min(rangeEnd, region.EndKey)
 	if len(region.GetEndKey()) > 0 && bytes.Compare(rangeEnd, region.GetEndKey()) > 0 {
 		rangeEnd = region.GetEndKey()
 	}
+
+	log.Debug("Get sstMeta",
```
overvenus, line 117:
```suggestion
	log.Debug("get sstMeta",
```

GregoryIan:
LGTM

PR #339, opened 2020-06-08
restore: replace ErrGRPC by recognizing raw gRPC errors

### LINKED ISSUE #333: Network failure during Ingest cannot be retried
Currently BR considers all gRPC error from an Ingest Request to be non-retryable:

https://github.com/pingcap/br/blob/a6643a811dbe726d32a50bb3822a3be33810b4f1/pkg/restore/import.go#L277-L279

Thus, if there is spurious network error dropping the gRPC connection, BR restore will immediately fail. 

We should change it to allow retrying on network failure.

kennytm:

Fix #333.

The gRPC error from `Ingest` is returned raw. The raw gRPC error is not recognized by `utils.WithRetry`, so any gRPC error from ingest will fail immediately.

This PR:

1. Gets rid of the `ErrGRPC` error, and instead just use `status.Codes` to check if the error is really a gRPC error. Thus gRPC errors from IngestSST would also be covered.
2. Restricts the retryable gRPC errors to only "Aborted" and "Unavailable", as described in https://pkg.go.dev/google.golang.org/grpc/codes?tab=doc#Code. Thus legit "Unknown" error wouldn't be retried.

(Potential problem: currently the entire import process would restart from scratch, rather than just the IngestSST RPC.)

### Check List 

Tests 

 - Unit test
    * (DBaaS is going to re-test using this PR)

Code changes

Side effects

Related changes

### Release Note

 - Improved robustness: spurious network failure during ingestion would now be retried.

### DIFF
--- a/pkg/restore/backoff.go
+++ b/pkg/restore/backoff.go
@@ -8,6 +8,8 @@ import (
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/pingcap/br/pkg/utils"
 )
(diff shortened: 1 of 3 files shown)


kennytm:
/merge

CI issue.

kennytm:
https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/1885/pipeline/73

PR #342, opened 2020-06-09
fix version check with hash version

3pointer:
Fix version check when version has hash suffix.

remove version hash to compare version

### Check List 

Tests 

 - Unit test

Related changes

### Release Note

 - No release note


pkg/utils/version.go
```diff
@@ -53,13 +54,18 @@ var minTiKVVersion *semver.Version = semver.New("3.1.0-beta.2")
 var incompatibleTiKVMajor3 *semver.Version = semver.New("3.1.0")
 var incompatibleTiKVMajor4 *semver.Version = semver.New("4.0.0-rc.1")
 
-func removeV(v string) string {
+func removeVAndHash(v string) string {
+	hash := regexp.MustCompile("-[0-9]+-g[0-9a-f]{7,}")
```
kennytm, line 58: move this outside of the function (make it a global)

pkg/utils/version.go
```diff
@@ -53,13 +54,18 @@ var minTiKVVersion *semver.Version = semver.New("3.1.0-beta.2")
...
+	if hash.Match([]byte(v)) {
+		v = hash.ReplaceAllLiteralString(v, "")
+	}
```
kennytm, line 61:
```suggestion
	v = hash.ReplaceAllLiteralString(v, "")
```

pkg/restore/client.go
```diff
@@ -504,6 +504,7 @@ func (rc *Client) RestoreFiles(
 			return err
 		}
 	}
+	wg.Wait()
```
overvenus, line 507: This is a bugfix?

3pointer: yes, I think it fix the updateCh send on close channel at [L490](https://github.com/pingcap/br/blob/1cda453e3feb265b223eb032225254b1ff3d5ab8/pkg/restore/client.go#L490), when last file import success, if will wait until updateCh.Inc finish.

PR #349, opened 2020-06-12
Fix some issue in qa test

3pointer:
fix some issue found in test.

1. check backupTs > lastBackupTs 
2. set upper bound of backup concurrency otherwise it will cause tikv panic.
3. set `download file skip` log to warn.

### Check List 

Tests 

### Release Note

 - No release note


pkg/task/backup.go
```diff
@@ -35,6 +35,7 @@ const (
 	flagGCTTL = "gcttl"
 
 	defaultBackupConcurrency = 4
+	defaultBackupMaxConcurrency = 128
```
kennytm, line 38:
```suggestion
	maxBackupConcurrency = 128
```

128 sounds too small, as 128-core CPU is quite feasible.

pkg/task/backup.go
```diff
@@ -35,6 +35,7 @@ const (
 	flagGCTTL = "gcttl"
 
 	defaultBackupConcurrency = 4
+	maxBackupConcurrency     = 1024
```
overvenus, line 38: This is too large, it will spawn 1024 thread, it could fail due to no resource (can not create threads). What about 128?

3pointer:
failed due to `dial tcp: lookup goproxy.onetool.net on 10.233.0.10:53: no such host`

Issue #328, opened 2020-06-02
BR check backup directory is empty before backup start


3pointer:
### Describe your feature request related problem:
User may backup to same directory when last backup failed and we only check `backupmeta` before backup start.

### Describe the feature you'd like:

Check backup directory is empty before start.

kennytm:
Empty means the directory has no content, or just does not contain `backupmeta`, `backupmeta.json` and `*.sst`?

3pointer:
> Empty means the directory has no content, or just does not contain `backupmeta`, `backupmeta.json` and `*.sst`?

I think  just check backupmeta and *.sst is enough.

PR #358, opened 2020-06-16
br: add lock file

3pointer:
Fix https://github.com/pingcap/br/issues/328
Add a lock file at beginning, to make sure other backup task won't be able to start backup.

Add a lock file when storage created.

### Check List 

Tests 

 - Integration test

### Release Note

 - Fix the issue that fast fail when multiple backup task start backup on same path.

### DIFF
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -122,6 +122,11 @@ func (bc *Client) GetTS(ctx context.Context, duration time.Duration, ts uint64) 
 	return backupTS, nil
 }
 
+// SetLockFile set write lock file.
+func (bc *Client) SetLockFile(ctx context.Context) error {
+	return bc.storage.Write(ctx, utils.LockFile, []byte{})
+}
+
 // SetGCTTL set gcTTL for client.
 func (bc *Client) SetGCTTL(ttl int64) {
 	bc.gcTTL = ttl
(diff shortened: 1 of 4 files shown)


3pointer:
Maybe we should move `backupmeta` check, since it not necessary from now.

kennytm:
cc #151 the previous attempt.

kennytm:
LGTM but the file name sounds odd.

Unlike a typical lock file, we do not delete `backup.lock` after backup has completed.

3pointer:
> LGTM but the file name sounds odd.
> 
> Unlike a typical lock file, we do not delete `backup.lock` after backup has completed.

yes, if we want to delete file we need add `delete` method for external storage, I wonder is it necessary, because it does make sense even one backup finished, other backup tasks still won't be able to use same path. maybe change lock file to another name is better, do you have any advise?

pkg/backup/client.go
```diff
@@ -122,6 +122,11 @@ func (bc *Client) GetTS(ctx context.Context, duration time.Duration, ts uint64)
 	return backupTS, nil
 }
 
+// SetLockFile set write lock file.
+func (bc *Client) SetLockFile(ctx context.Context) error {
+	return bc.storage.Write(ctx, utils.LockFile, []byte{})
```
overvenus, line 127: Could you write some warnings to the file? Eg. DO NOT DELETE and explain what the file does.

PR #367, opened 2020-06-18
tests: add retry on start services

YuJuncen:
BR unit tests sometimes will fail because cannot start cluster(sometimes TiKV, sometimes TiFlash).

- Change a new way to check whether TiFlash is started.
- Add a three times retry if fail to start test cluster.
- Change TiFlash config so it won't fail to synchronize.

### Check List 

Tests 

 - No code (Tests should be tested by what?)

### Release Note

 - No release note

### DIFF
--- a/tests/_utils/run_services
+++ b/tests/_utils/run_services
@@ -40,6 +40,19 @@ stop_services() {
 
 
 start_services() {
+    max_retry=3
+    for i in $(seq 1 $max_retry); do
+        if start_services_impl $@; then
+            return 0
+        fi
+        echo "Failed to start services, but let retry it after $(( $i * 30 )) seconds"
+        sleep $(( $i * 30 ))
+    done
+    echo "Failed to start services after retry $max_retry times."
+    return 1
+}
+
+start_services_impl() {
     stop_services
     source tests/_utils/make_tiflash_config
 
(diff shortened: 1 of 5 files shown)


YuJuncen:
BTW, this might only alleviate but not to cure the fail of CIs.

Maybe we need to 'merge' some cases into one, so that we can start less clusters. At the same time, we get less possibility to fail.

kennytm:
(Fix CLA)

YuJuncen:
/run-all-tests

Is that stable enough?

tests/_utils/run_services
```diff
@@ -27,6 +27,7 @@ TIKV_ADDR="127.0.0.1:2016"
 TIKV_STATUS_ADDR="127.0.0.1:2018"
 TIKV_COUNT=3
 TIFLASH_STATUS="127.0.0.1:17000"
```
kennytm, line 29: remove this line?

YuJuncen: `$TIFLASH_STATUS` would still be used at make_tiflash_config:

https://github.com/pingcap/br/blob/56e151ad8b672272772cd2e37c6c26b148d9942e/tests/_utils/make_tiflash_config#L16

I think maybe leave it here would be better?

kennytm: Ok.

PR #371, opened 2020-06-22
Cherry-pick #367 and #358 to 3.1

YuJuncen:
NOTE: because #367 requires #358, so I cherry-picked them together.

#358  
---

Fix https://github.com/pingcap/br/issues/328
Add a lock file at beginning, to make sure other backup task won't be able to start backup.

Add a lock file when storage created.

### Check List 

Tests 

 - Integration test

### Release Note

 - Fix the issue that fast fail when multiple backup task start backup on same path.

---

#367 

---

BR unit tests sometimes will fail because cannot start cluster(sometimes TiKV, sometimes TiFlash).

- Change a new way to check whether TiFlash is started.
- Add a three times retry if fail to start test cluster.
- Change TiFlash config so it won't fail to synchronize.

### Check List 

Tests 

 - No code (Tests should be tested by what?)

### DIFF
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -120,7 +120,14 @@ func (bc *Client) GetTS(ctx context.Context, duration time.Duration, ts uint64) 
 	return backupTS, nil
 }
 
-// SetStorage set ExternalStorage for client
+// SetLockFile set write lock file.
+func (bc *Client) SetLockFile(ctx context.Context) error {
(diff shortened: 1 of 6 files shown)


YuJuncen:
@3pointer PTAL(the #358 part)

kennytm:
@3pointer CLA needs to be signed again for the bot 🥺

PR #372, opened 2020-06-22
task: allow start pprof at runtime by unix signal

### LINKED ISSUE #350: Add a way to dynamically turn on the PProf port

There has been more than once on DBaaS that BR was started but both status-port and ptrace are disabled, rendering the living environment useless for debugging.

There should be a way to force-setup a pprof server even when --status-addr was not turned on in the beginning, perhaps with SIGUSR1.

### Teachability, Documentation, Adoption, Migration Strategy:

This is mainly an internal debugging feature. We could mention it in the troubleshooting docs, but no need to explain much in details.

YuJuncen:
fix #350 

We listen on Unix signal SIGUSR1(10), and start pprof server when received this signal.

### Check List 

Tests 

 - Integration test
 - Manual test
    We have tested listening a fixed port by failpoint, and I manually tested whether it can bind to a OS allocated port as expect.
图片
图片

### Release Note

 - Now, we can start pprof server dynamically by sending SIGUSR1 to BR.


kennytm:
The implementation is a strange. I'd expect 

1. the HTTP server should listen on `":"` after receiving SIGUSR1, so port collision will never happen. you could obtain the new real port using Listener and then Serve.

    ```go
	l, err := net.Listen("tcp", ":")
	if err != nil {
		panic(err)
	}
	fmt.Println(l.Addr())
	// then do http.Serve(l, nil)
    ```

    you may use failpoint for a deterministic port for testing.

2. the second SIGUSR1 should be ignored. perhaps just printing out which port we're currently listening on.

pkg/utils/pprof.go
```diff
@@ -0,0 +1,57 @@
...
+	mu.Lock()
+	defer mu.Unlock()
+
+	if startedPProf != "" {
+		log.Warn("Try to start pprof when it has been started, nothing will happen", zap.String("address", startedPProf))
+		return
+	}
+
+	go func() {
+		_ = pprof.Handler
```
kennytm, line 33: 🤔 

let use the proper way make sure pprof is registered:

```go
import ( 
    ...
    // register HTTP handler for /debug/pprof
    _ "net/http/pprof"
)
```

pkg/utils/pprof.go
```diff
@@ -0,0 +1,57 @@
...
+	"sync"
+
+	"github.com/pingcap/failpoint"
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+)
+
+var (
+	startedPProf = ""
+	mu           = new(sync.Mutex)
```
kennytm, line 19:
```suggestion
	mu           sync.Mutex
```

pkg/utils/dyn_pprof_unix.go
```diff
@@ -0,0 +1,40 @@
...
+	signalChan = make(chan os.Signal, 1)
+)
+
+// StartDynamicPProfListener starts the listener that will enable pprof when received `startPProfSignal`.
+func StartDynamicPProfListener() {
+	signal.Notify(signalChan, startPProfSignal)
+	go onSignalStartPProf(signalChan)
+	log.Info(fmt.Sprintf("dynamic pprof started, you can enable pprof by `kill -s %d %d`)",
+		startPProfSignal,
+		os.Getpid()))
```
kennytm, line 30: IMO we don't need to advertise this trick to all users.

```suggestion
```

pkg/utils/pprof.go
```diff
@@ -0,0 +1,57 @@
...
+		_ = pprof.Handler
+		if len(statusAddr) != 0 {
+			mu.Lock()
+			failpoint.Inject("determined-pprof-port", func(v failpoint.Value) {
+				port := v.(int)
+				statusAddr = fmt.Sprintf(":%d", port)
+				log.Info("injecting failpoint, pprof will start at determined port", zap.Int("port", port))
+			})
+			listener, err := net.Listen("tcp", statusAddr)
+			log.Info("bind pprof to addr", zap.String("addr", listener.Addr().String()))
```
kennytm, line 42: please also print the address to stderr.

pkg/utils/pprof.go
```diff
@@ -0,0 +1,57 @@
...
+			if err != nil {
+				log.Warn("failed to start pprof", zap.String("addr", statusAddr), zap.Error(err))
+			}
+			startedPProf = listener.Addr().String()
+			mu.Unlock()
```
kennytm, line 47: move these outside the goroutine. only `http.Serve` should be inside the goroutine.

possible race condition scenario:

1. user sent SIGUSR1, `startedPProf == ""`, good.
2. scheduled the goroutine A, which did not start immediately.
3. user sent another SIGUSR1, `startedPProf` is still `""`, good.
4. scheduled another goroutine B.
5. Goroutine A starts to run, and assigned `startedPProf = "10.11.12.13:12345"`.
6. Goroutine B starts to run, since we don't check `startedPProf` again, we reassigned `startedPProf = "10.11.12.13:12346"`.
7. Now we have two pprof servers running 😕.

pkg/utils/dyn_pprof_unix.go
YuJuncen: Well, how about change it to DEBUG level?

pkg/utils/pprof.go
YuJuncen: Oops... Thanks.

pkg/utils/pprof.go
YuJuncen: The original version of starting pprof uses this. Seems if we use import for side-effect, `gosec` goes unhappy...

`G108: Profiling endpoint is automatically exposed on /debug/pprof (gosec)`

YuJuncen: Seems [add a `nosec` pragma](https://github.com/securego/gosec/issues/418) would be a solution for this.

pkg/utils/dyn_pprof_unix.go
kennytm: Debug is OK but still not useful. I don't see any scenario where you can read the log and perform `kill -USR1 $PID` but cannot `ps aux` to get the PID.

pkg/utils/dyn_pprof_unix.go
```diff
@@ -0,0 +1,40 @@
...
+	signalChan = make(chan os.Signal, 1)
+)
+
+// StartDynamicPProfListener starts the listener that will enable pprof when received `startPProfSignal`.
+func StartDynamicPProfListener() {
+	signal.Notify(signalChan, startPProfSignal)
+	go onSignalStartPProf(signalChan)
+	log.Debug("dynamic pprof enabled",
+		zap.Stringer("enableBySignal", startPProfSignal),
+		zap.Int("ppid", os.Getppid()),
```
kennytm, line 29: why `Getppid()` (which produces the PID of the parent process)?

pkg/utils/pprof.go
```diff
@@ -0,0 +1,56 @@
...
+		statusAddr = fmt.Sprintf(":%d", port)
+		log.Info("injecting failpoint, pprof will start at determined port", zap.Int("port", port))
+	})
+	listener, err := net.Listen("tcp", statusAddr)
+	log.Info("bind pprof to addr", zap.Stringer("addr", listener.Addr()))
+	_, _ = fmt.Fprintf(os.Stderr, "bind pprof to addr %s\n", listener.Addr().String())
+	if err != nil {
+		log.Warn("failed to start pprof", zap.String("addr", statusAddr), zap.Error(err))
+	}
+	startedPProf = listener.Addr().String()
```
kennytm, line 45: move these before the log.

pkg/utils/dyn_pprof_unix.go
YuJuncen: Seems `Getpid()` will give us the pid of the carrier thread of this goroutine at my computer (Linux 5.7.5-arch1-1), which (IMO) might change during execution.

In practice, I use `pkill -10 -P $PPID` to send the signal... Or the pprof won't be started. :(

Any better idea here if possible?

YuJuncen: Since this might be useless and misleading, maybe just remove it would be a better idea...

pkg/utils/pprof.go
YuJuncen: Yep, maybe we also need a `return` at the error branch.

pkg/utils/pprof.go
```diff
@@ -0,0 +1,57 @@
...
+		log.Info("injecting failpoint, pprof will start at determined port", zap.Int("port", port))
+	})
+	listener, err := net.Listen("tcp", statusAddr)
+	if err != nil {
+		log.Warn("failed to start pprof", zap.String("addr", statusAddr), zap.Error(err))
+		return
+	}
+	log.Info("bound pprof to addr", zap.Stringer("addr", listener.Addr()))
+	_, _ = fmt.Fprintf(os.Stderr, "bound pprof to addr %s\n", listener.Addr().String())
+	startedPProf = listener.Addr().String()
```
kennytm, line 46:
```suggestion
	startedPProf = listener.Addr().String()
	log.Info("bound pprof to addr", zap.String("addr", startedPProf))
	_, _ = fmt.Fprintf(os.Stderr, "bound pprof to addr %s\n", startedPProf)
```

PR #377, opened 2020-06-23
Speed up DDL execution by send DDL concurrently

YuJuncen:
Currently, DDLs are send to TiDB cluster sequently, if we were the DDL owner, that is fine: we can execute this DDL immediately, and return very fast.

But we are not, and probably cannot. Then things getting bad, we **must** block and waiting our DDL job pushed to the queue, and executed by owner, then we can send next DDL. Even during waiting time, we can push more DDLs into the DDL job queue.

This PR make `GoCreateTabels` send create table jobs into DDL queue concurrently.

we change `GoCreateTables` and make it use below strategy to create tables:
1. if provided a `dbPool`, use this DB pool to execute DDLs concurrently.
2. if not, roll back to old version: send DDL sequentially.

### Check List 

Tests 

 - Integration test (`br_300_small_tables`)
 - Manual test
    We test it locally, by a 300 table, per table 100 records workload:
```bash
for i in $(seq 0 $1); do
    (echo "CREATE TABLE FOO.sbtest$i(k int primary key, v varchar (255), trailling varchar(1024))" | mysql -P $MYSQL_PORT -h $MYSQL_SERVER -u root &&
        echo "INSERT INTO FOO.sbtest$i(k, v, trailling) VALUES `make_values 100`" | mysql -P $MYSQL_PORT -h $MYSQL_SERVER -u root) 
done
```

With different concurrency, the result at my computer is:

```
test-result/1_concurrency
621.19 real        11.08 user         9.64 sys

test-result/4_concurrency
153.06 real         7.50 user         5.97 sys

test-result/8_concurrency
89.42 real         6.48 user         4.97 sys

test-result/16_concurrency
62.90 real         6.47 user         4.82 sys

test-result/baseline (master branch)
605.30 real        11.13 user         9.59 sys
```

### Release Note

 - Speed up restore.

### More Things

- Currently, we make the concurrency of sending DDL the same as `cfg.concurrency`, which sometimes may be too big and will make many transaction conflicts. Since the execution time of a DDL has no much relative to environment, maybe a fixed value(like 16 or 32) would be good?


pkg/restore/db.go
```diff
@@ -37,6 +44,23 @@ func NewDB(g glue.Glue, store kv.Storage) (*DB, error) {
...
 	}
+	return se, nil
+}
+
+func (db *DB) contextualSession(ctx context.Context) glue.Session {
+	if session, ok := ctx.Value(SessionInContext).(glue.Session); ok {
+		return session
+	}
+	return db.se
+}
```
kennytm, line 55: ok i don't understand why the change is at the Session level. why create a pool of Session rather than a pool of DB?

pkg/task/restore.go
```diff
@@ -188,7 +188,19 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 
 	// We make bigger errCh so we won't block on multi-part failed.
 	errCh := make(chan error, 32)
-	tableStream := client.GoCreateTables(ctx, mgr.GetDomain(), tables, newTS, errCh)
+	sessionPool := make([]glue.Session, 0, cfg.Concurrency)
```
kennytm, line 191: the pool size is 128 by default. good for tidb?

pkg/restore/db.go
YuJuncen: Just because this is the method to implement isolated sessions that I figured out :|

You're right.

pkg/task/restore.go
YuJuncen: Yes, I think so.

> Since the execution time of a DDL has no much relative to environment, maybe a fixed value(like 16 or 32) would be good?

pkg/restore/util.go
```diff
@@ -142,6 +142,19 @@ func GetSSTMetaFromFile(
 	}
 }
 
+// MakeSessionPool makes a session pool with specficated size by sessionFactory.
+func MakeSessionPool(size uint, sessionFactory func() (*DB, error)) ([]*DB, error) {
```
kennytm, line 146:
```suggestion
func MakeDBPool(size uint, sessionFactory func() (*DB, error)) ([]*DB, error) {
```

pkg/restore/client.go
```diff
@@ -355,11 +355,24 @@ func (rc *Client) CreateTables(
...
+func (rc *Client) createTable(
+	ctx context.Context,
+	db *DB,
+	dom *domain.Domain,
+	table *utils.Table,
+	newTS uint64,
+) (CreatedTable, error) {
+	if db == nil {
+		db = rc.db
+	}
```
kennytm, line 367: can we remove `rc.db`, or make the DB pool `[]*DB` a field of `rc *Client`?

YuJuncen: `rc.db` still be used for other DDLs (e.g. `CREATE DATABASE`, `ALTER TABLE SET TIFLASH_REPLICA = x`, incremental DDL execution, etc.), if we want to remove it, we must make many changes. (BTW It's seems a little strange if these uses db like `rc.dbs[n]` outside `GoCreateTables`, there is an array but we just need one of them at anytime, since those DDLs  are executed sequential.).

Maybe do that at another refactoring PR would be better?

kennytm: we could have a `rc.GetDB()` and rotate the db list on every call 🙃.

anyway, as we decide to refactor in the future, please add a TODO on the `db` field.

pkg/restore/util.go
```diff
@@ -142,6 +142,19 @@ func GetSSTMetaFromFile(
 	}
 }
 
+// MakeDBPool makes a session pool with specficated size by sessionFactory.
+func MakeDBPool(size uint, sessionFactory func() (*DB, error)) ([]*DB, error) {
```
kennytm, line 146:
```suggestion
func MakeDBPool(size uint, dbFactory func() (*DB, error)) ([]*DB, error) {
```

pkg/restore/util.go
```diff
@@ -142,6 +142,19 @@ func GetSSTMetaFromFile(
...
+	sessionPool := make([]*DB, 0, size)
```
kennytm, line 147:
```suggestion
	dbPool := make([]*DB, 0, size)
```

pkg/restore/util.go
```diff
@@ -142,6 +142,19 @@ func GetSSTMetaFromFile(
...
+	for i := uint(0); i < size; i++ {
+		session, e := sessionFactory()
+		if e != nil {
+			return nil, e
```
kennytm, line 151: is it supposed to return the `dbPool` here, even on error? the warning in `restore.go:201` suggests so.

pkg/task/restore.go
```diff
@@ -188,7 +189,21 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
...
+	// So these jobs won't be faster or slower when machine become faster or slower,
+	// hence make it a fixed value would be fine.
+	dbPool, err := restore.MakeDBPool(defaultDDLConcurrency, func() (*restore.DB, error) {
+		return restore.NewDB(g, mgr.GetTiKV())
+	})
+	if err != nil {
+		log.Warn("create session pool failed, we will send DDLs only by created sessions",
+			zap.Error(err),
+			zap.Int("sessionCount", len(dbPool)),
+		)
```
kennytm, line 204: (here)

pkg/restore/client.go
```diff
@@ -408,16 +424,45 @@ func (rc *Client) GoCreateTables(
...
+				if err := createOneTable(dbPool[selectedDB], t); err != nil {
+					errCh <- err
+					return
+				}
+			})
+		}
+	}
+
 	go func() {
+		wg := new(sync.WaitGroup)
```
kennytm, line 449: i'd suggest we use a `errgroup` here so if one table errors, all other tables would be canceled too.

add a TODO if it's not convenient to do it here.

YuJuncen: Because `startWork` is asynchronous (i.e. it returns its error by sending it to `errCh`), and cannot be synchronized since tasks executing `worker` is asynchronous. If we did that, the style of our code would be like:

```go
innerErrCh = make(chan error, 1)
startWork = // some func() error

errgroup.Go(func() { 
  worker.Apply(func() { innerErrCh <- startWork() })
  return <- innerErrCh
}
```

Which (IMO) is a little dirty... In fact, when we fire issue error to `errCh`, `RunRestore` should stop execution and cancel all works in the pipeline (even sometimes with confusing error message :( ).

pkg/restore/client.go
YuJuncen: But how `GetDB()` should be implemented? If we trivially return `rc.dbs[0]`, seems we cannot take any good from it... Because we still need to inject some 'id' to every call if we need to use other dbs (hence we can't remove the extra parameter of `createTable`), and we use just one db at once (hence we can't speed up execution outside `restoreTables`).

I think maybe we need farther discussion to make sure whether we need to prune the `db` field... :thinking:

kennytm: @YuJuncen round-robin?

```go
db := rc.dbs[rc.curDB]
rc.curDB++
if rc.curDB >= len(rc.dbs) {
    rc.curDB = 0
}
return db
```

YuJuncen: Yes, that's a good idea, but seem here would still be some risk if we want to use this algorithm to allocate session for processes concurrently running, that is, when some session stuck:

Assume we have two sessions, and two processes.

```
(Process 0) |(takes dbs[0]) -------------------------------------------------------------------->(done)|
(Process 1)      |(takes dbs[1])----->(done)|    |(takes dbs[0], bomb, race condition!)*
```

Maybe this cannot provide more for us... Our questions still present: we cannot remove the extra parameter of `createTable`, and out of `createTable`, we cannot take good of multi-dbs too, since they are executed sequentially.

YuJuncen:
/cherry-pick release-4.0

Can @ti-srebot get that done...?

PR #384, opened 2020-06-28
ignore exists table when restore tables

### LINKED ISSUE #365: Do not create irrelevant tables when use incremental restore

1. What did you do?
  a.  we have a cluster with 100+ table and full backup.
  b.  create a new table then do incremental backup.
  c.  restore full backup to new cluster
  d.  restore incremental backup,

   step d. will do lots of create table ddl according to log and cost lots of time. I believe that is unnecessary. 

3. What did you see instead?
   step d. only create newly table.

all

3pointer:
1. Resolve #365 

1. use `dom.InfoSchema()` to check whether db and tables already exists in target cluster. to skip execute unnecessary DDL.
2. record `backupts` in summary log to enhance usability.

### Check List 

Tests 

From below log, we can tell the increment restore will skip 
1. exists databases and tables created by full restore.
2. record DDL job during increment backup.
```
[client.go:538] ["execute ddl query"] [db=test] [query="create table t2(id int)"] [historySchemaVersion=102]
[restore.go:633] ["detective exists schemas to skip"] [db="[test]"] [table="[t,t2,warehouse,district,new_order,item,history,orders,customer,stock,order_line]"]
```

Code changes

Related changes

### Release Note

 - Speed up schema prepare during incremental restore.

### DIFF
--- a/pkg/task/backup.go
+++ b/pkg/task/backup.go
@@ -137,10 +137,7 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
 	}
 	g.Record("BackupTS", backupTS)
 
-	isIncrementalBackup := false
-	if cfg.LastBackupTS > 0 {
-		isIncrementalBackup = true
-	}
+	isIncrementalBackup := cfg.LastBackupTS > 0
 
 	ranges, backupSchemas, err := backup.BuildBackupRangeAndSchema(
 		mgr.GetDomain(), mgr.GetTiKV(), cfg.TableFilter, backupTS)
--- a/pkg/task/backup.go
+++ b/pkg/task/backup.go
@@ -137,6 +137,11 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
 	}
 	g.Record("BackupTS", backupTS)
 
+	isIncrementalBackup := false
+	if cfg.LastBackupTS > 0 {
+		isIncrementalBackup = true
+	}
+
 	ranges, backupSchemas, err := backup.BuildBackupRangeAndSchema(
 		mgr.GetDomain(), mgr.GetTiKV(), cfg.TableFilter, backupTS)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


kennytm:
> record `backupts` in summary log to enhance usability.

i suggest replacing the implementation in `glue.Record()`, and keep `CollectInt()` untouched. this way we could fix #95 together.

https://github.com/pingcap/br/blob/980627aa90e5d6f0349b423127e0221b4fa09ba0/pkg/gluetikv/glue.go#L53-L54

pkg/task/restore.go
```diff
@@ -599,3 +603,37 @@ func restoreTableStream(
...
+		} else {
+			existsDBs = append(existsDBs, db.Info.Name.String())
+		}
+	}
+
+	existsTables := make([]string, 0, len(dbs))
+	nonExistsTables := make([]*utils.Table, 0, len(tables))
+	for _, table := range tables {
+		if ok := infoSchema.TableExists(table.Db.Name, table.Info.Name); !ok {
+			nonExistsTables = append(nonExistsTables, table)
```
YuJuncen, line 628: We filtered out the table if it exists, but `GoValidateFileRanges` leads to that we won't restore files associated to tables if the table absent at `tableStream`... (i.e. The table won't join the pipeline, hence won't be restored...)

3pointer: ok, maybe we should send table to `tableStream` even table already exists

3pointer:
as @YuJuncen said, we cannot filter exists schema in restore side, because we need do some DDL such as alter auto_increment even table already exists. so I changed the filter logic in backup side, in incremental backup, we will filter the table that has nothing to backup.

pkg/task/backup.go
```diff
@@ -137,6 +137,11 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
 	}
 	g.Record("BackupTS", backupTS)
 
+	isIncrementalBackup := false
+	if cfg.LastBackupTS > 0 {
+		isIncrementalBackup = true
+	}
```
YuJuncen, line 143: How about `isIncrementalBackup := cfg.LastBackupTS > 0` ?

Issue #394, opened 2020-07-02
Adjust backup concurrency before BR started.


3pointer:
### Describe your feature request related problem:

Current default config (e.g. concurrency) for BR is not enough. for some highly performance cluster default backup concurrency 4 and restore concurrency 128 is not fit. 

### Describe the feature you'd like:

Maybe we should add some check before BR progress started, to tell user adjust these config manually or do it automatically.

kennytm:
See tikv/tikv#8076 for backup concurrency.

kennytm:
Regarding backup concurrency, I think we should modify the meaning like this:

1. change the TiKV backup thread-pool size (tikv/tikv#8076) from a value specified by the gRPC request to a TiKV config defaulting to 1/2 of number of vCPU.
2. change this to run in parallel, and make backup concurrency control the pool size.

    https://github.com/pingcap/br/blob/aa4b7d51a96d51c2236d92f0d16a1c09eda2062c/pkg/backup/client.go#L394-L404

kennytm:
Note that the cluster causing this issue contains a table with 6 secondary indices and 125 partitions, thus generating 875 ranges for this table alone.

PR #396, opened 2020-07-02
*: run BackupRanges in parallel, change meaning of backup concurrency

kennytm:

Step 2 of #394.

Previously `BackupRanges` execute each `BackupRange` sequentially. When the number of ranges is lot and each range cannot spawn too many backup jobs, this will create a bottleneck.

This PR changes `BackupRanges` to execute each `BackupRange` in parallel. 

As pointed out in #394, the meaning of "concurrency" is changed to mean the number of `BackupRange`s being executed in parallel, rather than the thread pool size in TiKV (this also prevents killing the TiKV servers by supplying an extremely large concurrency).

`BackupRawRanges` is currently not changed since it only has a single range.

**Update 2020/07/04**: It was found that running `BackupRanges` in parallel will cause a race condition since all writes to the same internal field `bc.backupMeta.Files`, and causes some records to be lost. Thus we refactored `backup.Client` to remove the `backupMeta` field entirely to prevent any write-write conflicts. The `backupMeta` object is constructed using the returned arrays from functions with explicit coordination between goroutines.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### Release Note

 - Improve parallel processing power when backing up a large database with many regions.

### DIFF
--- a/pkg/task/backup.go
+++ b/pkg/task/backup.go
@@ -139,6 +139,13 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
 
 	isIncrementalBackup := cfg.LastBackupTS > 0
 
+	req := kvproto.BackupRequest{
+		StartVersion: cfg.LastBackupTS,
+		EndVersion:   backupTS,
+		RateLimit:    cfg.RateLimit,
+		Concurrency:  defaultBackupConcurrency,
+	}
+
 	ranges, backupSchemas, err := backup.BuildBackupRangeAndSchema(
 		mgr.GetDomain(), mgr.GetTiKV(), cfg.TableFilter, backupTS)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/task/backup.go
```diff
@@ -188,10 +188,9 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
 		StartVersion: cfg.LastBackupTS,
 		EndVersion:   backupTS,
 		RateLimit:    cfg.RateLimit,
-		Concurrency:  cfg.Concurrency,
+		Concurrency:  defaultBackupConcurrency,
```
3pointer, line 191: If we only need to backup a few ranges in one backup task, for example backup one table without index,  we can not control the `defaultBackupConcurrency` to a proper value.

kennytm: yes that's why we also need the "Step 1" of #394 to complete the picture.

3pointer: got it

kennytm:
There is potential chance this PR causing data loss (detected as checksum mismatch). Please don't merge yet.

3pointer:
These backupmeta [changes](https://github.com/pingcap/br/blob/d3ec9f4cbd0777735e17d2f9418dcedd6201680a/pkg/backup/client.go#L516-L536) caused data loss,  we should add lock for it.

overvenus:
Please add a check that requires TiKV > v4.0.2 .

kennytm:
@overvenus The `req.concurrency` is still passed downstream. For ≤4.0.2 it will just run using concurrency = 4.

PR #399, opened 2020-07-03
Add a hint when running on incompatible cluster

YuJuncen:
Some times the version info can be lost (e.g. when build br with `go build` directly, instead of `make build`), at that time, the error message is confusing: 

```
None is not in dotted-tri format.
```

In all conditions when we failed to run checksum, we add a hint that "we can use `--check-requirements=false` to skip the version checker."

### Check List 

Tests 

  1. run `go build -o bin/br` directly.
  2. execute `br`.
image

### Release Note

 - No release note

### DIFF
--- a/pkg/conn/conn.go
+++ b/pkg/conn/conn.go
@@ -200,7 +200,10 @@ func NewMgr(
 	if checkRequirements {
 		err = utils.CheckClusterVersion(ctx, pdClient)
 		if err != nil {
-			return nil, err
+			errMsg := "running BR in incompatible version of cluster, " +
+				"error: (%s). " +
+				"if you believe it's OK, use --check-requirements=false to skip."
+			return nil, errors.Errorf(fmt.Sprintf(errMsg, err.Error()))
 		}
 	}
 	log.Info("new mgr", zap.String("pdAddrs", pdAddrs))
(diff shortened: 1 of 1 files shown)


pkg/conn/conn.go
```diff
@@ -200,7 +200,10 @@ func NewMgr(
 	if checkRequirements {
 		err = utils.CheckClusterVersion(ctx, pdClient)
 		if err != nil {
-			return nil, err
+			errMsg := "running BR in incompatible version of cluster, " +
+				"error: (%s). " +
+				"if you believe it's OK, use --check-requirements=false to skip."
+			return nil, errors.Errorf(fmt.Sprintf(errMsg, err.Error()))
```
overvenus, line 206:
```suggestion
			return nil, errors.Errorf(errMsg, err.Error())
```

kennytm: even better, use `errors.Annotate` 🤔

pkg/conn/conn.go
```diff
@@ -200,7 +200,10 @@ func NewMgr(
 	if checkRequirements {
 		err = utils.CheckClusterVersion(ctx, pdClient)
 		if err != nil {
-			return nil, err
+			errMsg := "running BR in incompatible version of cluster, " +
+				"error: (%s). " +
+				"if you believe it's OK, use --check-requirements=false to skip."
+			return nil, errors.Errorf(errMsg, err.Error())
```
overvenus, line 206: What about `errors.Annotate`

PR #402, opened 2020-07-06
Set safe point at background

YuJuncen:
Currently, we just set the service GC safe point when writing SST files, but not when checksumming.

If checksum time expires 5min (the service safe point lifetime) plus the TiDB GC lifetime, we might meet "GC life time is shorter than transaction duration" error.

I moved the logic of setting GC safe point into another goroutine, which will set the service safe point every (1/5 service safe point lifetime).

This goroutine will be canceled by context.

### Check List 

Tests 

    Though this test can be automatized by fail points, if we did so, our ingratiation test time will grow to 15 mins+, which is hard to accept. 
    Firstly, we insert fail point at here:    
image
    Then, we run current master branch with fail point:
image
    After fixed, we can finish the backup:
image

### Release Note

 - Fix a bug that may cause failed to backup when dataset is too large and GC lifetime is too brief.

### DIFF
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -416,23 +416,11 @@ func (bc *Client) BackupRanges(
 	log.Info("current backup safePoint job",
 		zap.Uint64("backupTS", backupTS))
 
-	finished := false
 	for {
-		err := CheckGCSafePoint(ctx, bc.mgr.GetPDClient(), backupTS)
-		if err != nil {
-			log.Error("check GC safePoint failed", zap.Error(err))
-			return err
-		}
-		if finished {
-			// Return error (if there is any) before finishing backup.
-			return err
-		}
 		select {
 		case err, ok := <-errCh:
 			if !ok {
-				// Before finish backup, we have to make sure
-				// the backup ts does not fall behind with GC safepoint.
-				finished = true
+				return nil
 			}
 			if err != nil {
 				return err
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -439,15 +439,6 @@ func (bc *Client) BackupRanges(
 	t := time.NewTicker(time.Second * 5)
 	defer t.Stop()
 
-	backupTS := req.EndVersion
-	// use lastBackupTS as safePoint if exists
-	if req.StartVersion > 0 {
-		backupTS = req.StartVersion
-	}
-
-	log.Info("current backup safePoint job",
-		zap.Uint64("backupTS", backupTS))
-
 	for err := range errCh {
 		if err != nil {
 			return nil, err
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -447,8 +447,14 @@ func (bc *Client) BackupRanges(
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/backup/safe_point.go
```diff
@@ -52,3 +54,32 @@ func UpdateServiceSafePoint(ctx context.Context, pdClient pd.Client, ttl int64,
...
+// StartServiceSafePointKeeper will run UpdateServiceSafePoint periodicity
+// hence keeping service safepoint won't lose.
+func StartServiceSafePointKeeper(
+	ctx context.Context,
+	ttl int64,
+	pdClient pd.Client,
+	backupTS uint64,
+) {
+	// At least 1 second gap, or time.NewTicker will blame us.
+	gapSec := time.Duration(utils.MaxInt(int(ttl/5), 1))
```
kennytm, line 67: what's special about "5"?

btw please make this a true duration:

```go
gap := time.Duration(ttl) * time.Second / 5
if gap < time.Second {
    gap = time.Second
}
```

YuJuncen: The default ttl is 5 mins... So I just choice one of its factor.

Do we need so frequently reset it? The origin version reset it every time a range is fully backed up.

YuJuncen: (And every 5 seconds...)

kennytm: As long as it is reset before the TTL it is fine

kennytm:
what if we just remove the loop inside BackupRange and move it entirely into StartServiceSafePointKeeper? @3pointer

YuJuncen:
> what if we just remove the loop inside BackupRange and move it entirely into StartServiceSafePointKeeper? @3pointer

I think that would be a good idea 👍

pkg/backup/client.go
```diff
@@ -439,44 +441,14 @@ func (bc *Client) BackupRanges(
...
-			}
-			if err != nil {
-				return nil, err
-			}
-		case <-t.C:
-		}
 	}
+
+	<-allFilesCollected
+	return allFiles, nil
```
kennytm, line 451:
```suggestion
	select {
	case <-allFilesCollected:
		return allFiles, nil
	case <-ctx.Done():
		return nil, errors.Trace(ctx.Err())
	}
```

maybe combine with the `errCh` too, like if `errCh` returns `nil` treat it as success.

YuJuncen: `errCh` is closed at this context, maybe there is no need to select on it.

pkg/backup/client.go
```diff
@@ -447,8 +447,14 @@ func (bc *Client) BackupRanges(
...
 
-	<-allFilesCollected
-	return allFiles, nil
+	select {
+	case <-allFilesCollected:
+		return allFiles, nil
+	case <-ctx.Done():
+		return nil, errors.Trace(ctx.Err())
+	case err := <-errCh:
+		return nil, errors.Trace(err)
```
kennytm: (you should remove the `for err := range errCh` loop above if handling this here)

PR #404, opened 2020-07-07
backup: support explicitly set sst compression type

glorv:
Add command line parameter for sst file compression type

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes
- https://github.com/pingcap/kvproto/pull/642
- https://github.com/tikv/tikv/pull/8200

### Release Note

 -


pkg/task/backup.go
```diff
@@ -60,6 +62,7 @@ func DefineBackupFlags(flags *pflag.FlagSet) {
 	flags.String(flagBackupTS, "", "the backup ts support TSO or datetime,"+
 		" e.g. '400036290571534337', '2018-05-11 01:42:23'")
 	flags.Int64(flagGCTTL, backup.DefaultBRGCSafePointTTL, "the TTL (in seconds) that PD holds for BR's GC safepoint")
+	flags.String(flagCompressionType, "", "backup sst file compression algorithm, value can be one of 'lz4|zstd|snappy'")
```
kennytm, line 65: provide a default value?

```suggestion
	flags.String(flagCompressionType, "zstd", "backup sst file compression algorithm, value can be one of 'lz4|zstd|snappy'")
```

pkg/task/backup.go
```diff
@@ -295,3 +309,20 @@ func parseTSString(ts string) (uint64, error) {
...
+	var ct kvproto.CompressionType
+	switch s {
+	case "lz4":
+		ct = kvproto.CompressionType_LZ4
+	case "snappy":
+		ct = kvproto.CompressionType_SNAPPY
+	case "zstd":
+		ct = kvproto.CompressionType_ZSTD
+	case "":
+		ct = kvproto.CompressionType_UNKNOWN
```
kennytm, line 323:
```suggestion
```

pkg/task/backup.go
glorv: the default value should be default, which means the compression type will be determined by tikv.  In currently logic, tikv will choose the fastest algorithm and in most case it will be 'lz4'.

kennytm: the default is called "unknown" though

glorv: e, so we should change the name in kv-proto to default?

pkg/task/backup.go
```diff
@@ -28,9 +28,10 @@ import (
 )
 
 const (
-	flagBackupTimeago = "timeago"
-	flagBackupTS      = "backupts"
-	flagLastBackupTS  = "lastbackupts"
+	flagBackupTimeago   = "timeago"
+	flagBackupTS        = "backupts"
+	flagLastBackupTS    = "lastbackupts"
+	flagCompressionType = "compression-type"
```
overvenus, line 34:
```suggestion
	flagCompressionType = "compression"
```

pkg/task/backup.go
overvenus: We should provide a validate default value, "zstd" looks good to me.

glorv: Ok. Then I think we can set default to `lz4` because it's the current compression type tikv uses. So if user don't set this flag, the behavior is same with previous version, so it's compatible will old version

3pointer:
maybe we should add a integration test for "zstd"

PR #405, opened 2020-07-08
Fix the problem that backup time may be negative

YuJuncen:
#396 made a little mistake, which make `last backup finish time - now` become backup time spent, and will lead to backup time become negative.

### Check List 

Tests 

image

### Release Note

 - No release note

### DIFF
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -409,7 +409,7 @@ func (bc *Client) BackupRanges(
 		for files := range filesCh {
 			cur, start = start, time.Now()
 			allFiles = append(allFiles, files...)
-			summary.CollectSuccessUnit("backup ranges", 1, cur.Sub(start))
+			summary.CollectSuccessUnit("backup ranges", 1, start.Sub(cur))
 		}
 		log.Info("Backup Ranges", zap.Duration("take", cur.Sub(init)))
 	}()
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -405,13 +405,13 @@ func (bc *Client) BackupRanges(
 	allFiles := make([]*kvproto.File, 0, len(ranges))
 	go func() {
 		init := time.Now()
-		start, cur := init, init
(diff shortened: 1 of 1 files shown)


pkg/backup/client.go
```diff
@@ -405,13 +405,14 @@ func (bc *Client) BackupRanges(
...
+		lastBackupStart, currentBackupStart := init, init
 		for files := range filesCh {
-			cur, start = start, time.Now()
+			lastBackupStart, currentBackupStart = currentBackupStart, time.Now()
 			allFiles = append(allFiles, files...)
-			summary.CollectSuccessUnit("backup ranges", 1, cur.Sub(start))
+			summary.CollectSuccessUnit("backup ranges", 1, currentBackupStart.Sub(lastBackupStart))
 		}
-		log.Info("Backup Ranges", zap.Duration("take", cur.Sub(init)))
+		_ = lastBackupStart
```
3pointer, line 414: 🤔️

YuJuncen:
@kennytm if we remove that line, golang-ci will get unhappy:

```
pkg/backup/client.go:408:23: ineffectual assignment to `lastBackupStart` (ineffassign)
 		currentBackupStart, lastBackupStart := init, init
```

It's really strange cuz if we leave `currentBackupStart` unused everything is alright but `lastBackupStart` cannot.
Maybe it wants me to move it into the inner scope, but I think the current version is more readable, and add this line to make CI happy.

pkg/backup/client.go
YuJuncen: OK. Let me replace it with `// nolint`...

kennytm:
@YuJuncen i think it's because `lastBackupStart` is immediately rewritten inside the for loop and unused outside the for loop. Maybe you could make it `var lastBackupStart time.Time` to silence the warning.

YuJuncen:
@kennytm I think so, but I used `// nolint:ineffassign` to silence it.

PR #406, opened 2020-07-08
glue: set query DDL when executing DDLs

### LINKED ISSUE #364: DDL through CreateDatabase or CreateTable gets wrong query

1. What did you do?

We need to set the query for the DDL, [like this](https://github.com/pingcap/tidb/blob/7daa4a7f88194a9d4317a27acdf2143312375cd5/session/session.go#L1187)

```go
se.SetValue(sessionctx.QueryString, s.OriginText())
```

But we didn't set it, so query in DDL history is wrong.

YuJuncen:
fix #364 

Before executing DDLs, we call `session.SetVar` firstly, inject the DDL by `SHOW CREATE TABLE`.

### Check List 

Tests 

 - Integration test

### Release Note

 - No release note

### DIFF
--- a/pkg/gluetidb/glue.go
+++ b/pkg/gluetidb/glue.go
@@ -6,9 +6,11 @@ import (
 	"bytes"
 	"context"
 
+	"github.com/pingcap/log"
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/mysql"
 	pd "github.com/pingcap/pd/v4/client"
+	"github.com/pingcap/tidb/config"
 	"github.com/pingcap/tidb/ddl"
 	"github.com/pingcap/tidb/domain"
 	"github.com/pingcap/tidb/executor"
@@ -26,6 +28,15 @@ const (
 	defaultCapOfCreateDatabase = 64
 	brComment                  = "/*from('br')*/"
 )
+
+// New makes a new tidb glue.
+func New() Glue {
+	log.Debug("enabling no register config")
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.SkipRegisterToDashboard = true
+	})
+	return Glue{}
+}
 
 // Glue is an implementation of glue.Glue using a new TiDB session.
 type Glue struct {
--- a/pkg/gluetidb/glue.go
+++ b/pkg/gluetidb/glue.go
@@ -26,7 +26,7 @@ import (
 const (
 	defaultCapOfCreateTable    = 512
 	defaultCapOfCreateDatabase = 64
-	brComment                  = `/*from("br")*/`
+	brComment                  = `/*from(br)*/`
 )
 
 // New makes a new tidb glue.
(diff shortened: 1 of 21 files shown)


pkg/gluetidb/glue.go
```diff
@@ -79,19 +105,63 @@ func (gs *tidbSession) CreateDatabase(ctx context.Context, schema *model.DBInfo)
...
+	table := tbl.Clone()
+	table.AutoIncID = 0
+	result := bytes.NewBuffer(make([]byte, 0, defaultCapOfCreateTable))
+	alloc, err := gs.allocatorFor(tbl, dbName)
+	if err != nil {
+		return "", err
+	}
+	// this can never fail.
+	_, _ = result.WriteString(brComment)
+	if err := executor.ConstructResultOfShowCreateTable(gs.se, tbl, alloc, result); err != nil {
```
kennytm, line 152: let's just pass `nil` here. the allocators are only used for reading out the auto-inc and auto-random IDs which we will explicitly reset anyway.

```suggestion
	if err := executor.ConstructResultOfShowCreateTable(gs.se, tbl, nil, result); err != nil {
```

YuJuncen: I do want to do that, but when this is `nil`, we will meet a panic if the table has auto increment columns...

YuJuncen: (at executor/show.go:907)

kennytm: @YuJencen what about `autoid.Allocators{}`

YuJuncen: `autoid.Allocators` doesn't implement `Allocator`. 😢

kennytm: Oops it was just [changed to `Allocators`](https://github.com/pingcap/tidb/blob/a99fdc098cb3bccc63ca74dad3d682b6f80a0fbd/executor/show.go#L722) in the latest master since pingcap/tidb#18217 👀

YuJuncen: OK, seems after we update tidb to newest master, we can give it a `autoid.Allcators{}`. 😄

YuJuncen:
Let's block it until the update TiDB version PR merged before the version frozen.

tests/br_db/run.sh
```diff
@@ -51,4 +51,9 @@ if [ "$table_count" -ne "2" ];then
     exit 1
 fi
 
+# Test BR DDL query string
+echo "testing DDL query..."
+curl 127.0.0.1:10080/ddl/history | grep -E '/\*from(.*)\*/CREATE TABLE'
+curl 127.0.0.1:10080/ddl/history | grep -E '/\*from(.*)\*/CREATE DATABASE'
```
kennytm, line 57:
```suggestion
curl 127.0.0.1:10080/ddl/history | grep -E '/\*from('"'br'"')\*/CREATE TABLE'
curl 127.0.0.1:10080/ddl/history | grep -E '/\*from('"'br'"')\*/CREATE DATABASE'
```

YuJuncen:
`java.net.ProtocolException: Expected HTTP 101 response but was '500 Internal Server Error'`

/run-integration-test

PR #411, opened 2020-07-09
Refactor: replace wait groups with errgroup, and drop db when test finished

YuJuncen:
After #396 merged, we can run tasks in `errgroup.Group` with `worker`.

So it's time to refactor some dirty patterns like:
```go
errCh := make(chan error, 1)
wg := new(sync.WaitGroup)
worker := utils.NewWorker(x, "xxx")
for range tasks {
    wg.Add(1)
    worker.Apply(func() { 
      defer wg.Done()
      if err := task(); err != nil {
        errCh <- err
      }
    }
}
wg.Wait()
// ... more logic about handling asynchronous errors...
```

With errgroup, we just write:

```go
eg, ectx := errgroup.WithContext(ctx)
worker := utils.NewWorker(x, "xxx")
for range tasks {
    worker.ApplyInErrGroup(eg, func() error { return task(ctx) })
}
err := eg.Wait()
```

What is more, we removed the `ctx` and `cancel` field of `Importer`, which is deprecated by Go [here](https://kapeli.com/dash_share?docset_file=Go&docset_name=Go&path=golang.org/pkg/context/index.html%23Context&platform=go&repo=Main&source=golang.org/pkg/context/&version=1.14.4):

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx.

And we add the `DROP DATABASE $DB` sql at end of some tests, so if we run `make integration_test` directly they won't influence each other.

### Check List 

Tests 

 - Unit test
 - Integration test
All old tests.

Code changes

We removed `ctx` and `cancel` field of `Importer`, so its methods that uses those fields must receive them as argument.

### Release Note

 - No release note.

### DIFF
--- a/pkg/restore/import.go
+++ b/pkg/restore/import.go
@@ -142,26 +142,19 @@ type FileImporter struct {
 	isRawKvMode bool
 	rawStartKey []byte
 	rawEndKey   []byte
-
-	ctx    context.Context
-	cancel context.CancelFunc
 }
 
 // NewFileImporter returns a new file importClient.
 func NewFileImporter(
-	ctx context.Context,
 	metaClient SplitClient,
 	importClient ImporterClient,
 	backend *backup.StorageBackend,
 	isRawKvMode bool,
 	rateLimit uint64,
 ) FileImporter {
-	ctx, cancel := context.WithCancel(ctx)
 	return FileImporter{
 		metaClient:   metaClient,
 		backend:      backend,
-		ctx:          ctx,
-		cancel:       cancel,
 		importClient: importClient,
 		isRawKvMode:  isRawKvMode,
 		rateLimit:    rateLimit,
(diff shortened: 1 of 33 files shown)


pkg/restore/import.go
```diff
@@ -242,12 +236,12 @@ func (importer *FileImporter) Import(
 			info := regionInfo
 			// Try to download file.
 			var downloadMeta *import_sstpb.SSTMeta
-			errDownload := utils.WithRetry(importer.ctx, func() error {
+			errDownload := utils.WithRetry(cctx, func() error {
```
3pointer, line 239: should this context be `importer.ctx` here? these contexts were different

YuJuncen: `importer.ctx` is removed, see:
> What is more, we removed the ctx and cancel field of Importer, which is deprecated by Go here:
> 
> > Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx.

3pointer: But the original context here won't timeout.  why choose `cctx` not `ctx` here?

YuJuncen: Oops...

PR #420, opened 2020-07-14
restore: don't restore auto id if the table doesn't has it.

YuJuncen:
Fix CI failure after [tidb#18326](https://github.com/pingcap/tidb/pull/18326)

We don't try to restore auto ID for tables that don't need it.

### Check List 

Tests 

 - Integration test

### Release Note

 - fix a bug that causes, we will restore auto id even the table doesn't has it.

### DIFF
--- a/pkg/restore/db.go
+++ b/pkg/restore/db.go
@@ -158,11 +158,13 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 		default:
 			alterAutoIncIDFormat = "alter table %s.%s auto_increment = %d;"
 		}
-		restoreMetaSQL = fmt.Sprintf(
-			alterAutoIncIDFormat,
-			utils.EncloseName(table.Db.Name.O),
-			utils.EncloseName(table.Info.Name.O),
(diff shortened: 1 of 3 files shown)


kennytm:
The `br_move_backup` test is stuck

https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/2308/pipeline/96

pkg/restore/db.go
```diff
@@ -163,9 +164,11 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 			utils.EncloseName(table.Db.Name.O),
 			utils.EncloseName(table.Info.Name.O),
 			table.Info.AutoIncID)
+		if table.NeedRebaseAutoID() {
```
kennytm, line 167: 🤔  why not just use `NeedAutoID(table.Info)` here?

YuJuncen: Because this function isn't written in the first version 🧐.

You'r right...

PR #426, opened 2020-07-16
Remove scheduler

### LINKED ISSUE #412: Remove pd scheduler in backup
### Is your feature request related to a problem? Please describe:

Add an option that BR can remove pd leader related scheduler automatically. just like restore [does](https://github.com/pingcap/br/blob/ef923324feec7cc8d58f7818cc7b6f8758cc4a0d/pkg/task/restore.go#L359-L423)
In our test: 
with balance leader scheduler cost:  9m33s
without balance leader scheduler cost: 2m31s

this could happen because there are so many incomplete retry when leader transfer.

Add an option that can remove scheduler automatically to speed up backup.

### Describe alternatives you've considered:

This change has some impact on backup cluster.so we can set default value to false.

YuJuncen:
fixed #412.

Firstly, we extract the logic of removing schedulers to `conn` package, as a method of `conn.Manager`.
Then, we add a flag that controls whether to remove scheduler when backing up.
Finally, we call the extracted method on backup if the flag is set. 

### Check List 

Tests 

 - Integration test

### Release Note

 - Add an option that can be used to speed up backing up by changing some cluster config.

### DIFF
--- a/pkg/conn/scheduler_utils.go
+++ b/pkg/conn/scheduler_utils.go
@@ -83,13 +83,21 @@ func restoreSchedulers(ctx context.Context, mgr *Mgr, clusterCfg clusterConfig) 
 	return nil
 }
 
+func (mgr *Mgr) makeUndoFunctionByConfig(config clusterConfig) utils.UndoFunc {
+	restore := func(ctx context.Context) error {
+		return restoreSchedulers(ctx, mgr, config)
+	}
+	return restore
+}
+
 // RemoveSchedulers removes the schedulers that may slow down BR speed.
-// TODO make each step returns a function that can restore schedulers it has removed.
-func (mgr *Mgr) RemoveSchedulers(ctx context.Context) (utils.UndoFunc, error) {
+func (mgr *Mgr) RemoveSchedulers(ctx context.Context) (undo utils.UndoFunc, err error) {
+	undo = utils.Nop
+
 	// Remove default PD scheduler that may affect restore process.
 	existSchedulers, err := mgr.ListSchedulers(ctx)
 	if err != nil {
-		return utils.Nop, nil
+		return
 	}
 	needRemoveSchedulers := make([]string, 0, len(existSchedulers))
 	for _, s := range existSchedulers {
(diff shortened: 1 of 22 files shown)


pkg/conn/scheduler_utils.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+	scheduler, err := removePDLeaderScheduler(ctx, mgr, needRemoveSchedulers)
+	if err != nil {
+		return utils.Nop, nil
+	}
+
+	stores, err := mgr.GetPDClient().GetAllStores(ctx)
+	if err != nil {
+		return utils.Nop, err
```
kennytm, line 107: i think you shouldn't return Nop given the removePDLeaderScheduler has succeeded.

pkg/task/backup.go
```diff
@@ -64,6 +66,9 @@ func DefineBackupFlags(flags *pflag.FlagSet) {
 	flags.Int64(flagGCTTL, backup.DefaultBRGCSafePointTTL, "the TTL (in seconds) that PD holds for BR's GC safepoint")
 	flags.String(flagCompressionType, "zstd",
 		"backup sst file compression algorithm, value can be one of 'lz4|zstd|snappy'")
+
+	flags.Bool(flagRemoveSchedulers, false,
+		"remove some of PD schedulers to speed up backup, but will make influence to cluster")
```
kennytm, line 71:
```suggestion
		"disable the balance, shuffle and region-merge schedulers in PD to speed up backup")
```

pkg/task/backup_raw.go
```diff
@@ -45,6 +48,8 @@ func DefineRawBackupFlags(command *cobra.Command) {
 	command.Flags().StringP(flagEndKey, "", "", "backup raw kv end key, key is exclusive")
 	command.Flags().String(flagCompressionType, "zstd",
 		"backup sst file compression algorithm, value can be one of 'lz4|zstd|snappy'")
+	command.Flags().Bool(flagRemoveSchedulers, false,
+		"remove some of PD schedulers to speed up backup, but will make influence to cluster")
```
kennytm, line 52: (ditto)

tests/br_other/run.sh
```diff
@@ -72,9 +73,11 @@ sleep 1
 curl "http://localhost:$PPROF_PORT/debug/pprof/trace?seconds=1" 2>&1 > /dev/null
 echo "pprof started..."
 
+curl http://$PD_ADDR/pd/api/v1/config/schedule | grep '"disable": true'
+
 backup_fail=0
 echo "another backup start expect to fail due to last backup add a lockfile"
-run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB/lock" --concurrency 4 || backup_fail=1
+run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB/lock"  --concurrency 4 || backup_fail=1
```
3pointer, line 80: remove space

pkg/task/backup.go
```diff
@@ -64,6 +66,9 @@ func DefineBackupFlags(flags *pflag.FlagSet) {
 	flags.Int64(flagGCTTL, backup.DefaultBRGCSafePointTTL, "the TTL (in seconds) that PD holds for BR's GC safepoint")
 	flags.String(flagCompressionType, "zstd",
 		"backup sst file compression algorithm, value can be one of 'lz4|zstd|snappy'")
+
+	flags.Bool(flagRemoveSchedulers, false,
+		"disable the balance, shuffle and region-merge schedulers in PD to speed up backup")
```
3pointer, line 71: maybe we should hide it or let user know this will impact on online cluster.

PR #436, opened 2020-07-22
restore: split region by range(file) size or kv count

3pointer:
Currently we do split region by each range endkey, this may not work well, 
because there are three boundary factors determine the backup file(*sst)
1.  the region size
2. table(partition) range
3. index range

so we could have lots of small backup files at sometimes. and finally it will do lots of split operation in restore cluster. 

according to current TiKV split policy. we do split region by range endkey in any of conditions

1. accumulated ranges logic size >= 96MB.
2. accumulated ranges KV count >= 960000.
3. the last range between record and index.
4. the last range between different index.

### Check List 

Tests 

Side effects

 - Increased code complexity

Related changes

### Release Note

 - No release note


pkg/rtree/rtree.go
```diff
@@ -16,14 +16,16 @@ import (
...
+	EndKey     []byte
+	Files      []*backup.File
+	TotalBytes uint64
+	TotalKVs   uint64
 }
 
 // String formats a range to a string.
 func (rg *Range) String() string {
-	return fmt.Sprintf("[%x %x]", rg.StartKey, rg.EndKey)
+	return fmt.Sprintf("[%x %x %d %d]", rg.StartKey, rg.EndKey, rg.TotalKVs, rg.TotalBytes)
```
YuJuncen, line 28: Log like `[<start> <end> 96000 960000000]` maybe confusing, how about add some hint?
```suggestion
	return fmt.Sprintf("[%x %x %dkvs %dbytes]", rg.StartKey, rg.EndKey, rg.TotalKVs, rg.TotalBytes)
```

pkg/restore/split.go
```diff
@@ -38,6 +38,9 @@ const (
 	RejectStoreCheckRetryTimes  = 64
 	RejectStoreCheckInterval    = 100 * time.Millisecond
 	RejectStoreMaxCheckInterval = 2 * time.Second
+
+	DefaultRegionSize = 96 << 20 // 96M
```
glorv, line 42:
```suggestion
	DefaultRegionSize = 96 * units.MB
```

pkg/restore/split.go
```diff
@@ -263,14 +268,46 @@ func (rs *RegionSplitter) splitAndScatterRegions(
...
+	curKVs := uint64(0)
+	lastEndKey := []byte{}
+	lastIndexID := int64(0)
+	isLastRecord := false
+	// No need to split region by every range's endkey
+	// we can split region by range endkey in any of conditions
+	// 1. accumulated ranges logic size >= 96MB.
+	// 2. accumulated ranges KV count >= 960000.
+	// 3. the last range between record and index.
+	// 4. the last range between different index.
```
glorv, line 289: Do we still need conditions 3,4? It's not sufficient for small tables

3pointer:
close due to https://github.com/pingcap/br/pull/578

PR #446, opened 2020-07-30
storage: extend the `ExternalStorage` interface and implements

glorv:
extend the `ExternalStorage` interface and implement the new methods for local and s3 storage. This extension is currently used  in lightning to support restore from s3 storage.

Add new methods for ExternalStorage:
```
type ReadSeekCloser interface {
	io.Reader
	io.Seeker
	io.Closer
}

type ExtrenalStorage interface {
	// Open a Reader by file name
	Open(ctx context.Context, name string) (ReadSeekCloser, error)
	// WalkDir traverse all the files in a dir
	WalkDir(ctx context.Context, fn func(string, int64) error) error
}
```

Since s3 GetObject does not provide seek directly, we implement this by: 1) read and skip the bytes if the distance is small, else we open a new reader set the start position to the seek position. 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - No release note


pkg/storage/s3.go
```diff
@@ -292,3 +299,143 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error)
...
+}
+
+// Close implement the io.Closer interface
+func (r *s3ObjectReader) Close() error {
+	return r.reader.Close()
+}
+
+// Seek implement the io.Seeker interface
+func (r *s3ObjectReader) Seek(offset int64, whence int) (int64, error) {
+	// if seek ahead no more than 64k, we read add drop these data
```
kennytm, line 398: move this comment to line 413.

pkg/storage/s3.go
```diff
@@ -292,3 +299,143 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error)
...
+	var realOffset int64
+	if whence == io.SeekStart {
+		realOffset = offset
+	} else if whence == io.SeekCurrent {
+		realOffset = r.pos + offset
+	} else {
+		// TODO
+		return 0, errors.New("seek by SeekEnd is not supported yet")
+	}
```
kennytm, line 407:
```suggestion
	switch whence {
	case io.SeekStart:
		realOffset = offset
	case io.SeekCurrent:
		realOffset = r.pos + offset
	default:
		// TODO
		return 0, errors.New("seek by SeekEnd is not supported yet")
	}
```

pkg/storage/s3.go
```diff
@@ -292,3 +299,143 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error)
...
+		buf := make([]byte, batch, batch)
+		total := offset - r.pos
+		remain := total
+		for remain > 0 {
+			n, err := r.Read(buf[:utils.MinInt64(remain, batch)])
+			if err != nil {
+				return total - remain + int64(n), err
+			}
+			remain -= int64(n)
+		}
```
kennytm, line 424:
```suggestion
		written, err := io.CopyN(ioutil.Discard, r, offset-r.pos)
		if err != nil {
			return r.pos + written, err
		}
```

pkg/storage/gcs.go
```diff
@@ -123,6 +123,18 @@ func (s *gcsStorage) FileExists(ctx context.Context, name string) (bool, error)
...
 }
 
+// Open a Reader by file name
+func (s *gcsStorage) Open(ctx context.Context, name string) (ReadSeekCloser, error) {
+	// TODO, implement this if needed
+	panic("Unsupported Operation")
+}
+
+// WalkDir traverse all the files in a dir
+func (s *gcsStorage) WalkDir(ctx context.Context, fn func(string, int64) error) error {
```
kennytm, line 133: please document that the first argument of `fn` is the full path to be passed into `Open()` and the second argument is the file size.

pkg/storage/flags.go
```diff
@@ -8,13 +8,13 @@ import (
 
 // DefineFlags adds flags to the flag set corresponding to all backend options.
 func DefineFlags(flags *pflag.FlagSet) {
-	defineS3Flags(flags)
+	DefineS3Flags(flags)
 	defineGCSFlags(flags)
 }
```
kennytm, line 13: what is the rationale of only exposing S3 but not GCS?

glorv:
@kennytm  Shall we change the `s3` and `gcs` flag definitions from `pflag` to `flag` so that lightning can share the same definitions without migrating into `pflag`?

kennytm:
@glorv we don't really need those flags since they can be passed via URL parameters (`s3://bucket/prefix?region=us-east-1` etc).

pkg/storage/s3.go
```diff
@@ -292,3 +300,143 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error)
...
+	if err != nil {
+		return nil, err
+	}
+
+	// FIXME: we test in minio, when request with Range, the result.AcceptRanges is a bare string 'range',
+	//  not sure whether this is a feature or bug
+	//if rangeOffset != nil && (result.AcceptRanges == nil || *result.AcceptRanges != *rangeOffset) {
+	//	return nil, errors.Errorf("open file '%s' failed, expected range: %s, got: %v",
+	//		name, *rangeOffset, result.AcceptRanges)
+	//}
```
overvenus, line 378: Have we tested in AWS S3?

pkg/storage/local.go
```diff
@@ -42,7 +72,10 @@ func pathExists(_path string) (bool, error) {
 	return true, nil
 }
 
-func newLocalStorage(base string) (*localStorage, error) {
+// NewLocalStorage return a LocalStorage at directory `base`.
+//
+// export for test.
+func NewLocalStorage(base string) (*LocalStorage, error) {
```
overvenus, line 78: Where is the test?

pkg/storage/s3.go
```diff
@@ -292,3 +300,143 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error)
...
+	default:
+		// TODO: maybe we can fetch the object stat and calculate the absolute offset
+		return 0, errors.New("seek by SeekEnd is not supported yet")
+	}
+
+	if realOffset == r.pos {
+		return realOffset, nil
+	}
+
+	// if seek ahead no more than 64k, we read add drop these data
```
overvenus, line 420:
```suggestion
	// if seek ahead no more than 64k, we discard these data
```

pkg/storage/s3.go
```diff
@@ -292,3 +300,143 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error)
...
+		return realOffset, nil
+	}
+
+	// close current read and open a new one which target offset
+	err := r.reader.Close()
+	if err != nil {
+		return 0, err
+	}
+
+	newReader, err := r.storage.open(context.Background(), r.name, realOffset, 0)
```
overvenus, line 435: Please avoid use Background context.

pkg/storage/local.go
glorv: used in lightning tests

overvenus:
LGTM

PR #448, opened 2020-08-05
brieSQL: fix multi-thread panic during SQL restore

3pointer:
In SQL restore. we choose one session to execute SQL, 
so we can't simple use dbPool to create table concurrently.

In restore. if we are using SQL `glue.OwnerStorage()` will return false. 
so we use this condition to tell whether we should make dbPool or not.

### Check List 

Tests 

test will include in https://github.com/pingcap/tidb/pull/18993

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release Note

 - No release note


overvenus:
LGTM

3pointer:
[2020-08-05T06:36:37.592Z] Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = System exception: cannot start thread: Resource temporarily unavailable, e.what() = System exception

PR #452, opened 2020-08-10
*: remove in-struct context

overvenus:

According to golang context doc, context should store inside a struct type; instead, pass a Context explicitly to each function that needs it.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### Release Note

 - No release note.


pkg/backup/client.go
```diff
@@ -400,8 +400,6 @@ func (bc *Client) BackupRanges(
 	updateCh glue.Progress,
 ) ([]*kvproto.File, error) {
 	errCh := make(chan error)
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
```
kennytm, line 404: Removing these WithCancel changes the semantic. Are you sure this refactoring is correct?

YuJuncen: Seems it's OK because this `cancel` is used for just canceling pending tasks while any error occurs. But after #402, the same task could be done by `errorgroup`.

YuJuncen:
/run-integration-test 

```
tikvRPC backoffer.maxSleep 80000ms is exceeded, errors:\nsend tikv request error: rpc error: code = Internal desc = grpc: failed to unmarshal the received message proto: wrong wireType = 2 for field UseAsyncCommit
```

mismatched `kvproto` version? Could retry it resolve it?

YuJuncen:
LGTM

pkg/backup/client.go
overvenus: Yes, `WithCancel` is unnecessary, all contexts derive from the `defaultContext` in package cmd.

3pointer (approved):
LGTM

ti-srebot (approved):

lance6716:
LGTM

PR #456, opened 2020-08-12
add s3 api to adapt cdc log

3pointer:
part of https://github.com/pingcap/ticdc/pull/826, to unify all s3 related api in the same repo. we decide to put new s3 api which ticdc needed into br.

move ticdc s3 api in br

### Check List 

Tests 

 
Related changes

### Release Note

 - No release note

### DIFF
--- a/pkg/storage/gcs.go
+++ b/pkg/storage/gcs.go
@@ -135,9 +135,14 @@ func (s *gcsStorage) Open(ctx context.Context, name string) (ReadSeekCloser, err
 // The first argument is the file path that can be used in `Open`
 // function; the second argument is the size in byte of the file determined
 // by path.
-func (s *gcsStorage) WalkDir(ctx context.Context, fn func(string, int64) error) error {
+func (s *gcsStorage) WalkDir(ctx context.Context, dir string, listCount int64, fn func(string, int64) error) error {
 	// TODO, implement this if needed
 	panic("Unsupported Operation")
+}
+
+// CreateUploader implenments ExternalStorage interface.
+func (s *gcsStorage) CreateUploader(ctx context.Context, name string) (Uploader, error) {
+	return nil, nil
 }
 
 func newGCSStorage(ctx context.Context, gcs *backup.GCS, sendCredential bool) (*gcsStorage, error) {
--- a/pkg/storage/gcs.go
+++ b/pkg/storage/gcs.go
@@ -142,7 +142,8 @@ func (s *gcsStorage) WalkDir(ctx context.Context, dir string, listCount int64, f
 
 // CreateUploader implenments ExternalStorage interface.
 func (s *gcsStorage) CreateUploader(ctx context.Context, name string) (Uploader, error) {
-	return nil, nil
+	// TODO, implement this if needed
+	panic("gcs storage not support multi-upload")
 }
 
 func newGCSStorage(ctx context.Context, gcs *backup.GCS, sendCredential bool) (*gcsStorage, error) {
--- a/pkg/storage/local.go
+++ b/pkg/storage/local.go
@@ -42,7 +42,7 @@ func (l *LocalStorage) FileExists(ctx context.Context, name string) (bool, error
 // The first argument is the file path that can be used in `Open`
 // function; the second argument is the size in byte of the file determined
 // by path.
-func (l *LocalStorage) WalkDir(ctx context.Context, fn func(string, int64) error) error {
+func (l *LocalStorage) WalkDir(ctx context.Context, dir string, listCount int64, fn func(string, int64) error) error {
 	return filepath.Walk(l.base, func(path string, f os.FileInfo, err error) error {
 		if err != nil {
 			return errors.Trace(err)
@@ -54,6 +54,11 @@ func (l *LocalStorage) WalkDir(ctx context.Context, fn func(string, int64) error
 
 		return fn(f.Name(), f.Size())
 	})
+}
+
+// CreateUploader implenments ExternalStorage interface.
+func (l *LocalStorage) CreateUploader(ctx context.Context, name string) (Uploader, error) {
+	return nil, nil
 }
 
 // Open a Reader by file name.
--- a/pkg/storage/local.go
+++ b/pkg/storage/local.go
@@ -58,7 +58,7 @@ func (l *LocalStorage) WalkDir(ctx context.Context, dir string, listCount int64,
 
 // CreateUploader implenments ExternalStorage interface.
 func (l *LocalStorage) CreateUploader(ctx context.Context, name string) (Uploader, error) {
-	return nil, nil
+	panic("local storage not support multi-upload")
 }
 
 // Open a Reader by file name.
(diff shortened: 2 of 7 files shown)


pkg/storage/s3.go
```diff
@@ -440,3 +452,64 @@ func (r *s3ObjectReader) Seek(offset int64, whence int) (int64, error) {
 	r.pos = realOffset
 	return realOffset, nil
 }
+
+// ListObject list `maxkeys` objects with specify path.
+func (rs *S3Storage) ListObject(ctx context.Context, name string, maxKeys int64) (*s3.ListObjectsV2Output, error) {
```
kennytm, line 457: can this be unified with `WalkDir`

pkg/storage/s3.go
```diff
@@ -440,3 +452,64 @@ func (r *s3ObjectReader) Seek(offset int64, whence int) (int64, error) {
...
+	input := &s3.ListObjectsV2Input{
+		Bucket:  aws.String(rs.options.Bucket),
+		Prefix:  aws.String(rs.options.Prefix + name),
+		MaxKeys: aws.Int64(maxKeys),
+	}
+	return rs.svc.ListObjectsV2WithContext(ctx, input)
+}
+
+// CreateMultipartUpload create multi upload request.
+func (rs *S3Storage) CreateMultipartUpload(ctx context.Context, name string) (*s3.CreateMultipartUploadOutput, error) {
```
kennytm, line 467: how CDC is using these 3 methods (CreateMultipartUpload, UploadPart, CompleteMultipartUpload)? can we combine them into one API and use [Uploader](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.34.3/service/s3/s3manager?tab=doc#Uploader) instead?

3pointer: CDC use these API int this [PR](https://github.com/pingcap/ticdc/pull/826/files#diff-9ee946c73b42d31cb07ff1fcb4defbaeR348), when flush every batch events, if the size of generate date > 5M. we will use multi-upload to upload data as a chunk. when the upload data size reached 100M, we complete chunks to a file.

Let me see Uploader first.

kennytm: the bottom line is it should not return a `*s3.CreateMultipartUploadOutput` which isn't applicable to GCS or Local.

3pointer: OK, I add a uploader interface, please take a look

pkg/storage/storage.go
```diff
@@ -17,6 +17,14 @@ type ReadSeekCloser interface {
 	io.Closer
 }
 
+// Uploader upload file with chunks.
+type Uploader interface {
+	// UploadPart upload part of file data to storage
+	UploadPart(ctx context.Context, data []byte, partNum int) error
```
kennytm, line 23: what is the purpose of `partNum`? 

as `S3Uploader` does not support running `UploadPart` concurrently, the number can be set as just `len(s.completeParts)+1`.

pkg/storage/gcs.go
```diff
@@ -135,11 +135,16 @@ func (s *gcsStorage) Open(ctx context.Context, name string) (ReadSeekCloser, err
...
 // by path.
-func (s *gcsStorage) WalkDir(ctx context.Context, fn func(string, int64) error) error {
+func (s *gcsStorage) WalkDir(ctx context.Context, dir string, listCount int64, fn func(string, int64) error) error {
 	// TODO, implement this if needed
 	panic("Unsupported Operation")
 }
 
+// CreateUploader implenments ExternalStorage interface.
+func (s *gcsStorage) CreateUploader(ctx context.Context, name string) (Uploader, error) {
+	return nil, nil
```
kennytm, line 145: return a "not implemented" error rather than `nil, nil`.

pkg/storage/local.go
```diff
@@ -56,6 +56,11 @@ func (l *LocalStorage) WalkDir(ctx context.Context, fn func(string, int64) error
 	})
 }
 
+// CreateUploader implenments ExternalStorage interface.
+func (l *LocalStorage) CreateUploader(ctx context.Context, name string) (Uploader, error) {
+	return nil, nil
```
kennytm, line 61: ditto

pkg/storage/noop.go
```diff
@@ -29,10 +29,15 @@ func (*noopStorage) Open(ctx context.Context, name string) (ReadSeekCloser, erro
...
 
 // WalkDir traverse all the files in a dir.
-func (*noopStorage) WalkDir(ctx context.Context, fn func(string, int64) error) error {
+func (*noopStorage) WalkDir(ctx context.Context, dir string, listCount int64, fn func(string, int64) error) error {
 	return nil
 }
 
+// CreateUploader implenments ExternalStorage interface.
+func (*noopStorage) CreateUploader(ctx context.Context, name string) (Uploader, error) {
+	return nil, nil
```
kennytm, line 38: ditto (or return an Uploader which does nothing)

pkg/storage/s3.go
```diff
@@ -75,12 +75,12 @@ type S3Uploader struct {
...
 // UploadPart update partial data to s3, we should call CreateMultipartUpload to start it,
 // and call CompleteMultipartUpload to finish it.
-func (u *S3Uploader) UploadPart(ctx context.Context, data []byte, partNum int) error {
+func (u *S3Uploader) UploadPart(ctx context.Context, data []byte) error {
 	partInput := &s3.UploadPartInput{
 		Body:          bytes.NewReader(data),
 		Bucket:        u.createOutput.Bucket,
 		Key:           u.createOutput.Key,
-		PartNumber:    aws.Int64(int64(partNum)),
+		PartNumber:    aws.Int64(int64(len(u.completeParts))),
```
kennytm, line 83:
```suggestion
		PartNumber:    aws.Int64(int64(len(u.completeParts) + 1)),
```

https://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html

> When uploading a part, in addition to the upload ID, you must specify a part number. You can choose any part number **between 1 and 10,000**.

0 is not a valid part number

pkg/storage/s3.go
```diff
@@ -91,7 +91,7 @@ func (u *S3Uploader) UploadPart(ctx context.Context, data []byte, partNum int) e
 	}
 	u.completeParts = append(u.completeParts, &s3.CompletedPart{
 		ETag:       uploadResult.ETag,
-		PartNumber: aws.Int64(int64(partNum)),
+		PartNumber: aws.Int64(int64(len(u.completeParts))),
```
kennytm, line 94: (ditto)

pkg/storage/s3.go
3pointer: fixed

pkg/storage/s3.go
```diff
@@ -307,13 +363,20 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error)
 // The first argument is the file path that can be used in `Open`
 // function; the second argument is the size in byte of the file determined
 // by path.
-func (rs *S3Storage) WalkDir(ctx context.Context, fn func(string, int64) error) error {
+func (rs *S3Storage) WalkDir(ctx context.Context, dir string, listCount int64, fn func(string, int64) error) error {
 	var marker *string
+	prefix := rs.options.Prefix
+	if len(dir) > 0 {
+		prefix += dir
+	}
```
kennytm, line 371:
```suggestion
	prefix := rs.options.Prefix + dir
```

(or should we use `path.Join`? 🤔)

3pointer: I think `path.Join` is not proper. since we hardcode [prefix](https://github.com/pingcap/br/blob/master/pkg/storage/s3.go#L214)  before

PR #458, opened 2020-08-14
fix restore auto increment ID overflow

3pointer:
when auto_increment id reached max.Int64, it can backs up success. but failed when restore. we should handle this situation.

if autoIncID overflowed. change is to max.Int64, to keep consistency with backup cluster.

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release Note

 - No release note

### DIFF
--- a/pkg/restore/db.go
+++ b/pkg/restore/db.go
@@ -5,6 +5,7 @@ package restore
 import (
 	"context"
 	"fmt"
+	"math"
 	"sort"
 
 	"github.com/pingcap/errors"
@@ -159,11 +160,17 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 		default:
 			alterAutoIncIDFormat = "alter table %s.%s auto_increment = %d;"
 		}
+
+		// auto inc id overflow
+		autoIncID := table.Info.AutoIncID
+		if table.Info.AutoIncID < 0 {
+			autoIncID = math.MaxInt64
+		}
 		restoreMetaSQL = fmt.Sprintf(
 			alterAutoIncIDFormat,
 			utils.EncloseName(table.Db.Name.O),
 			utils.EncloseName(table.Info.Name.O),
-			table.Info.AutoIncID)
+			autoIncID)
 		if utils.NeedAutoID(table.Info) {
 			err = db.se.Execute(ctx, restoreMetaSQL)
 		}
--- a/tests/br_other/run.sh
+++ b/tests/br_other/run.sh
@@ -103,3 +103,20 @@ run_sql "DROP DATABASE $DB;"
 # Test version
 run_br --version
 run_br -V
+
+
+# generate table with auto_inc_id
+run_sql "create table $DB.autoid(`auto_inc_id` bigint(20) NOT NULL AUTO_INCREMENT,UNIQUE KEY `auto_inc_id` (`auto_inc_id`))"
+run_sql "insert into $DB.autoid values(9223372036854775805)"
+
+# backup db
+echo "backup start overflow test..."
+run_br backup db --db "$DB" -s "local://$TEST_DIR/$DB" --pd $PD_ADDR
+
+run_sql "DROP DATABASE $DB;"
+
+# restore db
+echo "restore start..."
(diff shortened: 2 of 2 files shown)


pkg/restore/db.go
```diff
@@ -159,11 +160,17 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 		default:
 			alterAutoIncIDFormat = "alter table %s.%s auto_increment = %d;"
 		}
+
+		// auto inc id overflow
+		autoIncID := table.Info.AutoIncID
+		if table.Info.AutoIncID < 0 {
+			autoIncID = math.MaxInt64
```
kennytm, line 167: what about `bigint unsigned` columns?

```sql
create table a (id bigint unsigned primary key auto_increment);
insert into a values (12345678901234567890);
```

3pointer: I was wondering that the allocator can work well 🤔️，it set the allocator with unsigned=false

3pointer: How about add `globalAutoID -= 1` after https://github.com/pingcap/br/blob/master/pkg/backup/client.go#L277?

kennytm: I'm not sure what's the purpose of minusing 1 😅

pkg/restore/db.go
```diff
@@ -159,11 +160,23 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 		default:
 			alterAutoIncIDFormat = "alter table %s.%s auto_increment = %d;"
 		}
+
+		var autoIncID uint64
+		// auto inc id overflow
+		if table.Info.AutoIncID < 0 {
+			if table.Info.IsAutoIncColUnsigned() {
+				autoIncID = math.MaxUint64
```
kennytm, line 168: this still doesn't look correct. referring to the example before

```sql
create table a (id bigint unsigned primary key auto_increment);
insert into a values (12345678901234567890);
```

the auto_inc_id in the backupmeta is `-6101065172474953725` (12345678901234597891 &minus; 2<sup>64</sup>). 

when we restore we should use `12345678901234597891`, not `18446744073709551615` (MaxUint64).

3pointer: yes, you are right, when I try to fix this, I found that TiDB didn't handle unsigned auto_increment correctly, I file an issue to trace it https://github.com/pingcap/tidb/issues/19238

kennytm: do we need to wait for pingcap/tidb#19241? simply hard-casting as `uint64(table.Info.AutoIncID)` seems to be sufficient for AUTO_INCREMENT_ID (not sure about SEQUENCE 🤔)

pkg/restore/db.go
```diff
@@ -159,11 +159,22 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
 		default:
 			alterAutoIncIDFormat = "alter table %s.%s auto_increment = %d;"
 		}
+
+		var autoIncID uint64
```
glorv, line 163: default should be `table.Info.AutoIncId` if not overflow?

pkg/restore/db.go
kennytm: @3pointer pingcap/tidb#19241 has been merged

overvenus:
Any update?

pkg/restore/db.go
```diff
@@ -108,94 +111,131 @@ func (db *DB) CreateTable(ctx context.Context, table *utils.Table) error {
...
+				log.Info("table auto inc id overflow",
+					zap.Stringer("db", table.DB.Name),
+					zap.Stringer("table", table.Info.Name),
+					zap.Int64("auto inc id", table.Info.AutoIncID),
+				)
+				autoIncID = uint64(table.Info.AutoIncID)
 			} else {
-				setValSQL = fmt.Sprintf(setValFormat, table.Info.Sequence.MaxValue)
+				autoIncID = uint64(table.Info.AutoIncID)
 			}
```
kennytm, line 181: the two branches are the same? so this is simply

```go
    }
    autoIncID = uint64(table.Info.AutoIncID)
```

?

kennytm (commented):
rest LGTM

PR #459, opened 2020-08-17
storage: use path in WalkDir and Open

glorv:
Signed-off-by: glorv <glorvs@163.com>

Use file path in ExtrenalStorage method `WalkDir` and `Open`

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - No release note

### DIFF
--- a/pkg/storage/local.go
+++ b/pkg/storage/local.go
@@ -57,9 +57,9 @@ func (l *LocalStorage) WalkDir(ctx context.Context, fn func(string, int64) error
 	})
 }
 
-// Open a Reader by file path. path is a relative path to base path
-func (l *LocalStorage) Open(ctx context.Context, p string) (ReadSeekCloser, error) {
-	return os.Open(filepath.Join(l.base, p))
+// Open a Reader by file path, path is a relative path to base path.
+func (l *LocalStorage) Open(ctx context.Context, path string) (ReadSeekCloser, error) {
+	return os.Open(filepath.Join(l.base, path))
 }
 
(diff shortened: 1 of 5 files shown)


pkg/storage/local.go
```diff
@@ -52,13 +52,16 @@ func (l *LocalStorage) WalkDir(ctx context.Context, fn func(string, int64) error
...
 
-// Open a Reader by file name.
-func (l *LocalStorage) Open(ctx context.Context, name string) (ReadSeekCloser, error) {
-	return os.Open(path.Join(l.base, name))
+// Open a Reader by file path.
+func (l *LocalStorage) Open(ctx context.Context, p string) (ReadSeekCloser, error) {
+	if !filepath.IsAbs(p) {
+		p = path.Join(l.base, p)
+	}
+	return os.Open(p)
```
kennytm, line 64: this look like a security hole :eyes: why do we support absolute path? (btw use `filepath.Join` instead if `path.Join`.

glorv: the `path` parameter in `filepath.Walk(, func(path, info))`  is absolute path in max osx, and relative path in linux. So I'm not sure if we should be compatible with both

kennytm: use [`filepath.Rel`](https://golang.org/pkg/path/filepath/#Rel) to force the path to be relative to `l.base`?

glorv: done

PR #463, opened 2020-08-17
add s3 api to adapt cdc log (#456)

3pointer:
cherry-pick #456 to release-4.0

### What problem does this PR solve? 
part of https://github.com/pingcap/ticdc/pull/826, to unify all s3 related api in the same repo. we decide to put new s3 api which ticdc needed into br.

### What is changed and how it works?
move ticdc s3 api in br

### Check List 

Tests 

 
 - Manual test (add detailed scripts or steps below)

### Release Note

 - No release note

PR #467, opened 2020-08-21
storage: update WalkDir

3pointer:
resolve https://github.com/pingcap/br/pull/463#discussion_r474395528 
move `dir` `listCount` parameters into context

update WalkDir api

### Check List 

Tests 

 - No code

Related changes

### Release Note

 - No release note

### DIFF
--- a/pkg/storage/s3.go
+++ b/pkg/storage/s3.go
@@ -363,18 +363,12 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error) 
 // The first argument is the file path that can be used in `Open`
 // function; the second argument is the size in byte of the file determined
 // by path.
-func (rs *S3Storage) WalkDir(ctx context.Context, fn func(string, int64) error) error {
+func (rs *S3Storage) WalkDir(ctx context.Context, opt WalkOption, fn func(string, int64) error) error {
 	var marker *string
-	prefix := rs.options.Prefix
+	prefix := rs.options.Prefix + opt.subDir
 	maxKeys := int64(1000)
-
-	dir, ok := ctx.Value("subDir").(string)
-	if ok {
-		prefix += dir
-	}
-	listCount, ok := ctx.Value("listCount").(int64)
-	if ok {
-		maxKeys = listCount
+	if opt.listCount > 0 {
+		maxKeys = opt.listCount
 	}
 
 	req := &s3.ListObjectsInput{
--- a/pkg/storage/s3.go
+++ b/pkg/storage/s3.go
@@ -384,10 +384,10 @@ func (rs *S3Storage) WalkDir(ctx context.Context, opt WalkOption, fn func(string
 			return err
 		}
 		for _, r := range res.Contents {
-			path := *r.Key
-			if opt.RemovePrefix {
-				path = strings.TrimLeft(path, rs.options.Prefix)
-			}
+			// when walk on specify directory, the result include storage.Prefix,
+			// which can not be reuse in other API(Open/Read) directly.
+			// so we use TrimPrefix to filter Prefix for next Open/Read.
+			path := strings.TrimPrefix(*r.Key, rs.options.Prefix)
 			if err = fn(path, *r.Size); err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


pkg/storage/s3.go
```diff
@@ -363,13 +363,20 @@ func (rs *S3Storage) FileExists(ctx context.Context, file string) (bool, error)
...
 // by path.
-func (rs *S3Storage) WalkDir(ctx context.Context, dir string, listCount int64, fn func(string, int64) error) error {
+func (rs *S3Storage) WalkDir(ctx context.Context, fn func(string, int64) error) error {
 	var marker *string
-	prefix := rs.options.Prefix + dir
+	prefix := rs.options.Prefix
 	maxKeys := int64(1000)
-	if listCount > 0 {
+
+	dir, ok := ctx.Value("subDir").(string)
```
glorv, line 371: I think it's better to put these parameters into one struct and provide a func to make the context so we can avoid set the string keys, something like:
```
type s3StoreCtxKey struct{}

type S3Context struct {
	dir string
	listCount int64
}

func WithS3Context(ctx context.Context, s S3Context) context.Context {
	return context.WithValue(ctx, s3StoreCtxKey{}, s)
}
```

glorv:
> it is a bad idea to have behavior-changing arguments hidden inside the context.

Then shall we add another contextual generic param instead of put them in the context?

kennytm:
we should put these into an options struct but there's no need to make an `S3Context` type. 

it's easy to determine the common options between [S3](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html), [GCS](https://cloud.google.com/storage/docs/json_api/v1/objects/list#parameters) and local storage:

* the subdirectory path (prefix in S3 and GCS, sub directory in Local)
* the optimal "page size" (max-keys in S3, maxResults in GCS, ignored in Local)
* a nullable "page token" (continuation-token  in S3, pageToken in GCS, ignored in Local)

the return value of `WalkDir` should also include the next "page token" ( NextContinuationToken  in S3, nextPageToken in GCS, always nil in Local).

pkg/storage/s3.go
```diff
@@ -383,7 +384,11 @@ func (rs *S3Storage) WalkDir(ctx context.Context, opt WalkOption, fn func(string
 			return err
 		}
 		for _, r := range res.Contents {
-			if err = fn(*r.Key, *r.Size); err != nil {
+			path := *r.Key
+			if opt.removePrefix {
+				path = strings.TrimLeft(path, rs.options.Prefix)
```
kennytm, line 389:
```suggestion
				path = strings.TrimPrefix(path, rs.options.Prefix)
```

pkg/storage/storage.go
```diff
@@ -10,6 +10,18 @@ import (
...
+// WalkOption is the option of storage.WalkDir.
+type WalkOption struct {
+	// walk on SubDir of specify directory
+	SubDir string
+	// number of list count, default 1000
+	ListCount int64
+	// this is used for s3 storage, when walk on specify directory,
+	// the result include storage.Prefix, which can not be reuse in other API.(Open/Read),
+	// we can set RemovePrefix to true, to avoid this problem.
+	RemovePrefix bool
```
kennytm, line 22: when do we want this option to be `false`?

3pointer: Maybe in localStroage 🤔️，it depends on our implements of Read/Open API. Since out S3 Read API use prefix internally already.

kennytm: Local storage also uses relative path since #459

3pointer: So,  we can remote it, and then always TrimPrefix for each path?

kennytm: Yes

3pointer: OK, Done

kennytm (commented):
LGTM, though ideally we should get rid of ListCount.

glorv:
LGTM

overvenus (approved):
LGTM

PR #473, opened 2020-08-26
restore: switch mode immediately when restore start.

3pointer:
Fix the issue that switch to import mode not work at first 5 minutes.

do switchToImportMode at first.

### Check List 

Tests 

There is no "switch to import mode" log if we finish restore in 5 minutes.

Related changes

### Release Note

 - Fix the issue that switch to import mode not work at first 5 minutes.


glorv (commented):
LGTM

overvenus (approved):
LGTM

PR #476, opened 2020-08-27
storage: support s3 seek end

glorv:
support seek by `io.SeekEnd` because parquet reader need to read file meta from file footer.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - No release note

### DIFF
--- a/pkg/storage/s3.go
+++ b/pkg/storage/s3.go
@@ -481,15 +481,18 @@ func parseRangeInfo(info *string) (rangeInfo, error) {
 
 	start, err := strconv.ParseInt(subMatches[1], 10, 64)
 	if err != nil {
-		return rangeInfo{}, errors.Annotatef(err, "invalid start offset value '%s' in ContentRange '%s'", subMatches[1], *info)
+		return rangeInfo{}, errors.Annotatef(err,
+			"invalid start offset value '%s' in ContentRange '%s'", subMatches[1], *info)
 	}
 	end, err := strconv.ParseInt(subMatches[2], 10, 64)
 	if err != nil {
-		return rangeInfo{}, errors.Annotatef(err, "invalid end offset value '%s' in ContentRange '%s'", subMatches[2], *info)
+		return rangeInfo{}, errors.Annotatef(err,
+			"invalid end offset value '%s' in ContentRange '%s'", subMatches[2], *info)
 	}
 	size, err := strconv.ParseInt(subMatches[3], 10, 64)
 	if err != nil {
-		return rangeInfo{}, errors.Annotatef(err, "invalid size size value '%s' in ContentRange '%s'", subMatches[3], *info)
+		return rangeInfo{}, errors.Annotatef(err,
+			"invalid size size value '%s' in ContentRange '%s'", subMatches[3], *info)
 	}
 	return rangeInfo{start: start, end: end, size: size}, nil
 }
--- a/pkg/storage/s3.go
+++ b/pkg/storage/s3.go
@@ -504,7 +504,7 @@ type s3ObjectReader struct {
 	reader    io.ReadCloser
 	pos       int64
 	rangeInfo rangeInfo
-	// reader context used for seek
+	// reader context used for implement `io.Seek`
 	ctx context.Context
 }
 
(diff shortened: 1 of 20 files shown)


pkg/storage/s3.go
```diff
@@ -452,6 +453,9 @@ type s3ObjectReader struct {
 	name    string
 	reader  io.ReadCloser
 	pos     int64
+	size    *int64
```
kennytm, line 456: could we parse the Content-Range to retrieve the file size for ranged request (and use Content-Length for non-ranged request)? to save an extra HEAD request.

according to https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_Example_11, if we issue the GET request with

```
            Range: bytes=0-9
```

the reply will contain

```
            Accept-Ranges: bytes
            Content-Range: bytes 0-9/443
            Content-Type: text/plain
            Content-Length: 10
```

so that "433" should be the actual file size.

glorv: I will do some test for this

glorv: In my test, the `AcceptRanges` field return by minio is always `bytes`, not sure if s3 will return the right value

pkg/storage/s3.go
```diff
@@ -404,61 +409,129 @@ func (rs *S3Storage) WalkDir(ctx context.Context, opt *WalkOption, fn func(strin
...
 		return nil, err
 	}
 	return &s3ObjectReader{
-		storage: rs,
-		name:    path,
-		reader:  reader,
+		storage:   rs,
+		name:      path,
+		reader:    reader,
+		ctx:       ctx,
```
kennytm, line 420: Urgh, you're voiding the progress of #452. At this point it's better to change the method to be `SeekContext(ctx context.Context, offset int64, whence int) error` and abandon the Seeker interface.

pkg/storage/s3.go
```diff
@@ -404,61 +409,129 @@ func (rs *S3Storage) WalkDir(ctx context.Context, opt *WalkOption, fn func(strin
...
 func (r *s3ObjectReader) Read(p []byte) (n int, err error) {
-	n, err = r.reader.Read(p)
-	r.pos += int64(n)
-	return
+	maxCnt := r.rangeInfo.end + 1 - r.pos
+	if maxCnt > int64(len(p)) {
+		maxCnt = int64(len(p))
+	}
+	var c int
+	// s3 api may not return enough data, so we need to loop fetch enough
```
kennytm, line 518: why not use [`ReadFull`](https://pkg.go.dev/io/?tab=doc#ReadFull)

pkg/storage/s3.go
glorv: this ctx is mainly using for read parquet files, because parquet reader need open multi reader for each column, if we change this interface, how can we met the parquet file read requirement here: https://github.com/xitongsys/parquet-go/blob/master/source/source.go#L9-L16

pkg/storage/s3.go
```diff
@@ -474,18 +532,19 @@ func (r *s3ObjectReader) Seek(offset int64, whence int) (int64, error) {
 		realOffset = offset
 	case io.SeekCurrent:
 		realOffset = r.pos + offset
+	case io.SeekEnd:
+		realOffset = r.rangeInfo.size + offset
```
3pointer, line 536: why r.rangeInfo.size?🤔️

glorv: info.size is the total file size

3pointer (commented):
LGTM

pkg/storage/s3.go
```diff
@@ -404,59 +409,112 @@ func (rs *S3Storage) WalkDir(ctx context.Context, opt *WalkOption, fn func(strin
...
-	name    string
-	reader  io.ReadCloser
-	pos     int64
+	storage   *S3Storage
+	name      string
+	reader    io.ReadCloser
+	pos       int64
+	rangeInfo rangeInfo
+	// reader context used for seek
+	ctx context.Context
```
kennytm, line 508: please add a comment explaining why `io.Seeker` must be implemented then.

kennytm: @glorv and i mean mention the that Parquet interface.

glorv: I think implement `io.Seeker` is not only for read parquet files in lightning, but also should be convenient for other usage in the future. Put the `xitongsys/parquet-go` interface here seems somewhat weird to me

kennytm: but Parquet is currently the only reason we must use the Seeker interface

glorv: ok, I added a comment for this

kennytm (approved):
lgtm

ti-srebot (approved):

PR #481, opened 2020-08-31
Fix CI failed with "tikv timeout"

YuJuncen:

Fix CI failed with "tikv timeout"

Currently, when a lock resolved, the continuing proto message would be incorrect(for unknown reason).
We would get errors like:
```
grpc: failed to unmarshal the received message proto: wrong wireType = 2 for field UseAsyncCommit
```

After updating `kvproto`, the problem seems resolved magically(!).

### Check List 

Tests 

 - Integration test

### Release Note

 - No release note


YuJuncen:
/run-integration-test

can it resolve the problem stable?

3pointer (approved):
LGTM, we should find which PR bring the fail later

overvenus (approved):
LGTM

ti-srebot (approved):

Issue #440, opened 2020-07-24
Support restore data from TiCDC log


3pointer:
### Describe your feature request related problem:

Part of #325 , After https://github.com/pingcap/ticdc/issues/768 TiCDC support external storage.  BR need to restore data from TiCDC output log.

### Describe the feature you'd like:

Support restore data from TiCDC log.

### Describe alternatives you've considered:

Since there are two kind of TiCDC logs KV changes and DDL events. 
For KV changes, BR  should apply log in multi-thread.

## Workload estimation(person-day)

* Design: 2
* Develop: 4
* QA: 5
* Document: 3

kennytm:
note about merging Lightning and BR:

I oppose blindly combining the two repositories, as each has accumulated a long and diverging history. But I welcome efforts to move the KV Encoder and Local Backend to BR, and let Lightning reuse them as a library (similar to the current Split & Scatter).

PR #482, opened 2020-08-31
Restore: support restore data from cdc log

3pointer:
resolve https://github.com/pingcap/br/issues/440.
This PR is a part of PiTR. with this it can restore data from cdc log backup.

1. Add new command about log restore. which can restore data with specify ts range. It will restore all log backups if there is no specific ts range.
2. To break cycle dependency with other repo. I moved related write and ingest logic here from lightning. which contains `session.go` `manual.go` `bytes.go` `checksum.go`. And the log restore is different from lightining import data. we need not only AddRecord we also need DeleteRecord and UpdateRecord(delete and then add), this implement in `kv.go`.
3. To speed up restore performance. I decide to restore data concurrently at table level. which means one goroutine will restore one table and its related DDL job. this implement in `buffer.go` and `puller.go`.

### Few Questions need to confirm:
1. should we reset commit ts to latest pd ts?
2. how to checksum log data after restore finish?
 

### Check List 

Tests 

1. I built a cdc cluster which can generate log data on minio.
屏幕快照 2020-08-31 下午7 10 34

2. build a simple table at upstream.
```
mysql> admin checksum table t1;
+---------+------------+----------------------+-----------+-------------+
| Db_name | Table_name | Checksum_crc64_xor   | Total_kvs | Total_bytes |
+---------+------------+----------------------+-----------+-------------+
| cdc     | t1         | 10732300587279742478 |     35998 |    37725904 |
+---------+------------+----------------------+-----------+-------------+
1 row in set (0.17 sec)
```
3.  execute `./br restore log  --pd {downstreamIP} -s "s3://cdcs3/log14` and check data restore successfully.

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release Note

 - support restore data from cdc log.


pkg/restore/cdclog/buffer.go
```diff
@@ -0,0 +1,165 @@
...
+			if err != nil {
+				return err
+			}
+			pair, err = t.KvEncoder.AddRecord(newCols, item.RowID, t.colPerm)
+			if err != nil {
+				return errors.Trace(err)
+			}
+			t.KvPairs = append(t.KvPairs, pair)
+			t.Size++
+		} else {
```
glorv, line 120: this code block is duplicated with the if block. The `update` code should be same for insesrt/update event. So we can check if it is update event, then add the deleted keys

pkg/restore/cdclog/buffer.go
```diff
@@ -0,0 +1,165 @@
...
+		cols, err := t.translateToDatum(row.Delete)
+		if err != nil {
+			return err
+		}
+		pair, err := t.KvEncoder.RemoveRecord(cols, item.RowID, t.colPerm)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		t.KvPairs = append(t.KvPairs, pair)
+		t.Size++
```
glorv, line 145: seems these codes in insert/update/delete event are almost the same, so we can makes it a method like:
```
func (t *TAbleBuffer) appendRow(row map[string]column, encodeFn func(cols, item.RowID, t.colPerm) pair) error {
    cols, err := t.translateToDatum(row.Delete)
    if err != nil {
	return err
    }
    pair, err := encodeFn(cols, item.RowID, t.colPerm)
    if err != nil {
        return errors.Trace(err)
    }
    t.KvPairs = append(t.KvPairs, pair)
    t.Size++
}
```

pkg/restore/cdclog/puller.go
```diff
@@ -0,0 +1,177 @@
...
+		ddlFileIndex:        ddlFileIndex,
+		rowChangedFileIndex: rowFileIndex,
+
+		storage: storage,
+	}, nil
+}
+
+// PullOneEvent pulls one event in ts order.
+// The Next event which can be DDL item or Row changed Item depends on next commit ts.
+func (e *EventPuller) PullOneEvent(ctx context.Context) (*SortItem, error) {
```
glorv, line 103: This function actually implement a `mergingIterator` with heap length 2, the big block code for row/dll agains look quite the same

pkg/restore/cdclog/kv.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+// ClassifyAndAppend split KvPairs to data rows and index rows.
+func (kvs KvPairs) ClassifyAndAppend(
+	data *Rows,
+	dataChecksum *KVChecksum,
+	indices *Rows,
+	indexChecksum *KVChecksum,
+) {
+	dataKVs := (*data).(KvPairs)
```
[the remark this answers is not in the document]
3pointer: How to implement the `Append` method?🤔️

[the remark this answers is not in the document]
YuJuncen: Maybe we could implement `Row` and `Rows` for `*KvPair`...? 🤔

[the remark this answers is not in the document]
3pointer: I'm not sure about what you mean, could you give the signature of `Append`?

pkg/restore/cdclog/decoder.go
```diff
@@ -0,0 +1,303 @@
...
+		return nil, errors.Trace(err)
+	}
+	item := &SortItem{
+		ItemType: DDL,
+		Meta:     ddlMsg,
+		Schema:   b.nextKey.Schema,
+		Table:    b.nextKey.Table,
+		TS:       b.nextKey.Ts,
+	}
+	b.nextKey = nil
```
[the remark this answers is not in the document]
3pointer: Done, PTAL

pkg/restore/cdclog/decoder.go
```diff
@@ -0,0 +1,303 @@
...
+// Decode the Row message.
+func (m *MessageRow) Decode(data []byte) error {
+	decoder := json.NewDecoder(bytes.NewReader(data))
+	decoder.UseNumber()
+	err := decoder.Decode(m)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	for colName, column := range m.Update {
+		m.Update[colName] = formatColumnVal(column)
```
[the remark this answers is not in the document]
3pointer: yes, it's as same as other decoder implements in cdc https://github.com/pingcap/ticdc/blob/bdfeb7d56c6ddf2d88e0a7caeaae90207aa811ae/cdc/sink/codec/json.go#L150

pkg/restore/cdclog/kv.go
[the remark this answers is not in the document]
YuJuncen: The `KvPair` is a typo to `KvPairs`, sorry about it. 😿

Let me explain:

We can make a new method on `Rows`:
```
	// Append append a KV pair to the collection.
	Append(pair KvPair)
```

And we can implement `Rows` and `Row` for `*KvPairs`(instead of `KvPairs`), so we can:
```
func (kvs *KvPairs) Append(pair KvPair) {
	*kvs = append(*kvs, pair) 
}
```

pkg/restore/log_client.go
```diff
@@ -0,0 +1,902 @@
...
+				continue
+			}
+
+			// wait all previous kvs ingest finished
+			err = l.applyKVChanges(ctx, tableID)
+			if err != nil {
+				return errors.Trace(err)
+			}
+
+			err = l.restoreClient.db.se.Execute(ctx, ddl.Query)
```
glorv, line 741: After executing ddl job, maybe table columns and column permutations should also be updated?

pkg/restore/cdclog/kv.go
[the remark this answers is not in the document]
YuJuncen: The advantages of this are the internal `Rows` implementation won't be leaked to the outer world. However, it also introduces the two-level pointer overhead. If this isn't acceptable or you believe the abstraction leakage doesn't matter, we can also ignore this for now.

pkg/restore/cdclog/decoder.go
```diff
@@ -0,0 +1,273 @@
...
+	for colName, column := range m.PreColumns {
+		m.PreColumns[colName] = formatColumnVal(column)
+	}
+	return nil
+}
+
+// SortItem represents a DDL item or Row changed item.
+type SortItem struct {
+	ItemType ItemType
+	Meta     interface{}
```
YuJuncen, line 187: Seems this field stores the "content" of this `SortItem`. Hence the name `Meta` is someway misleading, how about renaming it to `Message` or `Data`?

pkg/restore/cdclog/buffer.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Append appends the item to this buffer.
+func (t *TableBuffer) Append(ctx context.Context, item *SortItem) error {
+	log.Debug("Append item to buffer",
+		zap.Stringer("table", t.tableInfo.Meta().Name),
+	)
+	row := item.Meta.(*MessageRow)
+
+	select {
+	case <-ctx.Done():
```
glorv, line 126: why need to do a select here, this append should be quite fast, are there any blocks? the `Context` parameter is only used in the select check, I think this isn't meaningful

pkg/restore/log_client.go
```diff
@@ -0,0 +1,935 @@
...
+		startKey := codec.EncodeBytes([]byte{}, pairStart)
+		endKey := codec.EncodeBytes([]byte{}, nextKey(pairEnd))
+		regions, err = PaginateScanRegion(ctx, l.splitClient, startKey, endKey, 128)
+		if err != nil || len(regions) == 0 {
+			log.L().Warn("scan region failed", zap.Error(err), zap.Int("region_len", len(regions)))
+			continue WriteAndIngest
+		}
+
+		shouldWait := false
+		errChan := make(chan error, len(regions))
```
YuJuncen, line 585: Maybe we can use `errgroup` here? 🤔

pkg/restore/log_client.go
```diff
@@ -0,0 +1,902 @@
...
+	for _, meta := range metas {
+		for i := 0; i < maxRetryTimes; i++ {
+			log.L().Debug("ingest meta", zap.Reflect("meta", meta))
+			resp, err := l.Ingest(ctx, meta, region)
+			if err != nil {
+				log.L().Warn("ingest failed", zap.Error(err), zap.Reflect("meta", meta),
+					zap.Reflect("region", region))
+				continue
+			}
+			needRetry, newRegion, errIngest := isIngestRetryable(resp, region, meta)
```
YuJuncen, line 530: We could use `multierr` to collect errors here.

pkg/restore/cdclog/buffer.go
3pointer: context has been removed.

pkg/restore/cdclog/decoder.go
3pointer: Done, move it to `Data`

pkg/restore/log_client.go
3pointer: sure, fixed it

pkg/restore/log_client.go
```diff
@@ -0,0 +1,929 @@
...
+
+	// collect ddl files
+	ddlFiles, err := l.collectDDLFiles(ctx)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	// execute create ddl first, because we need tableInfo to create kvEncoder
+	// we will filter all create ddl in later eventPuller
+	err = l.doCreateDDLJob(ctx, ddlFiles)
```
glorv, line 842: I think there maybe some problems here. If there are multi create/drop DDLs with the same table name in ddlFiles, execute all the create ddl make fails or cause unexpected errors

3pointer: You are right, I removed these create logic, and change the logic into `ReloadMeta`. 
every time after `ReloadMeta`, it will set kvEncoder to nil. and in `Append` if it found kvEncoder is nil, it will recreate encoder for next row.

pkg/restore/cdclog/buffer.go
```diff
@@ -0,0 +1,208 @@
...
+			return nil, errors.Trace(err)
+		}
+		return NewTableKVEncoder(encTable, &SessionOptions{
+			Timestamp: time.Now().Unix(),
+			// TODO get the version from TiDB cluster
+			// currently TiDB only support v1 and v2, and since 4.0
+			// the default RowFormatVersion is 2, so I think
+			// we can implement the row version retrieve from cluster in the future
+			// when TiDB decide to support v3 RowFormatVersion.
+			RowFormatVersion: "2",
```
glorv, line 59: why not directly read this info from tidb cluster like lightning do?

pkg/restore/cdclog/decoder.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	return msgKey, keyLen, nil
+}
+
+// NextEvent return next item depends on type.
+func (b *JSONEventBatchMixedDecoder) NextEvent(itemType ItemType) (*SortItem, error) {
+	if !b.HasNext() {
+		return nil, nil
+	}
+	nextKey, nextKeyLen, err := b.decodeNextKey()
```
glorv, line 224: the nextKeyLen is only used here, so we can directly set b.mixedBytes = b.mixedBytes[nextKeyLen+8:] in decodeNextKey, then we needn't return the nextKeyLen

pkg/restore/log_client.go
```diff
@@ -0,0 +1,1070 @@
...
+				return errors.Trace(err)
+			}
+			dbInfo, ok := dom.InfoSchema().SchemaByName(model.NewCIStr(schema))
+			if !ok {
+				return errors.Errorf("schema %s not exists", schema)
+			}
+			allocs = autoid.NewAllocatorsFromTblInfo(dom.Store(), dbInfo.ID, tableInfo.Meta())
+		}
+
+		l.tableBuffers[tableID] = cdclog.NewTableBuffer(tableInfo, allocs, l.concurrencyCfg.BatchFlushKVPairs)
```
glorv, line 1016: If the `infoSchema.TableExists` is false, then the tableInfo should be nil here, should add a check here

pkg/task/restore_log.go
```diff
@@ -0,0 +1,127 @@
...
+// adjustRestoreConfig is use for BR(binary) and BR in TiDB.
+// When new config was add and not included in parser.
+// we should set proper value in this function.
+// so that both binary and TiDB will use same default value.
+func (cfg *LogRestoreConfig) adjustRestoreConfig() {
+	if cfg.Config.Concurrency == 0 {
+		cfg.Config.Concurrency = defaultRestoreConcurrency
+	}
+	if cfg.BatchFlushKVPairs == 0 {
+		cfg.BatchFlushKVPairs = defaultFlushKV
```
glorv, line 76: Shall we make the flush triggered both by kv pairs and total size? In the case if each kv is quite large (maybe upto sevral MB), flush only by kv pairs maybe not sufficient

pkg/restore/cdclog/buffer.go
3pointer: I explain it in above comment.

3pointer: Currently we don't have tidb mgr like lightning in BR. BR use tidb.Domain to interact with TiKV. it's not easy to implement it in BR like lightning.

pkg/restore/log_client.go
3pointer: yes, the check already in `NewTableBuffer`

glorv: Then why still need the `infoSchema.TableExists` check

pkg/task/restore_log.go
3pointer: I add the flush size config to control the flush

pkg/restore/log_client.go
3pointer: If `infoSchema.TableExists` we can build allocators and tableInfo at the beginning.
if not exists, then we can not handle allocators and tableInfo here, we should wait for next DDL happen to Reload.

pkg/restore/cdclog/decoder.go
3pointer: Done, PTAL

pkg/restore/log_client.go
```diff
@@ -0,0 +1,1074 @@
...
+	}
+	if names[0] != ddlFilePrefix {
+		log.Warn("file doesn't start with ddl", zap.String("file", fileName))
+		return false, nil
+	}
+	ts, err := strconv.ParseUint(names[1], 10, 64)
+	if err != nil {
+		return false, errors.AddStack(err)
+	}
+	ts = maxUint64 - ts
```
glorv, line 231: please add a comment here for the need to do xor here

glorv:
please signed a issue to describe the plan the solve the ddl charset&collation issue and add some integration test for this feature.

Rest, LGTM

3pointer:
> please sign a issue to describe the plan the solve the ddl charset&collation issue and add some integration test for this feature.
> 
> Rest LGTM
okay, issues have been filed. 
collation & charset issue: https://github.com/pingcap/ticdc/issues/951
integration test issue: https://github.com/pingcap/br/issues/504

pkg/restore/cdclog/puller.go
```diff
@@ -0,0 +1,181 @@
...
+			if err != nil {
+				return nil, errors.Trace(err)
+			}
+		} else {
+			returnItem = e.currentRowChangedItem
+			e.currentRowChangedItem, err = e.rowChangedDecoder.NextEvent(RowChanged)
+			if err != nil {
+				return nil, errors.Trace(err)
+			}
+		}
```
kennytm, line 170:
```suggestion
			break
		}
		fallthrough
```

pkg/restore/cdclog/puller.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	var returnItem *SortItem
+	switch {
+	case e.currentDDLItem != nil:
+		if e.currentDDLItem.LessThan(e.currentRowChangedItem) {
```
kennytm, line 158: is it true that `e.currentRowChangedItem` must be non-nil here?

pkg/restore/log_client.go
```diff
@@ -0,0 +1,1079 @@
...
+	cancel()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	// Cache the conn.
+	l.grpcClis.clis[storeID] = conn
+	return conn, nil
+}
+
+func (l *LogClient) getImportClient(ctx context.Context, peer *metapb.Peer) (sst.ImportSSTClient, error) {
```
kennytm, line 185: why are we not reusing `pkg/restore/import.go`

pkg/restore/log_client.go
```diff
@@ -0,0 +1,1079 @@
...
+
+	// According to https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysUsingAPIs.html
+	// list API return in UTF-8 binary order, so the cdc log create DDL file used
+	// maxUint64 - the first DDL event's commit ts as the file name to return the latest ddl file.
+	// see details at https://github.com/pingcap/ticdc/pull/826/files#diff-d2e98b3ed211b7b9bb7b6da63dd48758R81
+	ts = maxUint64 - ts
+	if l.tsInRange(ts) {
+		return true, nil
+	}
+	return false, nil
```
kennytm, line 240:
```suggestion
	return l.tsInRange(ts), nil
```

pkg/restore/log_client.go
```diff
@@ -0,0 +1,1079 @@
...
+		}
+	}
+	return sst.NewImportSSTClient(conn), nil
+}
+
+func (l *LogClient) tsInRange(ts uint64) bool {
+	if ts < l.startTS || ts > l.endTs {
+		return false
+	}
+	return true
```
kennytm, line 205:
```suggestion
	return l.startTS <= ts && ts <= l.endTs
```

why the capitalization of the two TS are different 🤔

pkg/restore/log_client.go
```diff
@@ -0,0 +1,1079 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	sort.Slice(ddlFiles, func(i, j int) bool {
+		return ddlFiles[i] > ddlFiles[j]
+	})
```
kennytm, line 266:
```suggestion
	sort.Sort(sort.Reverse(sort.StringSlice(ddlFiles)))
```

pkg/restore/log_client.go
```diff
@@ -0,0 +1,1079 @@
...
+	var startKey, endKey []byte
+	if len(region.Region.StartKey) > 0 {
+		_, startKey, _ = codec.DecodeBytes(region.Region.StartKey, []byte{})
+	}
+	if len(region.Region.EndKey) > 0 {
+		_, endKey, _ = codec.DecodeBytes(region.Region.EndKey, []byte{})
+	}
+
+	var start, end int
+	// TODO use binary sort
```
kennytm, line 566:
```suggestion
	// TODO use binary search
```

pkg/restore/split.go
```diff
@@ -338,3 +339,28 @@ func replacePrefix(s []byte, rewriteRules *RewriteRules) ([]byte, *import_sstpb.
 
 	return s, nil
 }
+
+func beforeEnd(key []byte, end []byte) bool {
+	return bytes.Compare(key, end) < 0 || len(end) == 0
+}
+
+func keyInsideRegion(region *metapb.Region, key []byte) bool {
+	return bytes.Compare(key, region.GetStartKey()) >= 0 && (beforeEnd(key, region.GetEndKey()))
```
kennytm, line 348:
```suggestion
	return bytes.Compare(key, region.GetStartKey()) >= 0 && beforeEnd(key, region.GetEndKey())
```

pkg/restore/util.go
```diff
@@ -524,3 +524,23 @@ func ZapRanges(ranges []rtree.Range) []zapcore.Field {
...
+		}
+	}
+	schema := string(schemaRune)
+	return unQuoteName(schema), unQuoteName(strings.TrimPrefix(name, schema+"."))
+}
+
+func unQuoteName(name string) string {
+	name = strings.TrimLeft(name, "`")
+	return strings.TrimRight(name, "`")
+}
```
kennytm, line 546: i don't think this implementation produces the correct unquoting when the table name contains interior `` ` ``.

```go
schema, table := parseQuoteName("`a``b`.``````")
c.Assert(schema, Equals, "a`b")
c.Assert(table, Equals, "``")
```

cmd/restore.go
```diff
@@ -124,6 +138,19 @@ func newTableRestoreCommand() *cobra.Command {
 	return command
 }
 
+func newLogRestoreCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "log",
+		Short: "restore data from log backup",
```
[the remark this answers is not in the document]
3pointer: `cdclog` make sense.

pkg/restore/cdclog/decoder.go
```diff
@@ -0,0 +1,273 @@
...
+	GeneratedColumnFlag
+	// PrimaryKeyFlag means the column is primary key.
+	PrimaryKeyFlag
+	// UniqueKeyFlag means the column is unique key.
+	UniqueKeyFlag
+	// MultipleKeyFlag means the column is multiple key.
+	MultipleKeyFlag
+	// NullableFlag means the column is nullable.
+	NullableFlag
+)
```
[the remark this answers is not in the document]
3pointer: should I leave a TODO here?
And then give a PR for cdc to import BR.

pkg/restore/cdclog/kv.go
```diff
@@ -0,0 +1,360 @@
...
+	// RemoveRecord encode encodes a row of SQL values into a backend-friendly format.
+	RemoveRecord(
+		row []types.Datum,
+		rowID int64,
+		columnPermutation []int,
+	) (Row, int, error)
+}
+
+// Row represents a single encoded row.
+type Row interface {
```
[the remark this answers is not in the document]
3pointer: Ok, I'll remove it.

pkg/restore/cdclog/puller.go
3pointer: if it's nil, LessThan will return true.

pkg/restore/log_client.go
3pointer: It can be reused, I moved the grpc config setting in `pkg/restore/import.go` in this commit https://github.com/pingcap/br/pull/482/commits/1b6de45d4d111186e3eb43ee786816658e34d52d

pkg/restore/util.go
3pointer: yes, the corner case seems not easy. do you have some advices? there already have the solution.

pkg/restore/cdclog/decoder.go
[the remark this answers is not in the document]
kennytm: yes

afaik CDC already imported BR.

pkg/restore/cdclog/kv.go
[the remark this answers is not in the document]
kennytm: Yes. basically implement `ÇlassifyAndAppend` directly on `KvPairs` and take `*KvPairs` as input

pkg/restore/util.go
kennytm: You could regexp match for ``` `(?:[^`]|``)*` ``` and extract each part of name with ```strings.ReplaceAll(result[1:len(result)-1], "``", "`")```.

pkg/restore/util_test.go
```diff
@@ -21,6 +21,24 @@ var _ = Suite(&testRestoreUtilSuite{})
...
+	c.Assert(schema, Equals, "a`b")
+	c.Assert(table, Equals, "``")
+
+	schema, table = restore.ParseQuoteName("`.`.`.`")
+	c.Assert(schema, Equals, ".")
+	c.Assert(table, Equals, ".")
+
+	// schema, table = restore.ParseQuoteName("`.``.`.`.`")
+	// c.Assert(schema, Equals, ".`.")
+	// c.Assert(table, Equals, ".")
```
kennytm, line 39: 🤔

pkg/restore/util.go
3pointer: Done, PTAL again

pkg/restore/cdclog/decoder.go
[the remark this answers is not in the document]
3pointer: ok, I filed an issue https://github.com/pingcap/ticdc/issues/956 talk about this.

pkg/restore/bytes.go
```diff
@@ -0,0 +1,121 @@
...
+// recycleChan is used for reusing allocated []byte so we can use memory more efficiently
+//
+// NOTE: we don't used a `sync.Pool` because when will sync.Pool release is depending on the
+// garbage collector which always release the memory so late. Use a fixed size chan to reuse
+// can decrease the memory usage to 1/3 compare with sync.Pool.
+var recycleChan = &bytesRecycleChan{
+	ch: make(chan []byte, 1024),
+}
+
+// Acquire ...
```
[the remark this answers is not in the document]
3pointer: Done

cmd/restore.go
```diff
@@ -124,6 +138,19 @@ func newTableRestoreCommand() *cobra.Command {
...
+func newLogRestoreCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "cdclog",
+		Short: "restore data from cdc log backup",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			return runLogRestoreCommand(cmd)
+		},
+	}
+	task.DefineFilterFlags(command)
+	task.DefineLogRestoreFlags(command)
```
overvenus, line 150: Could you hide the subcommand until it's ready?

pkg/restore/manual/manual.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2020 The LevelDB-Go and Pebble Authors. All rights reserved. Use
```
overvenus, line 1: Please move this package to the `pkg` folder.

pkg/restore/cdclog/session.go
```diff
@@ -0,0 +1,251 @@
...
+package cdclog
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/pingcap/parser/model"
+	"github.com/pingcap/tidb/sessionctx"
+
```
overvenus, line 23:
```suggestion
```

pkg/restore/cdclog/decoder.go
```diff
@@ -0,0 +1,274 @@
+// Copyright 2020 PingCAP, Inc.
```
overvenus, line 1: Please move this package to the `pkg` folder.

overvenus (commented):
This PR is too large, to improve review efficiency, please split it into smaller PR. E.g., you can send Lightning code separately.

3pointer:
> This PR is too large, to improve review efficiency, please split it into smaller PR. E.g., you can send Lightning code separately.

I file a PR for lightning code. https://github.com/pingcap/br/pull/512, and after this merge into master, I can update this branch with master.

3pointer:
@kennytm @overvenus PTAL ~

pkg/utils/kv.go
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package cdclog
+package utils
```
kennytm, line 14: are these renames intentional? 🤔

3pointer: yes, not only br cdclog will use `kv.go`, lightning still need use it. so I put it in `utils` pkg.

overvenus: How about putting it into a new package? I don't want to package utils to be too bloat.

3pointer: OK, I put it into `kv` pkg, because these code are all related to kv.

pkg/restore/log_client.go
```diff
@@ -0,0 +1,1008 @@
...
+		}
+		if err = wstream.Send(req); err != nil {
+			return nil, err
+		}
+		req.Chunk = &sst.WriteRequest_Batch{
+			Batch: &sst.WriteBatch{
+				// FIXME we should discuss about the commit ts
+				// 1. give a batch of kv a specify commit ts
+				// 2. give each kv the backup commit ts
+				CommitTs: oracle.ComposeTS(time.Now().Unix()*1000, 0),
```
kennytm, line 383: 🤔 this mean every tikv peer will receive a different commit TS? how about move the calculation out of the loop?

pkg/cdclog/buffer.go
```diff
@@ -0,0 +1,215 @@
...
+
+// TableBuffer represents the kv buffer of this table.
+// we restore one tableBuffer in one goroutine.
+// this is the concurrent unit of log restore.
+type TableBuffer struct {
+	KvPairs []kv.Row
+	count   int
+	size    int64
+
+	KvEncoderFn func(autoid.Allocators, table.Table) (kv.Encoder, error)
```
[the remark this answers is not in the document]
3pointer: eh. settable member for lazy create and maybe for mock in the future? I can change it to private member.

pkg/restore/log_client.go
3pointer: sure, I'll move it out of loop, to make every peer receive the same commit ts.

pkg/cdclog/buffer.go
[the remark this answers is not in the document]
kennytm: Keeping KvEncoder public should be enough for mocking since you could fill in a non-nil mock instance, unless you do want to test the lazy init behavior.

pkg/cdclog/buffer.go
```diff
@@ -0,0 +1,215 @@
...
+
+// TableBuffer represents the kv buffer of this table.
+// we restore one tableBuffer in one goroutine.
+// this is the concurrent unit of log restore.
+type TableBuffer struct {
+	KvPairs []kv.Row
+	count   int
+	size    int64
+
+	kvEncoderFn func(autoid.Allocators, table.Table) (kv.Encoder, error)
```
kennytm, line 37: as this is now just a write-once private member perhaps we could simply make this a method?

kennytm:
rest LGTM

pkg/cdclog/buffer.go
3pointer: OK, Done

kennytm (approved):

PR #483, opened 2020-09-01
*: don't remove tiflash more.

YuJuncen:

Historically, we must remove all TiFlash nodes or the restore progress sticks. How ever, for now, TiFlash can support all RPCs BR needs.
The code handling TiFlash nodes costs and can bring many useless logs. It's time to remove it.

I removed all (I hope) logic removing the TiFlash nodes.

### Check List 

Tests 

 - Integration test

### Release Note

 - BR doesn't remove TiFlash nodes anymore.

### DIFF
--- a/pkg/utils/version.go
+++ b/pkg/utils/version.go
@@ -97,6 +97,7 @@ func CheckClusterVersion(ctx context.Context, client pd.Client) error {
 	for _, s := range stores {
 		isTiFlash := IsTiFlash(s)
 		log.Debug("checking compatibility of store in cluster",
+			zap.Uint64("ID", s.GetId()),
 			zap.Bool("TiFlash?", isTiFlash),
 			zap.String("address", s.GetAddress()),
 			zap.String("version", s.GetVersion()),
--- a/pkg/utils/version.go
+++ b/pkg/utils/version.go
@@ -10,6 +10,8 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/pingcap/kvproto/pkg/metapb"
+
 	"github.com/coreos/go-semver/semver"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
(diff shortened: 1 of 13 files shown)


3pointer:
It's better to have an integration test for this change. since we need support other RPC(#482 ) in later version.

3pointer:
> It's better to have an integration test for this change. since we need to make sure TiFlash support other RPC(#482 ) in later version.

oh, I saw it in `br_tiflash`

YuJuncen:
> > It's better to have an integration test for this change. since we need to make sure TiFlash support other RPC(#482 ) in later version.
> 
> oh, I saw it in `br_tiflash`

Yep, however, tiflash in this test case still cannot sync with TiDB till now. 

But even though, I think, since the TiFlash nodes did be in the cluster, the RPC compatibility could still be tested.

kennytm:
we should add a version check for TiFlash in `CheckClusterVersion()` (do we have the exact version where BR is supported? 4.0.4?)

YuJuncen:
> we should add a version check for TiFlash in `CheckClusterVersion()` (do we have the exact version where BR is supported? TiFlash 4.0.4?)

An embarrassing fact is that I don't really know from which version TiFlash supports BR. 🤥

Anyway, `v4.0.4` would be a safe choice.

kennytm (commented):
LGTM

pkg/utils/version.go
```diff
@@ -72,6 +95,12 @@ func CheckClusterVersion(ctx context.Context, client pd.Client) error {
 		return errors.Trace(err)
 	}
 	for _, s := range stores {
+		if IsTiFlash(s) {
```
3pointer, line 98: we can add some debug log here

3pointer (approved):
Rest LGTM

ti-srebot (approved):

PR #525, opened 2020-09-23
pkg: make TiDB not dependent on gcc

jackysp:
Signed-off-by: Shuaipeng Yu <jackysp@gmail.com>

br doesn't use CGO, but TiDB does. `pkg/manual/manual.go` makes TiDB dependent gcc when compiling, which is not necessary.

Use native go allocator for all tags.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

### Release Note

 - No release note


glorv (approved):
LGTM

3pointer (approved):
LGTM

ti-srebot (approved):

kennytm:
@glorv so is `manual.go` still needed for Lightning? if not i suggest we just remove this entire package

glorv:
> @glorv so is `manual.go` still needed for Lightning? if not i suggest we just remove this entire package

Yes, I manually tests latest lighting with/without manual.go , and  the memory usage and performance have no significant differenct, so I think it's ok to remove the `manual` package from lightning. I'll file a pr to do this later.

PR #531, opened 2020-09-25
adjust rawURL to local path if no u.Schema is given

lichunzhu:
1. If we don't specify for `rawURL`, it will return an error. 
2. If we use a relative path for `rawURL`, it will return a wrong path.

1. adjust rawURL to local path if no u.Schema is given
2. Use `filepath.Abs` to make sure path is transferred.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release Note

 - No release note

### DIFF
--- a/pkg/storage/parse_test.go
+++ b/pkg/storage/parse_test.go
@@ -106,8 +106,11 @@ func (r *testStorageSuite) TestCreateStorage(c *C) {
 	c.Assert(gcs.Prefix, Equals, "backup/")
 	c.Assert(gcs.CredentialsBlob, Equals, "fakeCreds2")
 
-	_, err = ParseBackend("/test", nil)
+	s, err = ParseBackend("/test", nil)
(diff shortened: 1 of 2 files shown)


pkg/storage/parse_test.go
```diff
@@ -105,6 +105,9 @@ func (r *testStorageSuite) TestCreateStorage(c *C) {
 	c.Assert(gcs.Bucket, Equals, "bucket4")
 	c.Assert(gcs.Prefix, Equals, "backup/")
 	c.Assert(gcs.CredentialsBlob, Equals, "fakeCreds2")
+
+	_, err = ParseBackend("/test", nil)
+	c.Assert(err, IsNil)
```
kennytm, line 110: also test that the ParseBackend result has `GetLocal`.

kennytm (commented):
rest LGTM

3pointer (approved):
LGTM

Issue #534, opened 2020-09-25
Restore raw doesn't work on a newly created cluster


fschoell:
I created a backup of my Tikv cluster using `br backup raw`. When trying to restore it to a newly created cluster the br process just hangs at `"switch to import mode"`. 

Backup created using:

`br backup raw --cf default --pd "172.22.5.120:2379" --s3.provider "ceph" --s3.endpoint "http://ceph-gw12.xxx.io:7480" -s "s3://tikv-backup/2020-09-24" --send-credentials-to-tikv=true --s3.region "none"  --ratelimit 20 --log-file /tmp/tikv-backup.log --start 00 --end ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`

Restore using:

`br restore raw --pd "172.22.5.178:2379" --cf default --s3.provider "ceph" --s3.endpoint "http://ceph-gw12.xxx.io:7480" -s "s3://tikv-backup/2020-09-24" --send-credentials-to-tikv=true --s3.region "none"  --ratelimit 20 --log-file /tmp/tikv-backup.log --start 00 --end ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff -L "debug"`

Restore logs from br:

```
[2020/09/25 08:40:04.356 +00:00] [INFO] [version.go:33] ["Welcome to Backup & Restore (BR)"]
[2020/09/25 08:40:04.357 +00:00] [INFO] [version.go:34] [BR] [release-version=v4.0.6]
[2020/09/25 08:40:04.357 +00:00] [INFO] [version.go:35] [BR] [git-hash=4e68a8d0333ff786911e15a4354e6f4a23d79386]
[2020/09/25 08:40:04.357 +00:00] [INFO] [version.go:36] [BR] [git-branch=heads/refs/tags/v4.0.6]
[2020/09/25 08:40:04.357 +00:00] [INFO] [version.go:37] [BR] [go-version=go1.13]
[2020/09/25 08:40:04.357 +00:00] [INFO] [version.go:38] [BR] [utc-build-time="2020-09-16 03:43:23"]
[2020/09/25 08:40:04.357 +00:00] [INFO] [version.go:39] [BR] [race-enabled=false]
[...]

PR #535, opened 2020-09-27
raw_restore: set concurrency to default when no config here

YuJuncen:

When we don't provide the `concurrency` config in raw restoring, the `file` worker pool size would be 0, and then block all the restore process.

probably relative to #534 

Set concurrency to default when no config here.

### Check List 

Tests 

 - Integration test

### Release Note

 - Fixed a bug that caused restoring blocks when no `concurrency` provided.


kennytm (commented):
LGTM

glorv:
LGTM

ti-srebot (approved):

PR #542, opened 2020-09-29
Support the AWS_REGION env var

tirsen:

Only override the region if it has been explicitly set, otherwise we clobber the standard AWS_REGION env var

https://github.com/pingcap/dumpling/pull/155#issuecomment-700136706

Support the standard AWS_REGION env var

### Check List 

Tests 

### DIFF
--- a/pkg/storage/s3.go
+++ b/pkg/storage/s3.go
@@ -228,8 +228,10 @@ func NewS3Storage( // revive:disable-line:flag-parameter
 	qs := *backend
 	awsConfig := aws.NewConfig().
 		WithMaxRetries(maxRetries).
-		WithS3ForcePathStyle(qs.ForcePathStyle).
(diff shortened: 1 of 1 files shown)


kennytm (commented):
please add a test case to `s3_test.go` (extending `TestApplyUpdate` should be good)

rest LGTM

tirsen:
Ah yes I removed the hard coded `us-east-1` and updated the tests. This somewhat breaks backwards compatibility though but probably not in a bad way. Hardcoding `us-east-1` could be disastrous.

kennytm:
i think we need a 

```go
if len(aws.StringValue(ses.Config.Region)) == 0 {
    ses.Config.Region = aws.String("please supply --s3.region")
}
```

after line 250 to prevent that `MissingRegion` error thrown by the AWS SDK.

PR #545, opened 2020-09-30
conn,task: add config about keepalive

YuJuncen:
Currently, keepalive time and timeout in both TiKV and TiDB are configurable. According to the `keepalive.ClientParameters`:

> Make sure these parameters are set in coordination with the keepalive policy on the server, as incompatible settings can result in closing of connection.

We should make those variables in BR configurable too so when they are changed, BR won't make confusing behavior.

Added two flags: `grpc-keepalive-time` and `grpc-keepalive-timeout`.

### Check List 

Tests 

 - Integration test 

### Release Note

 - Added config `grpc-keepalive-time` and `grpc-keepalive-timeout`.

### DIFF
--- a/pkg/task/common.go
+++ b/pkg/task/common.go
@@ -159,6 +159,8 @@ func DefineCommonFlags(flags *pflag.FlagSet) {
 		"the interval of pinging the server, must keep same value with TiKV and PD")
 	flags.Duration(flagGrpcKeepaliveTimeout, defaultGRPCKeepaliveTimeout,
 		"the max time a grpc conn can keep idel before killed, must keep same value with TiKV and PD")
+	_ = flags.MarkHidden(flagGrpcKeepaliveTime)
+	_ = flags.MarkHidden(flagGrpcKeepaliveTimeout)
 
 	storage.DefineFlags(flags)
 }
--- a/pkg/task/common.go
+++ b/pkg/task/common.go
@@ -21,6 +21,7 @@ import (
 	pd "github.com/tikv/pd/client"
 	"go.etcd.io/etcd/pkg/transport"
 	"go.uber.org/zap"
+	"google.golang.org/grpc/keepalive"
 
 	"github.com/pingcap/br/pkg/conn"
 	berrors "github.com/pingcap/br/pkg/errors"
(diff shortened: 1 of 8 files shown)


pkg/task/common.go
```diff
@@ -55,8 +56,14 @@ const (
 	flagRemoveTiFlash      = "remove-tiflash"
 	flagCheckRequirement   = "check-requirements"
 	flagSwitchModeInterval = "switch-mode-interval"
-
-	defaultSwitchInterval = 5 * time.Minute
+	// flagGrpcKeepaliveTime is the interval of pinging the server.
+	flagGrpcKeepaliveTime = "grpc-keepalive-time"
+	// flagGrpcKeepaliveTimeout is the max time a grpc conn can keep idel before killed.
```
kennytm, line 61:
```suggestion
	// flagGrpcKeepaliveTimeout is the max time a grpc conn can keep idle before killed.
```

pkg/task/common.go
```diff
@@ -143,6 +155,10 @@ func DefineCommonFlags(flags *pflag.FlagSet) {
 	flags.Bool(flagCheckRequirement, true,
 		"Whether start version check before execute command")
 	flags.Duration(flagSwitchModeInterval, defaultSwitchInterval, "maintain import mode on TiKV during restore")
+	flags.Duration(flagGrpcKeepaliveTime, defaultGRPCKeepaliveTime,
+		"the interval of pinging the server, must keep same value with TiKV and PD")
```
kennytm, line 159: what is "the server"

pkg/task/common.go
```diff
@@ -143,6 +155,10 @@ func DefineCommonFlags(flags *pflag.FlagSet) {
...
+	flags.Duration(flagGrpcKeepaliveTimeout, defaultGRPCKeepaliveTimeout,
+		"the max time a grpc conn can keep idel before killed, must keep same value with TiKV and PD")
```
kennytm, line 161:
```suggestion
		"the max time a gRPC connection can keep idle before killed, must keep same value with TiKV and PD")
```

overvenus (commented):
LGTM

kennytm:
/lgtm

Please remember to update https://github.com/pingcap/tidb/blob/03c6b5ab9efcf9abe3621d4ce9244ac56d90755b/executor/brie.go#L184 when updating TiDB's BR deps. Maybe we should expose a `DefaultConfig()` function 🤔

kennytm (approved):

ti-srebot (approved):

YuJuncen:
> /lgtm
> 
> Please remember to update https://github.com/pingcap/tidb/blob/03c6b5ab9efcf9abe3621d4ce9244ac56d90755b/executor/brie.go#L184 when updating TiDB's BR deps. Maybe we should expose a `DefaultConfig()` function 🤔

I think we'd better add this to `adjust` function now, in case of forgetting...

PR #546, opened 2020-09-30
*: allow config the concurrency of checksumming

YuJuncen:
In the online cluster, high-concurrency checksumming would make influence to the online transactions. This PR allows manually to use lower concurrency.

Add a new config `checksum-concurrency`.

### Check List 

Tests 

 - Integration test

### Release Note

 - Added a new config `checksum-concurrency`.

**To be discussed:** should we change the default value from 15(`variable.DefDistSQLScanConcurrency`) to 4(`variable.DefTableChecksumConcurrency`)?


pkg/checksum/executor.go
```diff
@@ -132,6 +142,7 @@ func buildRequest(
 		if err != nil {
 			return nil, err
 		}
+		req.NotFillCache = true
```
overvenus, line 145: Please set it in `build(Table|Index)Request`.

overvenus (commented):
Could you add a unit test to check if checksum requests are set the correct concurrency and NotFillCache?

pkg/checksum/executor.go
YuJuncen: Oops... This line should be removed since `SetChecksumRequest` did this for us.

pkg/task/backup.go
```diff
@@ -306,7 +306,7 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
 		updateCh = g.StartProgress(
 			ctx, "Checksum", int64(backupSchemas.Len()), !cfg.LogProgress)
 		backupSchemas.Start(
-			ctx, mgr.GetTiKV(), backupTS, uint(backupSchemasConcurrency), updateCh)
+			ctx, mgr.GetTiKV(), backupTS, uint(backupSchemasConcurrency), cfg.ChecksumConcurrency, updateCh)
```
3pointer, line 309: If in SQL backup. `cfg.ChecksumConcurrency` would be 0, `adjustBackupConfig` should handle this.

pkg/task/restore.go
```diff
@@ -264,7 +264,7 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 	// Checksum
 	if cfg.Checksum {
 		finish = client.GoValidateChecksum(
-			ctx, afterRestoreStream, mgr.GetTiKV().GetClient(), errCh, updateCh)
+			ctx, afterRestoreStream, mgr.GetTiKV().GetClient(), errCh, updateCh, cfg.ChecksumConcurrency)
```
3pointer, line 267: ditto

YuJuncen: Done at `Config.adjust`

pkg/checksum/executor.go
```diff
@@ -252,6 +266,17 @@ func (exec *Executor) Len() int {
 	return len(exec.reqs)
 }
 
+// Each executes the function to each requests in the executor.
+func (exec *Executor) Each(f func(*kv.Request) error) error {
```
overvenus, line 270: Could you remove it as it's only used in tests?

overvenus (commented):
Rest LGTM

pkg/checksum/executor.go
YuJuncen: The test cases are in other packages, hence cannot access private fields that we want to assert...

Keep this function might be necessary, otherwise, we must export the `reqs` field.

overvenus: I see, I prefer to export Reqs directly. `Each` is somewhat unnecessary.

overvenus (approved):

pkg/checksum/executor.go
YuJuncen: I prefer `Each` in some way because it keeps the field being immutable outside. Even definitely it's more complex than exporting this field...

3pointer (commented):
LGTM

ti-srebot (approved):

PR #547, opened 2020-10-02
Refactor uploaderWriter to have a context

tirsen:
If we instead have a context as a struct field we can fully implement the io.Writer and io.Closer interfaces correctly which makes things much more flexible. This makes it match the s3ObjectReader which uses the same technique to implement io.Reader.

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -35,14 +35,6 @@ type Uploader interface {
 	CompleteUpload(ctx context.Context) error
 }
 
-// Writer is like io.Writer but with Context, create a new writer on top of Uploader with NewUploaderWriter.
-type Writer interface {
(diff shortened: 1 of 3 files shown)


tirsen:
You can see here how much better this is to use :-) https://github.com/pingcap/dumpling/pull/161

kennytm:
We're passing contexts around instead of storing inside the uploader because https://pkg.go.dev/context

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx: 

(See also #476)

tirsen:
I see. Shame really, it's so much prettier :-)

kennytm:
As we're going to keep following Go's recommendation, I'm going to close this PR. Thanks for the contribution anyway :).

PR #553, opened 2020-10-12
checksum: fix checksum gc life time excceed issue

3pointer:
br encountered the `GC life time is shorter than transaction duration` error. because of this [check](https://github.com/pingcap/tidb/blob/180c02127105bed73712050594da6ead4d70a85f/store/tikv/kv.go#L186-L190) in restore checksum.
Add `StartServiceSafePointKeeper` for restore procedure just like backup.

### Check List 

Tests 

 - No code

Related changes

### Release Note

 - Fix the issue that restore failed when restore checksum task time more than gc life time.

### DIFF
--- a/pkg/task/restore.go
+++ b/pkg/task/restore.go
@@ -147,12 +147,24 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 		return errors.Annotate(berrors.ErrRestoreInvalidBackup, "contain tables but no databases")
 	}
 
+	restoreTS, err := client.GetTS(ctx)
+	if err != nil {
+		return err
+	}
+
+	sp := utils.BRServiceSafePoint{
+		BackupTS: restoreTS,
+		TTL:      utils.DefaultBRGCSafePointTTL,
+		ID:       utils.MakeSafePointID(),
+	}
+	// restore checksum will check safe point with its start ts, see details at
+	// https://github.com/pingcap/tidb/blob/180c02127105bed73712050594da6ead4d70a85f/store/tikv/kv.go#L186-L190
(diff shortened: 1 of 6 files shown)


pkg/task/restore.go
```diff
@@ -147,12 +147,24 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
...
 	}
 
+	restoreTS, err := client.GetTS(ctx)
+	if err != nil {
+		return err
+	}
+
+	sp := utils.BRServiceSafePoint{
+		BackupTS: restoreTS,
+		TTL:      utils.DefaultBRGCSafePointTTL,
```
kennytm, line 157: perhaps we should also make this configurable like backup?

kennytm (commented):
do we need to touch restore_raw?

pkg/task/restore.go
3pointer: 1. restore_raw doesn't have checksum logic. so we don't need to touch it.
2. I would like to make TTL a constant also in backup. because we know the default minimum gc life time is 10 minutes. we can set to 2 or 3 minutes, and the user seems never change these config?🤔️

YuJuncen:
LGTM

pkg/task/restore.go
kennytm: > I would like to make TTL a constant also in backup. 

Hmm perhaps file an issue or open a thread in SIG-Migrate so we can discuss it separately?

kennytm:
LGTM

ti-srebot (approved):

pkg/task/restore.go
3pointer: Ok, I will open a thread in SIG-Migrate. How about merge it first so that won't block other PRs?

PR #555, opened 2020-10-15
pd: disable location replacement

YuJuncen:
The PD will generate some operators for better performance, but they aren't needed by BR and sometimes will slow down the restore process.
*(This PR has disabled the config `enable-location-replacement` to prevent PD from generating  `move-to-better-location` operators.

### Check List 

Tests 

 - Integration test

### Release Note

 - Disable `enable-location-replacement` during restore.


pkg/pdutil/pd.go
```diff
@@ -68,11 +68,12 @@ var (
...
-		"max-merge-region-size": 20,
-		"leader-schedule-limit": 4,
-		"region-schedule-limit": 2048,
-		"max-snapshot-count":    3,
+		"max-merge-region-keys":       200000,
+		"max-merge-region-size":       20,
+		"leader-schedule-limit":       4,
+		"region-schedule-limit":       2048,
+		"max-snapshot-count":          3,
+		"enable-location-replacement": false,
```
kennytm, line 76: the default is `false` *and* we'll also set it to `false` in RemoveSchedulers? Sounds like we don't need to include this here 🤔

YuJuncen: That was a mistake, default should be `"true"`.

YuJuncen:
```
[2020-10-15T07:15:12.685Z] 4895 is running
[2020-10-15T07:15:17.983Z] tests/br_other/run.sh: line 103:  4895 Terminated              GO_FAILPOINTS="github.com/pingcap/br/pkg/utils/determined-pprof-port=return($PPROF_PORT)" run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB/lock" --remove-schedulers --ratelimit 1 --ratelimit-unit 1 --concurrency 4 2>&1 > $TEST_DIR/background-br.log
```

It is killed after 5 secs... How strange...

YuJuncen:
`[2020-10-15T08:48:04.196Z] failed to restore, before: 1000; after: 600`...

/run-all-tests

tests/br_other/run.sh
```diff
@@ -64,22 +64,22 @@ fi
 echo "backup start to test lock file"
 PPROF_PORT=6080
 GO_FAILPOINTS="github.com/pingcap/br/pkg/utils/determined-pprof-port=return($PPROF_PORT)" \
-run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB/lock" --remove-schedulers --ratelimit 1 --ratelimit-unit 1 --concurrency 4 2>&1 >/dev/null &
+run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB/lock" --remove-schedulers --ratelimit 1 --ratelimit-unit 1 --concurrency 4 2>&1 >$TEST_DIR/background-br.log &
```
overvenus, line 67:
```suggestion
run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB/lock" --remove-schedulers --ratelimit 1 --ratelimit-unit 1 --concurrency 4 2>&1 >$TEST_DIR/br_stdout.log &
```

YuJuncen: That log is just for debugging the test case, I'm thinking that keeps it `/dev/null` would be fine. (Anyway, we must connect to the Jenkins machine to view the file...)

YuJuncen:
`[2020-10-15T12:22:14.091Z] failed to restore, before: 1000; after: 600` again...🤔
/run-all-tests

pkg/pdutil/pd.go
```diff
@@ -441,6 +443,12 @@ func restoreSchedulers(ctx context.Context, pd *PdController, clusterCfg cluster
 	if err := pd.UpdatePDScheduleConfig(ctx, scheduleLimitCfg); err != nil {
 		return errors.Annotate(err, "fail to update PD schedule config")
 	}
+	if locationPlacement, ok := clusterCfg.scheduleCfg["enable-location-replacement"]; ok {
```
overvenus, line 446: How about adding `"enable-location-replacement"` to `pdScheduleLimitCfg`?

overvenus (commented):
Rest LGTM

pkg/pdutil/pd.go
YuJuncen: The map `pdScheduleLimitCfg` is also used in `removeSchedulers`, and they also have some generic traits(e.g. both of their key are number), so I'm fearing adding `enable-location-replacement` would make extra overhead.

See here:

https://github.com/pingcap/br/blob/6734867f0fd1d6f12dc38c51fdd3ae11749bed01/pkg/pdutil/pd.go#L511-L522

overvenus:
LGTM

kennytm (commented):
lgtm

ti-srebot (approved):

kennytm:
```
[2020-10-21T13:14:10.396Z] TEST: [br_other] failed because paused scheduler has changed
```

likely legit.

tests/br_other/run.sh
```diff
@@ -136,22 +137,24 @@ for key in $(echo $default_pd_values | jq 'keys[]'); do
...
 done
 
-pd_settings=5
 
 # check is there still exists scheduler in pause.
-pause_schedulers=$(curl http://$PD_ADDR/pd/api/v1/schedulers?status="paused" | grep "scheduler" | wc -l)
-# There shouldn't be any paused schedulers since BR gracfully shutdown.
-if [ "$pause_schedulers" -ne "0" ];then
+pause_schedulers_after=$(curl http://$PD_ADDR/pd/api/v1/schedulers?status="paused" | grep "scheduler" | wc -l)
+if [ "$pause_schedulers" -ne "$pause_schedulers_after" ];then
```
YuJuncen, line 143: #559 got merged, we shouldn't revert its change...

```suggestion
pause_schedulers=$(curl http://$PD_ADDR/pd/api/v1/schedulers?status="paused" | grep "scheduler" | wc -l)
 # There shouldn't be any paused schedulers since BR gracfully shutdown.
 if [ "$pause_schedulers" -ne "0" ];then
```

YuJuncen:
`ssh: connect to host github.com port 22: Operation timed out`

/run-integration-tests

pkg/pdutil/pd.go
```diff
@@ -73,11 +73,12 @@ var (
...
 	// DefaultPDCfg find by https://github.com/tikv/pd/blob/master/conf/config.toml.
 	DefaultPDCfg = map[string]interface{}{
-		"max-merge-region-keys": 200000,
-		"max-merge-region-size": 20,
-		"leader-schedule-limit": 4,
-		"region-schedule-limit": 2048,
-		"max-snapshot-count":    3,
+		"max-merge-region-keys":       200000,
+		"max-merge-region-size":       20,
+		"leader-schedule-limit":       4,
```
3pointer, line 78: According to @Yisaer, these two configs "leader-schedule-limit" and "region-schedule-limit" don't help the schedule speed. should we remove it in this PR?

YuJuncen: I think we could leave a TODO here and try to merge this firstly... I'm afraid that if any error occurs during editing, we cannot merge this in the current unstable CI environment...

overvenus:
Data loss found in test https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/3442/pipeline#step-1354-log-318

pkg/pdutil/pd.go
Yisaer: currently `leader-schedule-limit` and `region-schedule-limit` is not supported in temporary configuration https://github.com/tikv/pd/pull/3088

YuJuncen:
/run-integration-test

fixed?

tests/br_tiflash/run.sh
```diff
@@ -52,8 +52,9 @@ run_br backup full -s "local://$TEST_DIR/$DB" --pd $PD_ADDR
 run_sql "DROP DATABASE $DB"
 run_br restore full -s "local://$TEST_DIR/$DB" --pd $PD_ADDR
 
-# FIXME after stopping schedulers, tiflash takes many time to sync with TiKV(even 30s isn't enough).
-AFTER_BR_COUNT=`run_sql "SELECT count(*) /*+ READ_FROM_STORAGE(TIKV[$DB.kv]) */ FROM $DB.kv;" | sed -n "s/[^0-9]//g;/^[0-9]*$/p" | tail -n1`
+# wating for TiFlash sync
+sleep 80
```
kennytm: 🤔

PR #559, opened 2020-10-16
*: when cleaning up, use isolated context

### LINKED ISSUE #557: BR won't clean up the environment when exit by SIGTERM

1. What did you do?
- start BR (restore or backcup with `--remove-schedulers`)
- waiting for the progress bar present, then press <kbd>ctrl</kbd> + <kbd>c</kbd>

The cluster config changed by BR should be undone, since SIGTERM allows us to gracefully stop.

3. What did you see instead?
The cluster has stuck in the config that BR has set. (For current master, PD schedulers could be reset due to #551 )
image

v4.0.7

#### Note:

We listen to signals here:

https://github.com/pingcap/br/blob/d2d5bbaf29bdbc2b1d9453ec65096e04c52b529e/main.go#L34-L39

Canceling the context could make other goroutines eventually exit and clean up, but we leave no time for them.

<del>Add a `time.Sleep(30 * time.Second)`</del> remove those `os.Exit`s could help. But there are still some problems:

https://github.com/pingcap/br/blob/d2d5bbaf29bdbc2b1d9453ec65096e04c52b529e/pkg/task/backup.go#L222-L227

We use the global context to do the cleanup tasks, which will always fail if the outer context is canceled. We should change it to a new context with a timeout, the timeout could be the same as the sleep time before stopping.

YuJuncen:
fix #557
This PR also added a guard for the `glue.Progress` implement, to prevent the `send on close channel` panics.

see #557 

### Check List 

Tests 

 - Integration test

### Release Note

 - Fixed a bug that caused BR won't do cleanup tasks when receiving SIGTERM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,12 +121,12 @@ lint: tools
 	CGO_ENABLED=0 tools/bin/revive -formatter friendly -config revive.toml $$($(PACKAGES))
 
 tidy: prepare
-	# tidy isn't a read-only task for go.mod, run FINISH_MOD always, 
-	# so our go.mod1 won't stick in old state
-	trap "$(FINISH_MOD)" EXIT
 	@echo "go mod tidy"
 	GO111MODULE=on go mod tidy
-	git diff --quiet go.mod go.sum
+	# tidy isn't a read-only task for go.mod, run FINISH_MOD always, 
+	# so our go.mod1 won't stick in old state
+	git diff --quiet go.mod go.sum || ("$(FINISH_MOD)" && exit 1)
+	$(FINISH_MOD)
 
 failpoint-enable: tools
 	tools/bin/failpoint-ctl enable
--- a/Makefile
+++ b/Makefile
@@ -121,10 +121,12 @@ lint: tools
 	CGO_ENABLED=0 tools/bin/revive -formatter friendly -config revive.toml $$($(PACKAGES))
 
 tidy: prepare
+	# tidy isn't a read-only task for go.mod, run FINISH_MOD always, 
+	# so our go.mod1 won't stick in old state
+	trap "$(FINISH_MOD)" EXIT
 	@echo "go mod tidy"
 	GO111MODULE=on go mod tidy
 	git diff --quiet go.mod go.sum
-	$(FINISH_MOD)
 
 failpoint-enable: tools
 	tools/bin/failpoint-ctl enable
--- a/main.go
+++ b/main.go
@@ -31,6 +31,11 @@ func main() {
 		fmt.Printf("\nGot signal [%v] to exit.\n", sig)
 		log.Warn("received signal to exit", zap.Stringer("signal", sig))
 		cancel()
+		fmt.Println("gracefully shuting down, press ^C again to force exit")
+		<-sc
+		// Even user use SIGTERM to exit, there isn't any checkpoint for resuming,
+		// hence returning fail exit code.
+		os.Exit(1)
 	}()
 
 	rootCmd := &cobra.Command{
--- a/main.go
+++ b/main.go
@@ -31,7 +31,7 @@ func main() {
 		fmt.Printf("\nGot signal [%v] to exit.\n", sig)
 		log.Warn("received signal to exit", zap.Stringer("signal", sig))
 		cancel()
-		fmt.Println("gracefully shuting down, press ^C again to force exit")
+		fmt.Fprintln(os.Stderr, "gracefully shuting down, press ^C again to force exit")
 		<-sc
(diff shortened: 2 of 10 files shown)


main.go
```diff
@@ -30,14 +30,7 @@ func main() {
...
 		log.Warn("received signal to exit", zap.Stringer("signal", sig))
-		switch sig {
-		case syscall.SIGTERM:
-			cancel()
-			os.Exit(0)
-		default:
-			cancel()
-			os.Exit(1)
-		}
+		cancel()
```
overvenus, line 33: What about trying to graceful shutdown at the first time BR receives signal, and os.Exit() when it receives signal twice? So we can avoid hard code 30 seconds timeout.

YuJuncen: Good idea!

YuJuncen: Done, PTAL~

Makefile
```diff
@@ -121,10 +121,12 @@ lint: tools
 	CGO_ENABLED=0 tools/bin/revive -formatter friendly -config revive.toml $$($(PACKAGES))
 
 tidy: prepare
+	# tidy isn't a read-only task for go.mod, run FINISH_MOD always, 
+	# so our go.mod1 won't stick in old state
+	trap "$(FINISH_MOD)" EXIT
```
kennytm, line 126: this `trap` likely won't work, since every line in the makefile starts a new shell.

YuJuncen: Oops... So it's a little hard to gracefully run it...

pkg/gluetikv/glue.go
```diff
@@ -5,10 +5,13 @@ package gluetikv
 import (
 	"context"
 
+	"sync/atomic"
```
overvenus, line 8:
```suggestion
	"context"
	"sync/atomic"
```

pkg/task/restore.go
```diff
@@ -367,6 +367,10 @@ func restorePreWork(ctx context.Context, client *restore.Client, mgr *conn.Mgr)
 func restorePostWork(
 	ctx context.Context, client *restore.Client, restoreSchedulers utils.UndoFunc,
 ) {
+	if ctx.Err() != nil {
+		log.Warn("context canceled, doing clean work with background context")
```
overvenus, line 371:
```suggestion
		log.Warn("context canceled, try shutdown")
```

pkg/task/backup_raw.go
```diff
@@ -147,6 +147,10 @@ func RunBackupRaw(c context.Context, g glue.Glue, cmdName string, cfg *RawKvConf
 	if cfg.RemoveSchedulers {
 		restore, e := mgr.RemoveSchedulers(ctx)
 		defer func() {
+			if ctx.Err() != nil {
+				log.Warn("context canceled, doing clean work with background context")
```
overvenus, line 151:
```suggestion
				log.Warn("context canceled, try shutdown")
```

overvenus (approved):
Rest LGTM

main.go
```diff
@@ -30,14 +30,12 @@ func main() {
...
+		fmt.Println("gracefully shuting down, press ^C again to force exit")
```
kennytm, line 34: print to stderr please.

```suggestion
		fmt.Fprintln(os.Stderr, "gracefully shuting down, press ^C again to force exit")
```

pkg/gluetikv/glue.go
```diff
@@ -57,15 +60,28 @@ func (Glue) Record(name string, val uint64) {
...
+	if atomic.LoadInt32(&p.closed) != 0 {
+		log.Warn("proposing a closed progress")
+		return
+	}
+	// there might be buggy if the thread is yielded here.
+	// however, there should not be gosched, at most time.
+	// so send here probably is safe, even not totally safe.
+	// since adding an extra lock should be costly, we just be optimistic.
+	// (Maybe a spin lock here would be better?)
 	p.ch <- struct{}{}
```
kennytm, line 78: it is possible to just make this entire progress structure an atomic integer instead of a channel, like https://github.com/pingcap/tidb/blob/c7651f02be36b5faf95a204d20f8859eb9c2080a/executor/brie.go#L49-L73? tbh i don't see any reason why we use a channel in the first place, since we're using a 1-second timer to update the progress bar in any case.

YuJuncen: well there would be more change than this pr itself if we refactor the progess struct here. how about leaving a TODO here and leave it another PR?

YuJuncen: (But I could have a try this PR, anyway)

kennytm (commented):
lgtm

ti-srebot (approved):

PR #561, opened 2020-10-21
progress: removed the progress struct

YuJuncen:
remove the progress struct.
This struct was error-prone(It has multi send-part but only close once, hence may leading to`send on closed channel`), hence we embedded its function to an atomic integer in `ProgressPrinter`. 

Remove the struct `glue.progress`, implement `glue.Progress` for `ProgressPrinter` directly.
This may cause the progress value won't immediately be added when calling `Inc`. But we update the progress status by polling the progress every second, it would be ok to not react immediately.

### Check List 

Tests 

 - Unit test

### Release Note

 - No release note.


pkg/utils/progress.go
```diff
@@ -6,20 +6,32 @@ import (
...
+func (a *atomicAdder) get() int64 {
+	return atomic.LoadInt64((*int64)(a))
+}
+
 // ProgressPrinter prints a progress bar.
 type ProgressPrinter struct {
 	name        string
 	total       int64
 	redirectLog bool
+	progress    atomicAdder
```
kennytm, line 32: i think using `int64` directly here is fine? using a new type here does not provide any real protection as `progress += 1` is still accepted even using the `atomicAdder` type.

pkg/utils/progress.go
```diff
@@ -64,7 +85,7 @@ func (pp *ProgressPrinter) goPrintProgress(
 	}
 	if testWriter != nil {
 		bar.SetWriter(testWriter)
-		bar.SetRefreshRate(10 * time.Millisecond)
+		bar.SetRefreshRate(2 * time.Second)
```
kennytm, line 88: why changing this to 2s

YuJuncen: For testing, if keeping it 10ms, "progress = 0%" would be filled in the testing writer, hence we cannot see any progress.

pkg/utils/progress.go
YuJuncen: `progress.add(1)` will be shorter than `atomic.AddInt64(&progress, 1)`, and more semantic...?

If more protection is needed, I can wrap it like `type atomicAdder struct { inner int64 }`.

kennytm: it's shorter in the call, but you need to define an entire type and function which is only used once. i don't think it is more semantic here.

YuJuncen: well, i have removed it.

YuJuncen:
`Operation: [get]  for kind: [Pod]  with name: [br-test-go1130-1hk2t-xkcn3]  in namespace: [jenkins-ci]  failed.` 🤔

/run-integration-test

YuJuncen:
/build

....

YuJuncen:
/run-integration-test

default not found? 🤔

overvenus:
> /run-integration-test
> 
> default not found? 🤔

@YuJuncen   Do you have a link to the error? It may be a bug and indicates data loss.

kennytm (commented):
LGTM

3pointer (commented):
LGTM

ti-srebot (approved):

PR #566, opened 2020-10-23
add uploader compress writer

lichunzhu:
https://github.com/pingcap/dumpling/issues/7 

Add uploader compress writer

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### Release Note

 - None


pkg/storage/writer.go
```diff
@@ -54,6 +55,65 @@ func (u *uploaderWriter) Close(ctx context.Context) error {
...
 }
 
+type uploaderCompressWriter struct {
+	*uploaderWriter
+	bytesWritten int
+	cap          int
+	gzipWriter   *gzip.Writer
+}
+
+func (u *uploaderCompressWriter) Write(ctx context.Context, p []byte) (int, error) {
```
kennytm, line 65: 🤔 it is possible to make this function and `(*uploaderWriter).Write` share code? they look too similar.

lichunzhu: fixed

pkg/storage/writer.go
```diff
@@ -2,11 +2,71 @@ package storage
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 )
 
+type interceptBuffer struct {
+	*bytes.Buffer
+	compressWriter *gzip.Writer
```
kennytm, line 11: make this an `interface { io.Writer; io.Closer; Flush() error }`, we may support other compression algorithms.

pkg/storage/writer.go
```diff
@@ -2,11 +2,71 @@ package storage
...
+}
+
+func (b *interceptBuffer) Close() error {
+	if b.compressWriter == nil {
+		return nil
+	}
+	return b.compressWriter.Close()
+}
+
+func newInterceptBuffer(chunkSize int, compress bool) *interceptBuffer {
```
kennytm, line 58: use an enum instead of a bool

kennytm (commented):
rest LGTM

kennytm (commented):
LGTM but I wonder why do we want to support zlib :shrug: ("other compression algorithms" mostly means lz4 and zstd)

lichunzhu:
> LGTM but I wonder why do we want to support zlib 🤷 ("other compression algorithms" mostly means lz4 and zstd)

Just because it's in compress package like `gzip`. Should I delete this?

kennytm:
> Just because it's in compress package like `gzip`. Should I delete this?

gzip and zlib are equivalent except the container format, so yeah there's no point adding this. we're only interested in:

* gzip (legacy fast & low compression format)
* xz (somewhat-legacy slow & high compression format)
* lz4 (faster than gzip)
* zstd (faster than xz)

kennytm:
@lichunzhu i mean we can keep `simpleCompressBuffer` just not add the `zlib` support 👀

lichunzhu:
OK

kennytm:
/run-integration-test

https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/3609/pipeline

Hmm can't read the log at all.

```
Error: failed to validate checksum: [BR:Restore:ErrRestoreChecksumMismatch]restore checksum mismatch
```

kennytm:
(retrigger CLA)

3pointer (approved):
LGTM

PR #567, opened 2020-10-24
restore: set configed keepalive for restore

YuJuncen:
#545 added the config value of keepalive, but it has forgotten the restore part(!).

Apply the config to restoring.

### Check List 

Tests 

 - Unit test
 - Integration test
(If the old tests passed, then the default config should be applied.)

### Release Note

 - Fix a bug that causes the keepalive configs don't apply to restore.

### DIFF
--- a/pkg/task/restore.go
+++ b/pkg/task/restore.go
@@ -102,7 +102,9 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 	}
 	defer mgr.Close()
 
-	client, err := restore.NewRestoreClient(g, mgr.GetPDClient(), mgr.GetTiKV(), mgr.GetTLSConfig())
+	keepaliveCfg := GetKeepalive(&cfg.Config)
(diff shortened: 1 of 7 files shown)


pkg/task/restore.go
```diff
@@ -102,7 +102,9 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 	}
 	defer mgr.Close()
 
-	client, err := restore.NewRestoreClient(g, mgr.GetPDClient(), mgr.GetTiKV(), mgr.GetTLSConfig())
+	keepaliveCfg := GetKeepalive(&cfg.Config)
+	keepaliveCfg.PermitWithoutStream = true
```
3pointer, line 106: How about put this line in GetKeepalive function?

YuJuncen: This function is also used in the backup site, which doesn't needs `PermitWithoutStream`.

3pointer (commented):
LGTM

kennytm (approved):
LGTM

PR #568, opened 2020-10-26
cdclog: Fix serval bugs about restore cdclog feature.

3pointer:
Fix serval bugs about restore cdclog feature.
1. Fix the issue that when restore binary/varbinary field will lost bytes.
2. Fix the issue that wrong events filtered by ts range filter.
3. Fix panic by add the length check for cdclog data.
4. Change the hidden parameter to the experimental feature.
5. Add more log for debug.

### Check List 

Tests 

 - Integration test

Related changes

### Release Note

 - No release note


pkg/cdclog/puller.go
```diff
@@ -61,10 +61,12 @@ func NewEventPuller(
 		if err != nil {
 			return nil, errors.Trace(err)
 		}
-		ddlFileIndex++
```
glorv, line 64: what if the len(ddlFiles[0]) == 0 and len(ddlFiles[1]) > 0. Can this happened?

3pointer: This shouldn't happen because we sort the ddl files by its' event timestamp.

glorv:
LGTM

kennytm (approved):
LGTM

ti-srebot (approved):

overvenus:
How do you find these bugs?

PR #575, opened 2020-11-01
go.mod,tests: update tidb deps and add list partition test case

kennytm:

BR was using a TiDB deps before pingcap/tidb#20279. If the cluster actually has a list partitioned table, BR will crash when loading the info schema since a list partition is unexpected.

Upgrade the TiDB dependency to latest master.

Enhanced `br_table_partition` to include HASH and LIST partitions.

(No need to cherry-pick to release-4.0. List partition does not exist on 4.0.)

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### Release Note

 - BR 5.0 now supports list partitioned tables.


3pointer (commented):
LGTM

overvenus (approved):
LGTM

ti-srebot (approved):

PR #578, opened 2020-11-02
restore: merge small ranges

overvenus:

Speed up restore by merging small range to reduce split and scatter.

Merge adjacent ranges that total_bytes is less 96MB or total_kv is less than 960000.

### Check List 

Tests 

 - Integration test

Side effects

 - ~Breaking backward compatibility~

### Release Note

 - Speed up restore by merging small backup files.


tests/br_debug_meta/run.sh
```diff
@@ -15,24 +15,23 @@
...
-) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;"
-
-run_sql "INSERT INTO $DB.$TABLE VALUES (\"a\", \"b\");"
-run_sql "INSERT INTO $DB.$TABLE VALUES (\"aa\", \"b\");"
+for i in $(seq 10); do
+    regioncount=$(run_sql "SHOW TABLE $DB.$TABLE REGIONS;" | grep "REGION_ID" | wc -l)
+    [ $regioncount -ge 2 ] && break
+    sleep 3
+done
+run_sql "SHOW TABLE $DB.$TABLE REGIONS;" | grep "REGION_ID"
```
kennytm, line 28: 🤔 

```sql
select count(*) from information_schema.tikv_region_status
where db_name = '$DB' and table_name = '$TABLE';
```

also you could use `split table $DB.$TABLE by (some_pk_value)` to immediately split the table into two regions, no need to do this loop.

overvenus: Could you elaborate on how to split table?  Tried `run_sql "split table $DB.$TABLE by ('user6282602628620641459', 'user6282603728132269670', 'user6282604827643897881', 'user6282610325202038936')"`, but it does not work out.

kennytm: @overvenus try `split table $DB.$TABLE by ('user6282602628620641459'), ('user6282603728132269670'), /*etc*/`;

pkg/task/restore.go
```diff
@@ -39,16 +44,28 @@ type RestoreConfig struct {
 
 	Online   bool `json:"online" toml:"online"`
 	NoSchema bool `json:"no-schema" toml:"no-schema"`
+
+	// MergeSmallRegionSizeBytes is the threshold of merging small regions (Default 96MB, region split size).
+	// MergeSmallRegionKeyCount is the threshold of merging smalle regions (Default 960_000, region split key count).
+	// See https://github.com/tikv/tikv/blob/v4.0.8/components/raftstore/src/coprocessor/config.rs#L35-L38
+	MergeSmallRegionSizeBytes uint64 `json:"merge-region-size-bytes" toml:"merge-region-size-bytes"`
+	MergeSmallRegionKeyCount  uint64 `json:"merge-region-key-count" toml:"merge-region-key-count"`
```
kennytm, line 52: include these into `adjustRestoreConfig`, otherwise before the next update, BR in SQL will use 0 for both of these (which is probably fine 🤔).

tests/br_debug_meta/run.sh
```diff
@@ -15,24 +15,25 @@
...
-
-run_sql "INSERT INTO $DB.$TABLE VALUES (\"a\", \"b\");"
-run_sql "INSERT INTO $DB.$TABLE VALUES (\"aa\", \"b\");"
+table_region_sql="SELECT COUNT(*) FROM information_schema.tikv_region_status WHERE db_name = '$DB' AND table_name = '$TABLE';"
+for i in $(seq 10); do
+    regioncount=$(run_sql "$table_region_sql" | awk '/COUNT/{print $2}')
+    [ $regioncount -ge 5 ] && break
+    sleep 3
+done
+run_sql "$table_region_sql"
```
kennytm, line 30: `SPLIT TABLE` should be effective immediately, so no need to check region count again 🤔 (you could check the `TOTAL_SPLIT_REGION` field returned by `SPLIT TABLE` if you need to be absolutely sure)

kennytm (commented):
rest LGTM

tests/br_debug_meta/run.sh
overvenus: It does split, however, data is not distributed to multiple regions, so I think just leave the loop and let TiKV splits.

kennytm: ok.

overvenus:
/run-integration-test

Fails with #641 
```
start ts:421372453438095363 is greater than resolved ts:0: [BR:Restore:ErrRestoreResolvedTsConstrain]resolved ts constrain violation
```

kennytm:
/run-integration-test

CI hiccup

```
[2020-12-11T08:55:55.828Z] java.io.IOException: Tried to load head FlowNodes for execution Owner[br_ghpr_unit_and_integration_test/4180:br_ghpr_unit_and_integration_test #4180] but FlowNode was not found in storage for head id:FlowNodeId 1:29
```

pkg/restore/merge.go
```diff
@@ -0,0 +1,144 @@
...
+			return true
+		}
+		if leftBytes+rightBytes > splitSizeBytes {
+			return false
+		}
+		if leftKeys+rightKeys > splitKeyCount {
+			return false
+		}
+		// Do not merge ranges in different tables.
+		if tablecodec.DecodeTableID(kv.Key(left.StartKey)) != tablecodec.DecodeTableID(kv.Key(right.StartKey)) {
```
3pointer, line 111: Can we merge index range and row key range with same table id here?

overvenus: It's ok as long as they belong to the same table.

pkg/task/restore.go
```diff
@@ -29,41 +29,99 @@ const (
...
+	RestoreCommonConfig
 
-	Online   bool `json:"online" toml:"online"`
 	NoSchema bool `json:"no-schema" toml:"no-schema"`
 }
 
-// DefineRestoreFlags defines common flags for the restore command.
+// DefineRestoreFlags defines common flags for the restore tidb command.
 func DefineRestoreFlags(flags *pflag.FlagSet) {
 	// TODO remove experimental tag if it's stable
```
kennytm, line 105:
```suggestion
```

kennytm (commented):
rest LGTM

pkg/restore/merge.go
3pointer: If we merge record key and index key into one file. we need to find the correct rewrite rule. https://github.com/pingcap/br/blob/c344d1edf20cf17d353b9d8b789919fd1172039f/pkg/restore/import.go#L212
but default rewrite rule starts with record prefix 
https://github.com/pingcap/br/blob/c344d1edf20cf17d353b9d8b789919fd1172039f/pkg/restore/util.go#L73
and with this rule sst_importer cannot rewrite index key correctly.

lichunzhu (approved):
LGTM

pkg/restore/merge.go
overvenus: You are right, we need to separate ranges by index ID too.

pkg/restore/merge.go
```diff
@@ -0,0 +1,164 @@
...
+				// Do not merge if they are not the same index.
+				return false
+			}
+			// They are in the same index, merge left and right.
+			return true
+		default:
+			// err1 != nil && err2 != nil
+			// They are record keys, merge left and right.
+			return true
+		}
```
kennytm, line 134:
```suggestion
		// If both of them are index keys, ...
		if err1 == nil && err2 == nil {
			// Merge left and right if they are in the same index.
			return indexID1 == indexID2
		}
		// Otherwise, merge if they are both record keys
		return err1 != nil && err2 != nil
```

we have also already checked the table ID in line 111 so `tableID1 != tableID2` is impossible.

3pointer:
LGTM

ti-srebot (approved):

PR #592, opened 2020-11-10
pdutil: use temp pause config

3pointer:
After https://github.com/tikv/pd/pull/3088/ pd support set temporary config with ttlSecond. so BR can improve the behavior of pd config setting.

1. refactor pdutil: merge all configs together
2. add pauseConfig for 4.0.8. 

### Check List 

Tests 

 - No code

Code changes

Related changes

### Release Note
- No release note.
 -


pkg/pdutil/pd.go
```diff
@@ -153,14 +160,18 @@ func NewPdController(
...
 		if failure == nil {
 			break
 		}
 	}
 	if failure != nil {
 		return nil, errors.Annotatef(failure, "pd address (%s) not available, please check network", pdAddrs)
 	}
+	version, err := semver.NewVersion(string(versionBytes))
+	if err != nil {
+		return nil, errors.Annotatef(err, "transform pd version failed", string(versionBytes))
```
overvenus, line 173: Could you choose an error in pkg/errors?

overvenus (commented):
Rest LGTM

pkg/pdutil/pd.go
```diff
@@ -170,7 +170,12 @@ func NewPdController(
 		return nil, errors.Annotatef(berrors.ErrPDUpdateFailed, "pd address (%s) not available, please check network", pdAddrs)
 	}
 
-	version := semver.New(string(versionBytes))
+	version, err := semver.NewVersion(string(versionBytes))
+	if err != nil {
+		log.Warn("fail back to old version",
+			zap.Binary("version", versionBytes), zap.Error(err))
+		version = pauseConfigVersion
```
kennytm, line 177: looks strange that `pauseConfigVersion` have to become an exclusive bound because of this line. 

and the fallback version should be v0.0.0?

pkg/pdutil/pd.go
```diff
@@ -61,15 +68,14 @@ var (
...
-		"max-merge-region-size",
-	}
-	pdScheduleLimitCfg = []string{
-		"leader-schedule-limit",
-		"region-schedule-limit",
-		"max-snapshot-count",
+	expectPDCfg = map[string]interface{}{
+		"max-merge-region-keys": 0,
+		"max-merge-region-size": 0,
+		// TODO remove this schedule-limits, see https://github.com/pingcap/br/pull/555#discussion_r509855972
```
kennytm, line 74: as the rationale keeping the two `-limit`s in #555 was

> I think we could leave a TODO here and try to merge this firstly... I'm afraid that if any error occurs during editing, we cannot merge this in the current unstable CI environment...

do we still need to keep them here?

pkg/pdutil/pd.go
```diff
@@ -153,15 +161,21 @@ func NewPdController(
...
 	}
 	if failure != nil {
-		return nil, errors.Annotatef(failure, "pd address (%s) not available, please check network", pdAddrs)
+		return nil, errors.Annotatef(berrors.ErrPDUpdateFailed, "pd address (%s) not available, please check network", pdAddrs)
 	}
 
+	version, err := semver.NewVersion(string(versionBytes))
+	if err != nil {
+		log.Warn("fail back to v0.0.0 version",
+			zap.Binary("version", versionBytes), zap.Error(err))
```
kennytm, line 176:
```suggestion
			zap.ByteString("version", versionBytes), zap.Error(err))
```

pkg/pdutil/pd.go
```diff
@@ -153,15 +161,21 @@ func NewPdController(
...
+		version = semver.New("0.0.0")
```
kennytm, line 177:
```suggestion
		version = &semver.Version{Major: 0, Minor: 0, Patch: 0}
```

or `new(semver.Version)` also produces v0.0.0 but less explicit.

pkg/pdutil/pd.go
```diff
@@ -37,6 +38,12 @@ const (
 	pauseTimeout         = 5 * time.Minute
 )
 
+var (
+	// in v4.0.8 version we can use pause configs
+	// see https://github.com/tikv/pd/pull/3088
+	pauseConfigVersion = semver.New("4.0.8")
```
kennytm, line 44:
```suggestion
	pauseConfigVersion = semver.Version{Major: 4, Minor: 0, Patch: 8}
```

pkg/pdutil/pd.go
```diff
@@ -257,19 +276,17 @@ func (p *PdController) getRegionCountWith(
 
 // PauseSchedulers remove pd scheduler temporarily.
 func (p *PdController) PauseSchedulers(ctx context.Context, schedulers []string) ([]string, error) {
-	return p.pauseSchedulersWith(ctx, schedulers, pdRequest)
+	return p.pauseSchedulersAndConfigWith(ctx, schedulers, nil, PDRequest)
 }
```
kennytm, line 280: can we remove this function? looks problematic if accidentally called instead of RemoveSchedulers  (and no one else uses PauseSchedulers anyway).

3pointer: ok, I'll remove it,and also `pauseSchedulersAndConfig`

pkg/pdutil/pd.go
3pointer: There are some misunderstanding. 
1. `leader-schedule-limit` and `region-schedule-limit` does need in previous and current version to reduce the schedule. I'll update the comment later.
2. Before https://github.com/tikv/pd/pull/3131/ `leader-schedule-limit` and `region-schedule-limit`  doesn't support ttl second.  I think we should wait PD merge 3131 then test. 

I want to make sure this PR is compatible with the old BR version and easy to support set pd ttl config. I'll file another PR to update config when 3131 merged.

pkg/pdutil/pd.go
```diff
@@ -417,42 +455,29 @@ func (p *PdController) UpdatePDScheduleConfig(
 	return errors.Annotate(berrors.ErrPDUpdateFailed, "failed to update PD schedule config")
 }
 
+func (p *PdController) doPauseConfigs(ctx context.Context, cfg map[string]interface{}, post pdHTTPRequest) error {
+	// pause this scheduler with 300 seconds
+	prefix := fmt.Sprintf("%s?ttlSecond=%d", schedulerPrefix, 300)
```
kennytm, line 460:
```suggestion
	prefix := fmt.Sprintf("%s?ttlSecond=%.0f", schedulerPrefix, pauseTimeout.Seconds())
```

kennytm (commented):
LGTM

pkg/pdutil/pd.go
```diff
@@ -400,14 +434,18 @@ func (p *PdController) GetPDScheduleConfig(
 
 // UpdatePDScheduleConfig updates PD schedule config value associated with the key.
 func (p *PdController) UpdatePDScheduleConfig(
-	ctx context.Context, cfg map[string]interface{},
+	ctx context.Context, cfg map[string]interface{}, post pdHTTPRequest, prefixs ...string,
```
glorv, line 437: Seems the parameter `post pdHTTPRequest` is only used for unit test. So I think a better way to handle this is adding a delegate method to expose the interface and add a inner implement to export the pdRequest for tests

glorv, line 437:
```suggestion
func (p *PdController) UpdatePDScheduleConfig(
    ctx context.Context, cfg map[string]interface{}, prefixs ...string,
) {
    return p.doUpdatePDScheduleConfig(ctx, cfg, pdRequest, prefixs...)
}

// all the in package call should depend on this method
func (p *PdController) doUpdatePDScheduleConfig(
    ctx context.Context, cfg map[string]interface{}, post pdHTTPRequest, prefixs ...string,
) {
...
}
```

pkg/pdutil/pd.go
```diff
@@ -86,8 +95,8 @@ var (
 // pdHTTPRequest defines the interface to send a request to pd and return the result in bytes.
 type pdHTTPRequest func(context.Context, string, string, *http.Client, string, io.Reader) ([]byte, error)
 
-// pdRequest is a func to send a HTTP to pd and return the result bytes.
-func pdRequest(
+// PDRequest is a func to send a HTTP to pd and return the result bytes.
+func PDRequest(
```
glorv, line 99: We needn't expose this function as public, see below

pkg/pdutil/pd.go
3pointer: yes, it's only used for test. @kennytm what do you think?

kennytm: @3pointer expose another method for `cmd/debug.go`, which only accepts a `ctx` and no `cfg` or `post` or `prefixes`.

pkg/pdutil/pd.go
3pointer: Fixed

pkg/pdutil/pd.go
```diff
@@ -492,43 +503,51 @@ func (p *PdController) RemoveSchedulers(ctx context.Context) (undo utils.UndoFun
...
 		}
-		// Disable region merge by setting config to 0.
-		disableMergeCfg[cfgKey] = 0
+		switch v := cfgVal.(type) {
+		case bool:
+			disablePDCfg[cfgKey] = "false"
+		case int:
+			limit := int(value.(float64))
+			disablePDCfg[cfgKey] = int(math.Min(40, float64(limit*len(stores)))) * v
+		}
```
kennytm, line 519: ok i finally got this logic, and i gotta say this is very convoluted to implement this behavior. i suggest:

1. introduce an enum to specify the expected behavior, instead of using the magic values `0`, `1` and `false`.

    ```go
    type pauseConfigExpectation uint8
    const (
        // pauseConfigSetZero sets the config to 0.
        pauseConfigSetZero pauseConfigExpectation = iota
        // pauseConfigSetMulStoresCount multiplies the existing value by
        // number of stores. The value is limited to 40, as larger value
        // may make the cluster unstable.
        pauseConfigMulStoresCount
        // pauseConfigSetFalse sets the config to "false".
        pauseConfigSetFalse
    )
    ```

2. change the type `expectPDCfg` to `map[string]pauseConfigExpectation`
3. change this switch to 

    ```go
    switch cfgVal {
    case pauseConfigSetZero:
        disablePDCfg[cfgKey] = 0
    case pauseConfigSetFalse:
        // this has to be a string instead of a boolean because <...>
        disablePDCfg[cfgKey] = "false"
    case pauseConfigSetMulStoresCount:
        ...
    }
    ```

kennytm (commented):
rest LGTM but that's a big rest.

glorv:
LGTM

ti-srebot (approved):

PR #612, opened 2020-11-24
restore: add a new interface for SplitClient to return the original region in BatchSplitRegions

glorv:
Signed-off-by: glorv <glorvs@163.com>

- Add a new interface `SplitClient.BatchSplitRegionsRaw` to return the split region after batch split.
- Add a warn log if scatters region failed, so we can determine that region distributes unevenly is due to scattering fails.

Return the split region in `BatchSplitRegion` result.

Because this change is incompatible, and directly change the original interface will cause build failure, so add a new method instead to be backward compatible.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - No release note


pkg/restore/split_client.go
```diff
@@ -51,6 +51,8 @@ type SplitClient interface {
 	// BatchSplitRegions splits a region from a batch of keys.
 	// note: the keys should not be encoded
 	BatchSplitRegions(ctx context.Context, regionInfo *RegionInfo, keys [][]byte) ([]*RegionInfo, error)
+	// BatchSplitRegionsRaw splits a region from a batch of keys and return the original region and split new regions
+	BatchSplitRegionsRaw(ctx context.Context, regionInfo *RegionInfo, keys [][]byte) (*RegionInfo, []*RegionInfo, error)
```
kennytm, line 55: maybe better call it "BatchSplitRegionsWithOriginal" than calling it "raw".

kennytm (commented):
rest LGTM

3pointer (approved):
LGTM

PR #645, opened 2020-12-09
log_restore: refactor write and ingest logic.

3pointer:
Both Lightning and BR have a similar logic. write and ingest kv as well as retry policy. This abstract Ingest method for BR and Lightning easy to use.

1. Refactor write and ingest code from log_client to `Ingestser`.

### Check List 

Tests 

 - No code

Related changes

### Release Note

 - No release note.


pkg/kv/kv.go
```diff
@@ -30,6 +30,95 @@ import (
...
+	return len(s.pairs) == 0
+}
+
+// Error implements PairIter.Error
+func (s *SimplePairIter) Error() error {
+	return nil
+}
+
+// First implements PairIter.First
+func (s *SimplePairIter) First() []byte {
```
glorv, line 81: The `First` and `Last` should also have a `seek to` semantic meaning, so call of these two functions should also set the index to target position

pkg/kv/kv.go
```diff
@@ -30,6 +30,95 @@ import (
...
+// which is used for log restore.
+type SimplePairIter struct {
+	index int
+	pairs Pairs
+}
+
+// NewSimpleKeyIter creates SimplePairIter.
+func NewSimpleKeyIter(pairs Pairs) PairIter {
+	return &SimplePairIter{
+		index: 0,
```
glorv, line 65: With a fresh new iterator without calling `Next` for the first time, the iterator should be invalid then. And the first call `Next`'s return value should be the same as `First`

pkg/kv/kv.go
```diff
@@ -309,3 +398,15 @@ func (kvs Pairs) ClassifyAndAppend(
 func (kvs Pairs) Clear() Pairs {
 	return kvs[:0]
 }
+
+// NextKey return the smallest []byte that is bigger than current bytes.
+// special case when key is empty, empty bytes means infinity in our context, so directly return itself.
+func NextKey(key []byte) []byte {
```
glorv, line 404: maybe this should also apply https://github.com/pingcap/tidb-lightning/pull/523

pkg/restore/ingester.go
```diff
@@ -0,0 +1,573 @@
...
+	TS uint64
+
+	conns gRPCConns
+
+	splitCli   SplitClient
+	workerPool *utils.WorkerPool
+
+	batchWriteKVPairs int
+	regionSplitSize   int64
+	tcpConcurrency    int
```
glorv, line 89: this field is useless?

glorv, line 89: I think it's better to put this field into `gRPCCoons` since it means the capacity of the connection pool

pkg/restore/ingester.go
```diff
@@ -0,0 +1,573 @@
...
+				var retryTy retryType
+				var newRegion *RegionInfo
+				retryTy, newRegion, err = i.isIngestRetryable(ctx, resp, region, meta)
+				if err == nil {
+					// ingest next meta
+					break
+				}
+				switch retryTy {
+				case retryNone:
+					log.L().Warn("ingest failed and do not retry", zap.Error(err), zap.Reflect("meta", meta),
```
glorv, line 261: use `log.ShortError` instead

pkg/restore/range.go
```diff
@@ -17,6 +19,44 @@ import (
 	"github.com/pingcap/br/pkg/rtree"
 )
 
+// Range record start and end key for localStoreDir.DB
+// so we can write it to tikv in streaming
+type Range struct {
+	Start  []byte
+	End    []byte
+	Length int
```
glorv, line 27: Seems the `Length` field is never used, maybe we can just remove it

3pointer: Sure, it has been moved.

pkg/restore/ingester.go
3pointer: Done

pkg/kv/kv.go
3pointer: Done

pkg/kv/kv.go
3pointer: Done, PTAL

pkg/kv/kv.go
3pointer: Done

pkg/kv/kv.go
```diff
@@ -32,6 +35,141 @@ import (
...
+// Produce implements Iter.Producer.Produce.
+func (p *SimpleKVIterProducer) Produce(start []byte, end []byte) Iter {
+	startIndex := sort.Search(len(p.pairs), func(i int) bool {
+		return bytes.Compare(start, p.pairs[i].Key) < 1
+	})
+	endIndex := sort.Search(len(p.pairs), func(i int) bool {
+		return bytes.Compare(end, p.pairs[i].Key) < 1
+	})
+	if startIndex >= endIndex {
+		log.Error("produce failed due to start key is large than end key",
```
glorv, line 88:
```suggestion
		log.WARN("produce failed due to start key is large than end key",
```

pkg/kv/kv.go
```diff
@@ -32,6 +35,141 @@ import (
...
 
+// Iter abstract iterator method for Ingester.
+type Iter interface {
+	// Seek seek to specify position.
+	// if key not found, seeks next key position in iter.
+	Seek(key []byte) bool
+	// Error return current error on this iter.
+	Error() error
+	// First return the first key in this iter.
+	First() []byte
```
glorv, line 46:
```suggestion
	First() bool
```
Return a bool to indicate if the key is valid

pkg/kv/kv.go
```diff
@@ -32,6 +35,141 @@ import (
...
+	// Last return the last key in this iter.
+	Last() []byte
```
glorv, line 48: ditto

glorv (commented):
rest LGTM

glorv:
LGTM

pkg/kv/kv.go
```diff
@@ -32,6 +35,141 @@ import (
...
+type Iter interface {
+	// Seek seek to specify position.
+	// if key not found, seeks next key position in iter.
+	Seek(key []byte) bool
+	// Error return current error on this iter.
+	Error() error
+	// First return the first key in this iter.
+	First() bool
+	// Last return the last key in this iter.
+	Last() bool
```
kennytm, line 48:
```suggestion
	// First moves this iter to the first key.
	First() bool
	// Last moves this iter to the last key.
	Last() bool
```

pkg/kv/kv.go
```diff
@@ -32,6 +35,141 @@ import (
...
+
+// Valid implements Iter.Valid.
+func (s *SimpleKVIter) Valid() bool {
+	return s.index >= 0 && s.index < len(s.pairs)
+}
+
+// Next implements Iter.Next.
+func (s *SimpleKVIter) Next() bool {
+	s.index++
+	return s.index <= len(s.pairs)
```
kennytm, line 149: why use `<=` instead of `<` here 🤔 what does the boolean return supposed to mean

pkg/restore/range_test.go
```diff
@@ -1,6 +1,6 @@
 // Copyright 2020 PingCAP, Inc. Licensed under Apache-2.0.
 
-package restore_test
+package restore
```
kennytm, line 3: ????

pkg/restore/split_client.go
```diff
@@ -478,3 +478,12 @@ func (c *pdClient) getPDAPIAddr() string {
...
 	return strings.TrimRight(addr, "/")
 }
+
+func checkRegionEpoch(new, old *RegionInfo) bool {
+	if new.Region.GetId() == old.Region.GetId() &&
+		new.Region.GetRegionEpoch().GetVersion() == old.Region.GetRegionEpoch().GetVersion() &&
+		new.Region.GetRegionEpoch().GetConfVer() == old.Region.GetRegionEpoch().GetConfVer() {
+		return true
+	}
+	return false
```
kennytm, line 488:
```suggestion
	return new.Region.GetId() == old.Region.GetId() &&
		new.Region.GetRegionEpoch().GetVersion() == old.Region.GetRegionEpoch().GetVersion() &&
		new.Region.GetRegionEpoch().GetConfVer() == old.Region.GetRegionEpoch().GetConfVer()
```

pkg/kv/kv.go
3pointer: fixed.`<=` is meaningless

ti-srebot (approved):

Issue #651, opened 2020-12-18
refactor storage.ExternalStorage interface


lichunzhu:
### Describe your feature request related problem:

Currently, `ExternalStorage`'s logic has some problems:
1. `ExternalStorage.Write` and `ExternalStorage.Read` are hardly used. In most situations, we need streaming write & read.
2. It's hard to use `multi-uploader` now. We need to do these things to open a `UploderWriter`:
  a. Create a new `ExternalStorage`.
  b. Create a new `Uploader` through `ExternalStorage.CreateUploader`.
  c. Create a new `Writer` though `NewUploaderWriter`. Then use `Writer.Write` and `Writer.Close` to upload.
3. The concept of `ExternalStorage.Write` and `Writer.Write` are confusing. Users need to read our code to know what do they do.
4. It's hard to add the compress option in the future.

### Describe the feature you'd like:

Delete `Uploader`, `Writer`, refactor `ExternalStorage` and add interface `ExternalFile`.

```go
type ExternalStorage interface {
        // Equal to original ExternalStorage.Write, similar to ioutil.WriteFile
	WriteFile(ctx context.Context, name string, data []byte) error
        // Equal to original ExternalStorage.Read, similar to ioutil.ReadFile
	ReadFile(ctx context.Context, name string) ([]byte, error)
	FileExists(ctx context.Context, name string) (bool, error)
        // Open a file reader
	Open(ctx context.Context, path string) (ExternalFileReader, error)
        // Open a file writer
        Create(ctx context.Context, path string) (ExternalFileWriter, error)
	WalkDir(ctx context.Context, opt *WalkOption, fn func(path string, size int64) error) error
	URI() string
}

// Streaming external file reader
type ExternalFileReader interface {
        io.Reader
        io.Seeker
        io.Closer
}

// Streaming external file writer
type ExternalFileWriter interface {
        io.WriteCloser
}

type withCompression struct {
        inner ExternalStorage
        compressionType CompressionType
}

func WithCompression(inner ExternalStorage, compressionType CompressionType) ExternalStorage {
    return &withCompression{inner: inner, compressionType: compressionType}
}
```

### Describe alternatives you've considered:

None

### Teachability, Documentation, Adoption, Migration Strategy:

None

kennytm:
`WithCompress` shouldn't be a method of `ExternalStorage`

lichunzhu:
> `WithCompress` shouldn't be a method of `ExternalStorage`

[...]

PR #676, opened 2021-01-05
storage/: refactor storage.ExternalStorage interface

lichunzhu:
fix https://github.com/pingcap/br/issues/651

1. Add `WithCompression` to generate storage with compress option.
2. Refactor `storage.ExternalStorage` as we depict in the proposal.

### Check List 

Tests 

 - Unit test
 - Integration test
 
Code changes

Side effects

 - Increased code complexity

Related changes

### Release Note

 - refactor storage.ExternalStorage interface to support compress


pkg/storage/storage.go
```diff
@@ -75,10 +75,22 @@ type ExternalStorage interface {
...
+}
+
+// ExternalFileReader represents a kind of file system storage.
+type ExternalFileReader interface {
+	io.ReadCloser
+	io.Seeker
+}
+
+// ExternalFileWriter represents a kind of file system storage.
+type ExternalFileWriter interface {
```
kennytm, line 89: inaccurate doc comment

pkg/storage/uploader.go
```diff
@@ -203,3 +239,15 @@ func (u *BufferWriter) Reset() {
...
+func CreateUploader(ctx context.Context, s ExternalStorage, name string) (Uploader, error) {
+	switch storage := s.(type) {
+	case *S3Storage:
+		return storage.CreateUploader(ctx, name)
+	case *LocalStorage:
+		return storage.CreateUploader(ctx, name)
+	default:
+		return nil, errors.Annotatef(berrors.ErrStorageInvalidConfig, "unsupported externalStorage type %T", s)
+	}
+}
```
kennytm, line 253: why do we still need this function

pkg/storage/compress.go
```diff
@@ -0,0 +1,80 @@
...
+	compressType CompressType
+}
+
+// UnwrapCompression unwraps withCompression to original ExternalStorage
+func UnwrapCompression(storage ExternalStorage) ExternalStorage {
+	if compressExt, ok := storage.(*withCompression); ok {
+		return UnwrapCompression(compressExt.ExternalStorage)
+	}
+	return storage
+}
```
kennytm, line 24: what's the purpose of this method

pkg/storage/compress.go
```diff
@@ -0,0 +1,80 @@
...
+
+// WithCompression returns an ExternalStorage with compress option
+func WithCompression(inner ExternalStorage, compressionType CompressType) ExternalStorage {
+	if compressionType == NoCompression {
+		return UnwrapCompression(inner)
+	}
+	return &withCompression{ExternalStorage: UnwrapCompression(inner), compressType: compressionType}
+}
```
kennytm, line 32: `WithCompression` should *add* compression to an external storage, not replace it. If we make `WithCompression(WithCompression(inner, Gzip), Gzip)` it should creates a storage that compresses twice (no matter how silly it looks).

pkg/storage/uploader.go
lichunzhu: Because we still have the concept of `pure uploader` and `uploader with buffer cache` so I kept the `Uploader` interface. But the users don't need to `CreateUploader` manually anymore. Any advice?

kennytm: I think only CDC needs `CreateUploader`?

lichunzhu: After this PR no one will need `CreateUploader` anymore. If CDC wants to open a streaming file write, it should use `ExternalStorage.Open` and `ExternalFile.Write`.

But, we still have the concepts of `Uploader` and `UploaderWriter` but they are not exposed to users.

For `Uploader`, it will upload data to cloud storage and complete the upload when the file is finished. It will upload data directly without any compression or buffer.

`uploaderWriter` will wrap `Uploader` and buffer the data to make it satisfy the upload requirements (for example: s3 5M). `uploaderWriter` can also help buffer and compress the data to upload.

So I still keep this function.

kennytm: Open() doesn't create a writable file. You mean Create()? 

And this Uploader sounds like an implementation detail now. Why do we still need an abstract Uploader interface then? For instance, in Local FS and noop:// [*and GCS*](https://pkg.go.dev/cloud.google.com/go/storage#Writer) we don't need any chunking behavior and can skip the `UploaderWriter` entirely; while in S3 we automatically buffer by 5 MB and use the UploadPart API.

lichunzhu: Yes, I mean Create().

overvenus:
LGTM

ti-srebot (approved):

Issue #675, opened 2021-01-04
BR 4.0.9 is incompatible with 4.0.8 when using GCS to store backup


overvenus:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?

Backup TiDB to GCS with a URL has no tail '/' via BR 4.0.8.
Restore TiDB from GCS.

2. What did you expect to see?

Restore success.

3. What did you see instead?

Restore fails.

------------

This incompatibility is caused by https://github.com/pingcap/br/pull/647. We need to add a fullback to compatible with older backup (<=4.0.8)

kennytm:
The bug fixed by #647 has two effects.

1. If the storage is set as `gcs://bucket/prefix`, the SSTs are written correctly to `gcs://bucket/prefix/*.sst` but the backupmeta is written wrongly to `gcs://bucket/prefixbackupmeta`.
2. If the storage is set as `gcs://bucket/prefix/`, the backupmeta is written correctly to `gcs://bucket/prefix/backupmeta`, but the SSTs are written wrongly to `gcs://bucket/prefix//*.sst` (note the extra slash).

Case 1 can be supported simply by adding a `--backupmeta-file` flag to specify an alternative path of the backupmeta.

Case 2 is hard to fix when #647 is in effect since we Trim() all slashes, meaning `-s gcs://bucket/prefix//` is equivalent to `-s gcs://bucket/prefix`, and it won't let TiKV find `gcs://bucket/prefix//*.sst`.

kennytm:
Moving (renaming) objects in GCS: https://cloud.google.com/storage/docs/gsutil/commands/mv

Like S3, moving is equivalent to Copy + Delete and is not free.

PR #647, opened 2020-12-14
gcs: fix wrong path of gcs prefix

3pointer:
### What problem does this PR solve? 
backup meta file out of backup directory in gcs storage.
### What is changed and how it works?
ensure gcs prefix endwith "/"

### Check List 

Tests 

 - Unit test
 - Manual test (add detailed scripts or steps below)
before:
image
after:
image

Related changes

 - Need to cherry-pick to the release branch

### Release Note

 - Fix the issue that write backup meta file out of backup directory in gcs storage.

PR #677, opened 2021-01-05
gcs: compatible with old gcs bug

3pointer:
solve #675

for case 1: add a check before Read & Write meta file, decide which prefix to use.
for case 2: check whether sst file is in Prefix(after #647) or in PrefixSlash(before #647), adjust proper prefix for TiKV request.
### Check List 

Tests 
 - Integration test

Related changes

### Release Note

 - Fix the issue that br 4.0.9 cannot restore br 4.0.8 gcs backup files.


kennytm (commented):
This PR will also affect all non-BR users of this package. can you do we fix outside `gcsStorage`? Especially not do the request inside `objectName` since it affects all GCS calls not just "reading backupmeta"

3pointer:
> This PR will also affect all non-BR users of this package. can you do we fix outside `gcsStorage`? Especially not do the request inside `objectName` since it affects all GCS calls not just "reading backupmeta"

You are right. Let me think

pkg/task/common.go
```diff
@@ -477,3 +498,9 @@ func normalizePDURL(pd string, useTLS bool) (string, error) {
 	}
 	return pd, nil
 }
+
+// check whether it's a bug before #647, to solve case #1
+// see details https://github.com/pingcap/br/issues/675#issuecomment-753780742
```
overvenus, line 503: Please inline comments in the link here.

overvenus (commented):
Please move compatibility code to a new package `compat`.

overvenus:
LGTM

go.mod1
```diff
@@ -38,3 +38,8 @@ require (
 	google.golang.org/grpc v1.27.1
 	sourcegraph.com/sourcegraph/appdash v0.0.0-20190731080439-ebfcffb1b5c0
 )
+
+replace (
+	cloud.google.com/go/storage => github.com/3pointer/google-cloud-go/storage v1.6.1-0.20210108125931-b59bfa0720b2
```
kennytm, line 43: (googleapis/google-cloud-go#3509 for future reference)

go.mod1
```diff
@@ -38,3 +38,8 @@ require (
...
+	golang.org/x/sys => golang.org/x/sys v0.0.0-20200625212154-ddb9806d33ae
```
kennytm, line 44: but why have this replace?

pkg/storage/gcs.go
```diff
@@ -205,3 +230,22 @@ func newGCSStorage(ctx context.Context, gcs *backup.GCS, opts *ExternalStorageOp
 	}
 	return &gcsStorage{gcs: gcs, bucket: bucket}, nil
 }
+
+func hasSSTFiles(ctx context.Context, bucket *storage.BucketHandle, prefix string) bool {
+	it := bucket.Objects(ctx, &storage.Query{Prefix: prefix})
```
kennytm, line 235: consider [fetching the `Name` field only](https://pkg.go.dev/cloud.google.com/go/storage#Query.SetAttrSelection):

```suggestion
	query := storage.Query{Prefix: prefix}
	query.SetAttrSelection([]string{"Name"})
	it := bucket.Objects(ctx, &query)
```

pkg/task/common.go
```diff
@@ -7,9 +7,11 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/url"
+	filepath "path"
```
kennytm, line 10: ??? there isn't another `path` package. calling it `filepath` is confusing since `path/filepath` package does exist.

tests/br_gcs/oauth.py
```diff
@@ -0,0 +1,7 @@
+from flask import Flask
```
kennytm, line 1:
```suggestion
#!/usr/bin/env python3

from flask import Flask
```

kennytm (commented):
rest LGTM

pkg/task/common.go
3pointer: yes, my mistake, I didn't notice IDE add this package automatically...

go.mod1
3pointer: It used to avoid some complie error on my mac os when I try to upgrade  cloud.google.com/go/storage from `v1.6.0` to `v1.12.0`. but for now it can be removed. because I have picked googleapis/google-cloud-go#3509 to v1.6.0 already.

pkg/task/common.go
```diff
@@ -405,7 +407,27 @@ func ReadBackupMeta(
...
 	metaData, err := s.ReadFile(ctx, fileName)
 	if err != nil {
-		return nil, nil, nil, errors.Annotate(err, "load backupmeta failed")
+		if gcsObjectNotFound(err) {
+			// change gcs://bucket/abc/def to gcs://bucket/abc and read defbackupmeta
+			parsedURL, err := storage.ParseRawURL(cfg.Storage)
+			if err != nil {
+				return nil, nil, nil, errors.Trace(err)
+			}
+			newPrefix, file := filepath.Split(parsedURL.Path)
```
kennytm, line 416: I mean we should just use `path.Split` here since the path comes from a URL which always use `/` as separator, instead of `filepath.Split` which the separator depends on the OS

pkg/task/common.go
```diff
@@ -405,7 +407,27 @@ func ReadBackupMeta(
...
+			newFileName := file + fileName
+			cfg.Storage = strings.ReplaceAll(cfg.Storage, parsedURL.Path, newPrefix)
```
kennytm, line 418: Can you reassemble the storage from `u` instead?

kennytm (commented):
LGTM

ti-srebot (approved):

PR #685, opened 2021-01-08
restore: fast fail when backup table's cluster index option is not consistent

lichunzhu:
If we backup table with cluster index option A, restore table with cluster index option B. If A != B, the table data might be corrupted.

Check table's info with existed tables before exec DDL and restore tables.

### Check List 

Tests 

 - Unit test
 - Integration test
1. Backup tables with `tidb_enable_clustered_index=OFF`.
2. Set `tidb_enable_clustered_index=ON`.
3. Restore tables through BR and get:
```
Error: CommonIsHandle: backup table false, existed table true not match: [BR:Restore:ErrRestoreModeMismatch]restore mode mismatch
```

Code changes

### Release Note

 - fast fail when backup table's cluster index option is not consistent

### DIFF
--- a/pkg/restore/client.go
+++ b/pkg/restore/client.go
@@ -1007,6 +1007,7 @@ func (rc *Client) IsSkipCreateSQL() bool {
 	return rc.noSchema
 }
 
+// PreCheckTableClusterIndex checks whether backup tables and existed tables have different cluster index options
 func (rc *Client) PreCheckTableClusterIndex(
 	tables []*utils.Table,
 	ddlJobs []*model.Job,
--- a/pkg/restore/client.go
+++ b/pkg/restore/client.go
@@ -772,11 +772,13 @@ func (rc *Client) GoValidateChecksum(
 }
 
 func (rc *Client) execChecksum(ctx context.Context, tbl CreatedTable, kvClient kv.Client, concurrency uint) error {
+	logger := log.With(
+		zap.String("db", tbl.OldTable.DB.Name.O),
+		zap.String("table", tbl.OldTable.Info.Name.O),
+	)
+
 	if tbl.OldTable.NoChecksum() {
-		log.Warn("table has no checksum, skipping checksum",
-			zap.Stringer("table", tbl.OldTable.Info.Name),
-			zap.Stringer("database", tbl.OldTable.DB.Name),
-		)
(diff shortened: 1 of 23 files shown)


pkg/restore/client.go
```diff
@@ -400,6 +400,12 @@ func (rc *Client) createTable(
 	if err != nil {
 		return CreatedTable{}, errors.Trace(err)
 	}
+	if newTableInfo.IsCommonHandle != table.Info.IsCommonHandle {
+		return CreatedTable{}, errors.Annotatef(berrors.ErrRestoreModeMismatch,
+			"CommonIsHandle: backup table %v, new table %v not match",
```
kennytm, line 405: the wording is too cryptic for users.

lichunzhu: 🤔 How about `ClusterIndex option: backup table %v, created table %v not match`?

kennytm: > Clustered index option mismatch. Restored cluster's @@tidb_enable_clustered_index should be %[1]v (backup table = %[1]v, created table = %[2]v).

kennytm (commented):
please add a test case. rest LGTM.

pkg/restore/client.go
```diff
@@ -1000,3 +1007,50 @@ func (rc *Client) EnableSkipCreateSQL() {
...
 }
+
+// PreCheckTableClusterIndex checks whether backup tables and existed tables have different cluster index options。
+func (rc *Client) PreCheckTableClusterIndex(
+	tables []*utils.Table,
+	ddlJobs []*model.Job,
+	dom *domain.Domain,
+) error {
+	for _, table := range tables {
+		oldTableInfo, err := rc.GetTableSchema(dom, table.DB.Name, table.Info.Name)
```
3pointer, line 1018: rest LGTM.
`oldTableInfo` should be `newTableInfo`, just like L403

lichunzhu: L403 is table info caught from database after table is created.
L1018 is table info before table is created.
So I think they should be different.

kennytm (commented):
LGTM

3pointer (approved):
LGTM

ti-srebot (approved):

PR #653, opened 2020-12-21
*: redact log and error messages, add log-redact parameter

lichunzhu:
### What problem does this PR solve? 
Current BR will print some sensitive information in log and error messages.

### What is changed and how it works?
1. add `log-redact` parameter. When it's set to `true`, all sensitive info will be deleted from logs.
2. Remove sensitive info in error messages and print them to logs. 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

### Release Note

 - support --log-redact parameter

PR #692, opened 2021-01-12
cmd/: change redact log parameter name

lichunzhu:
In https://github.com/pingcap/br/pull/653, we added the `redact-log` parameter.

The redact log parameter name is different from TiKV's and PD's.
https://github.com/tikv/tikv/blob/master/etc/config-template.toml#L846
In TiKV it's `redact-info-log` configuration in config file.
https://github.com/tikv/pd/blob/master/server/config/config.go#L1373
In PD it's `redact-info-log` configuration in config file.

Change `redact-log` to `redact-info-log`.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes

### Release Note

 - No release note

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -43,8 +43,10 @@ const (
 	FlagStatusAddr = "status-addr"
 	// FlagSlowLogFile is the name of slow-log-file flag.
 	FlagSlowLogFile = "slow-log-file"
-	// FlagRedactLog is whether to redact sensitive information in log.
-	FlagRedactLog = "redact-info-log"
+	// FlagRedactLog is whether to redact sensitive information in log, already deprecated by FlagRedactInfoLog
+	FlagRedactLog = "redact-log"
+	// FlagRedactInfoLog is whether to redact sensitive information in log.
+	FlagRedactInfoLog = "redact-info-log"
 
 	flagVersion      = "version"
 	flagVersionShort = "V"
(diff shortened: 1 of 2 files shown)


overvenus (approved):
LGTM

cmd/cmd.go
```diff
@@ -44,7 +44,7 @@ const (
 	// FlagSlowLogFile is the name of slow-log-file flag.
 	FlagSlowLogFile = "slow-log-file"
 	// FlagRedactLog is whether to redact sensitive information in log.
-	FlagRedactLog = "redact-log"
+	FlagRedactLog = "redact-info-log"
```
3pointer, line 47: 🤔️ can we keep both "redact-log" and "redact-info-log" for compatibility.

overvenus: Seems it will be release in 5.0.0-rc.

lichunzhu: I have already added the compatibility support. PTAL again.

cmd/cmd.go
```diff
@@ -43,8 +43,10 @@ const (
 	FlagStatusAddr = "status-addr"
 	// FlagSlowLogFile is the name of slow-log-file flag.
 	FlagSlowLogFile = "slow-log-file"
-	// FlagRedactLog is whether to redact sensitive information in log.
-	FlagRedactLog = "redact-info-log"
+	// FlagRedactLog is whether to redact sensitive information in log, already deprecated by FlagRedactInfoLog
+	FlagRedactLog = "redact-log"
```
overvenus, line 47: Could you add a test in tests/br_other?

lichunzhu: addressed in https://github.com/pingcap/br/pull/692/commits/3fd299bf1e3e49da29dad8728b5d782e7672c48b

overvenus (commented):
LGTM

cmd/cmd.go
kennytm: i don't think we need to support `--redact-log` if it does not appear in any GA version.

PR #693, opened 2021-01-14
stats: disable stats by default

3pointer:
Disable backup stats by default. because of
1. DumpStatsToJson is not stable
2. It increases memory usage may cause BR OOM

1. set `cfg.IgnoreStats` default to true
2. if `cfg.IgnoreStats` is true, set `statsLease` to -1

### Check List 

Tests 

 - Integration test

Related changes

### Release Note

 - Disable backup stats by default.

### DIFF
--- a/cmd/backup.go
+++ b/cmd/backup.go
@@ -6,6 +6,7 @@ import (
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"github.com/pingcap/tidb/ddl"
+	"github.com/pingcap/tidb/session"
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 	"sourcegraph.com/sourcegraph/appdash"
@@ -30,6 +31,11 @@ func runBackupCommand(command *cobra.Command, cmdName string) error {
 		ctx, store = trace.TracerStartSpan(ctx)
 		defer trace.TracerFinishSpan(ctx, store)
 	}
+	if cfg.IgnoreStats {
+		// Do not run stat worker in BR.
+		session.DisableStats4Test()
+	}
+
 	if err := task.RunBackup(ctx, tidbGlue, cmdName, &cfg); err != nil {
 		log.Error("failed to backup", zap.Error(err))
 		return errors.Trace(err)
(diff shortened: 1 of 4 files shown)


cmd/backup.go
```diff
@@ -30,6 +31,11 @@ func runBackupCommand(command *cobra.Command, cmdName string) error {
 		ctx, store = trace.TracerStartSpan(ctx)
 		defer trace.TracerFinishSpan(ctx, store)
 	}
+	if cfg.IgnoreStats {
+		// Do not run stat worker in BR.
+		session.DisableStats4Test()
```
overvenus, line 36: What about restore?

3pointer: It's not easy to distinguish whether we should `DisableStats4Test` in restoration. I had two ideas
1.  Read backupmeta and check all tables whether stats != nil. but we should do it carefully, because read backupmeta is in `task.runRestore` and we cannot run `DisableStats4Test` otherwise it will influence BR via SQL.
2. Change kvproto to record the status of `--ignore-stats`. and this will involve other repo. 
So both of them will change a lot. we need to confirm the cost of `DisableStats4Test` is worth for above changes.

kennytm (commented):
lgtm

cmd/backup.go
3pointer: [LoadStatsFromJson](https://github.com/pingcap/tidb/blob/76054200b6340f064e25046038d1ac1fdbe7e4ff/statistics/handle/dump.go#L198) only execute SQL to storage. so we can disable stats loop directly.

3pointer:
Test restore with sysbench 3000 tables.
v4.0.8 mem inuse 61M
v4.0.9 mem inuse 88.96M
v4.0.9 + #693 mem inuse 81.66M

overvenus:
LGTM

ti-srebot (approved):

3pointer:
Test backup with sysbench 3000 tables.
v4.0.9 + #693 inuse Showing nodes accounting for 21.65MB, 72.98% of 29.66MB total
v4.0.9 inuse Showing nodes accounting for 30.05MB, 62.90% of 47.78MB total

We can reduce the memory inuse with this PR.

Issue #736, opened 2021-02-08
Add log to make error more clear when restore file does not exist using s3 sink


yuqi1129:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

./br restore cdclog --pd=http://10.142.99.8:2379 --start-ts=422756483571646465 --end-ts=422756525015040000  -s "s3://tidb/tx31?endpoint=http://endpoint"

2. What did you expect to see?

```
Error: NoSuchKey: The specified key does not exist.
	status code: 404, request id: e12fcec0-1ca2-0e8c-0000-01777fbe7ce4, host id:
```
and log is 
```
[...]

PR #737, opened 2021-02-08
Add log to make error more clear when restore error using s3 sink

yuqi1129:

see https://github.com/pingcap/br/issues/736

Just add some log

### Check List 

Tests 

 - No code

Code changes

 - Has persistent data change

Side effects

None

Related changes

### Release Note

 -


pkg/storage/s3.go
```diff
@@ -338,6 +338,7 @@ func (rs *S3Storage) ReadFile(ctx context.Context, file string) ([]byte, error)
 
 	result, err := rs.svc.GetObjectWithContext(ctx, input)
 	if err != nil {
+		log.Error("In read s3 file", zap.Any("file info", input))
 		return nil, errors.Trace(err)
```
3pointer, line 342: I think it's better to wrap this info in error. such as `	return nil, errors.Annotatef(err,"failed to read s3 file in bucket %s %s.", input.Bucket, input.Key)
`

yuqi1129: @3pointer OK, I will make a change

pkg/storage/s3.go
```diff
@@ -338,7 +338,10 @@ func (rs *S3Storage) ReadFile(ctx context.Context, file string) ([]byte, error)
 
 	result, err := rs.svc.GetObjectWithContext(ctx, input)
 	if err != nil {
-		return nil, errors.Trace(err)
+		er := errors.Trace(err)
```
3pointer, line 341: `Annotatef` already record stack info. so this line can be removed

yuqi1129: Thanks for your kind review, i will update the code later

3pointer (commented):
LGTM

ti-srebot (approved):

kennytm:
/lgtm cancel

```
[2021-02-19T16:14:45.874Z] pkg/storage/s3.go:341:15: printf: Annotatef format %s has arg input.Bucket of wrong type *string (govet)
[2021-02-19T16:14:45.874Z] 		return nil, errors.Annotatef(err,
[2021-02-19T16:14:45.874Z] 		            ^
```

kennytm:
```

[2021-02-20T03:02:28.690Z] ----------------------------------------------------------------------
[2021-02-20T03:02:28.690Z] FAIL: s3_test.go:547: s3Suite.TestReadError
[2021-02-20T03:02:28.690Z] 
[2021-02-20T03:02:28.690Z] s3_test.go:559:
[2021-02-20T03:02:28.690Z]     c.Assert(err, ErrorMatches, `\Q`+expectedErr.Error()+`\E`)
[2021-02-20T03:02:28.690Z] ... error string = "failed to read s3 file, file info: input.bucket='bucket', input.key='prefix/file-missing': NoSuchKey: no such key"
[2021-02-20T03:02:28.690Z] ... regex string = "\\QNoSuchKey: no such key\\E"
[2021-02-20T03:02:28.690Z] 
[2021-02-20T03:02:28.690Z] OOPS: 26 passed, 1 FAILED
```

yuqi1129:
@kennytm It's seems that we should not add info in the stack, some unit test verify the error string , i suggest just add a log to output the related info and revert the following comment?

kennytm:
@yuqi1129 we could also change the unit test. but wdyt @3pointer

3pointer:
> @yuqi1129 we could also change the unit test. but wdyt @3pointer

I think change the unit test is better. This PR aims to make the s3 read error clear. and the former error message "no such key" should be changed. @yuqi1129 could you change the test error message ?https://github.com/pingcap/br/blob/578be7fb516599dc2c6aca2bfeec73cb9c378943/pkg/storage/s3_test.go#L559

yuqi1129:
@3pointer , No problem

PR #739, opened 2021-02-19
Log fold is unstable when doing create and drop table operation and thus affect br restore using cdc log

yuqi1129:
see https://github.com/pingcap/ticdc/issues/1415

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release Note

 -

### DIFF
--- a/pkg/restore/log_client.go
+++ b/pkg/restore/log_client.go
@@ -91,6 +91,11 @@ type LogClient struct {
 	dropTSMap sync.Map
 }
 
+// GetMeta is for test
+func (l *LogClient) GetMeta() *LogMeta {
+	return l.meta
+}
+
 // NewLogRestoreClient returns a new LogRestoreClient.
 func NewLogRestoreClient(
 	ctx context.Context,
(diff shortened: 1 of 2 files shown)


pkg/restore/log_client_test.go
```diff
@@ -96,3 +96,35 @@ func (s *testLogRestoreSuite) TestTsInRange(c *C) {
...
+	meta.Names[23] = schemaAndTable
+
+	maps := s.client.GetNameIDMap()
+	ids, err := maps[schemaAndTable]
+
+	c.Assert(err, IsNil)
+
+	l := len(ids) == 3
+
+	c.Assert(l, IsTrue)
```
kennytm, line 116:
```suggestion
	c.Assert(ids, HasLen, 3)
```

pkg/restore/log_client_test.go
```diff
@@ -96,3 +96,32 @@ func (s *testLogRestoreSuite) TestTsInRange(c *C) {
...
 }
+
+func (s *testLogRestoreSuite) TestCollectRowChangeFiles(c *C) {
+	meta := s.client.GetMeta()
+
+	schemaAndTable := "`db1`.`t1`"
+	// t1 with three table id
+	meta.Names[1] = schemaAndTable
+	meta.Names[2] = schemaAndTable
+	meta.Names[23] = schemaAndTable
```
3pointer, line 107: `23` should be `3` to pass this test.

3pointer:
I'm afraid this PR will introduce a new issue. I think the proper way to fix it is choose the proper table id directory with given `start-ts` and `end-ts`(not largest table id always) in collectRowChangeFiles.

yuqi1129:
Hi, as far as i can see if a table have several table ids, times spans in
each id directory will not intersect. When restore code will choose the
correct ids thats contains the restore spans. so logically this will be OK
someway, am i right?

Your suggestion is more detailed and nice than this pr, i will follow your
advice

3pointer <notifications@github.com>于2021年3月9日 周二11:25写道：

yuqi1129:
@3pointer 
```
	nameIdsMap := l.GetNameIDMap()

	log.Debug("nameIdsMap", zap.Any("name and ids:", nameIdsMap))
	for _, ids := range nameIdsMap {
		for _, tID := range ids {
			tableID := tID
			// FIXME update log meta logic here
			dir := fmt.Sprintf("%s%d", tableLogPrefix, tableID)
			opt := &storage.WalkOption{
				SubDir:    dir,
				ListCount: -1,
			}
			err := l.restoreClient.storage.WalkDir(ctx, opt, func(path string, size int64) error {
				fileName := filepath.Base(path)
				shouldRestore, err := l.NeedRestoreRowChange(fileName)
				if err != nil {
					return errors.Trace(err)
				}
				if shouldRestore {
					rowChangeFiles[tableID] = append(rowChangeFiles[tableID], path)
				}
				return nil
			})
			if err != nil {
				return nil, errors.Trace(err)
			}
		}
	}
```
Hi, The code thats follows `l.GetNameIDMap()` will traverse cdc storage directory and will choose proper ids thats contains the start time and end time

pkg/restore/log_client_test.go
yuqi1129: Sorry, i miss it...

pkg/restore/log_client.go
```diff
@@ -301,59 +311,62 @@ func (l *LogClient) NeedRestoreRowChange(fileName string) (bool, error) {
...
 }
 
-func (l *LogClient) collectRowChangeFiles(ctx context.Context) (map[int64][]string, error) {
-	// we should collect all related tables row change files
-	// by log meta info and by given table filter
-	rowChangeFiles := make(map[int64][]string)
-
-	// need collect restore tableIDs
-	tableIDs := make([]int64, 0, len(l.meta.Names))
+// SetMeta return the tablename to table ids map
```
kennytm, line 314:
```suggestion
// GetNameIDMap return the tablename to table ids map
```

tisonkun:
@kennytm @3pointer I guess this PR should migrate to pingcap/tidb now? Does it go in a right direction so that we can continue it?

PR #743, opened 2021-02-19
*: update tidb and prepare for upcoming refactor

disksing:
We are implementing refactor of tidb's go client, which involves some interface changes. Currently, br relies on the fact the `kv.Storage` is a subset of `tikv.Storage`, which will be false later. This PR prepares for later refactoring.

See https://github.com/pingcap/tidb/issues/22513

- Update tidb dependency
- Use `tikv.Storage` and `kv.Storage` separately

### Check List 

Tests 
 - Unit test

### Release Note
- No release note


AndreMouche (commented):
LGTM

pkg/conn/conn.go
```diff
@@ -98,10 +99,11 @@ func NewConnPool(cap int, newConn func(ctx context.Context) (*grpc.ClientConn, e
...
 type Mgr struct {
 	*pdutil.PdController
-	tlsConf  *tls.Config
-	dom      *domain.Domain
-	storage  tikv.Storage
-	grpcClis struct {
+	tlsConf   *tls.Config
+	dom       *domain.Domain
+	storage   kv.Storage
+	tikvStore tikv.Storage
```
overvenus, line 105: Can we keep just one storage?

overvenus (commented):
Please keep go.sum1 and remove `// indirect` dependency in go.mod1.

pkg/conn/conn.go
disksing: Sadly we do need two storages. In our plan, `kv.Storage` contains coprocessor related interfaces, and `tikv.Storage` contains client-go specific interfaces (such as `GetLockResolver`). BR needs both interfaces, so we have to use different storages here.

amyangfei:
By the way, TiCDC also uses the interface in both `kv.Storage` and `tikv.Storage`, would you please update TiCDC in the same way @disksing 
such as 
- `kv.Storage`: https://github.com/pingcap/ticdc/blob/395d94a42c4feb6e586ce28c4f676da4f2d39393/cdc/kv/store_op.go#L89
- `tikv.Storage`: https://github.com/pingcap/ticdc/blob/395d94a42c4feb6e586ce28c4f676da4f2d39393/cdc/puller/puller.go#L79

disksing:
@amyangfei no need to hurry, we can do it after tidb side is done because tidb does not import ticdc :), unless if you need to update tidb for other reason, in that case feel free to contact me!

kennytm:
🤔 not relevant to the validity of the PR itself, but it can't be built on Windows

```console
$ make tidy
go mod tidy
cp go.mod1 go.mod && cp go.sum1 go.sum
GO111MODULE=on go mod tidy
verifying github.com/oleiade/reflections@v1.0.0/go.mod: checksum mismatch
        downloaded: h1:rdFxbxq4QXVZWj0F+e9jqjDkc7dbp97vkRixKo2JR60=
        go.sum:     h1:RbATFBbKYkVdqmSFtx13Bb/tVhR0lgOBXunWTZKeL4w=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
make: *** [Makefile:229: tidy] Error 1
```

disksing:
@kennytm I can pass `go mod tidy` on my windows (with go1.16). Maybe something wrong with go version or related to dirty data cached in some go proxy.

overvenus:
LGTM

andylokandy (approved):
LGTM

overvenus:
@3pointer PTAL

3pointer (approved):
LGTM

ti-srebot (approved):

PR #753, opened 2021-02-25
linghtning/backend: optimize local writer concurrency and memory usage

glorv:
Optimize local writer performance and memory usage.

Optimize local writer by:
- Open an index LocalWriter for each chunk to avoid the bottle neck at write index kvs
- Replace the async ApplendRows with sync operation, to optimize the memory usage and simplify the logic
- decrease the encode/deliverLoop chan to decrease memory usage
- use a memory buffer to restore the temp kv pairs before write them to SST file
- Manually compaction small SST files into a single SST file and then ingest it into pebble

Bench result:
The benchmark tests were run on a 40core machine with an NVME disk. and based on the following three data and table schema:
- DataSet1. 14k warehouse tpcc data 
- DataSet2. 1k warehouse `order_line` table with 3 indexes.  Thus each row generates 4 kvs.
  ```
  PRIMARY KEY (`ol_w_id`,`ol_d_id`,`ol_o_id`,`ol_number`),
  KEY `idx_d_i` (`ol_d_id`, `ol_i_id`),
  KEY `idx_d_w_supply` (`ol_d_id`, `ol_w_id`, `ol_supply_w_id`)
  ```
- DataSet3. 1k warehouse `order_line` table with 3 indexes.  Thus each row generates 8 kvs.
  ```
  PRIMARY KEY (`ol_w_id`,`ol_d_id`,`ol_o_id`,`ol_number`),
  KEY `idx_d_i` (`ol_d_id`,`ol_i_id`),
  KEY `idx_d_w_supply` (`ol_d_id`,`ol_w_id`,`ol_supply_w_id`),
  KEY `idx_o_amount` (`ol_o_id`,`ol_amount`),
  KEY `idx_d_supply` (`ol_d_id`,`ol_supply_w_id`),
  KEY `idx_o_d_i` (`ol_o_id`,`ol_d_id`,`ol_i_id`),
  KEY `idx_i_id` (`ol_i_id`)
  ```

Bench Result:
| DataSet | Branch | Peak Memory | Cost Time |
| --- | --- | ---- | --- |
| DataSet1 | master | 34GB | 2h25m |
| DataSet1 | opt-local-writer | 29GB | 1h46m |
| DataSet2 | master | 30GB | 24m57s |
| DataSet2 | opt-local-writer | 30GB | 9m32s |
| DataSet3 | master | **64GB** | 1h22m |
| DataSet3 | opt-local-writer | 33GB | 33m |

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

### Release Note

 - Optimize lightning local backend memory efficiency and performance


pkg/lightning/backend/local/local.go
```diff
@@ -217,21 +239,41 @@ func (e *File) getEngineFileSize() backend.EngineFileSize {
...
+		DiskSize:    total.Size + pendingSize,
 		MemSize:     memSize,
 		IsImporting: e.isLocked(),
 	}
 }
 
+// rLock locks the local file with shard read state. Only used for flush and ingest SST files.
+func (e *File) rLock(state importMutexState) {
+	e.mutex.RLock()
+	e.isImportingAtomic.Store(uint32(state))
```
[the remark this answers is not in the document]
glorv: The atomic is used as a high-performance flag for the lock status. The lock is not used for the atomic variable. 😅

pkg/lightning/backend/local/local.go
```diff
@@ -217,21 +239,41 @@ func (e *File) getEngineFileSize() backend.EngineFileSize {
...
 	e.localWriters.Range(func(k, v interface{}) bool {
 		w := k.(*Writer)
-		memSize += w.writeBatch.totalSize
 		if w.writer != nil {
-			total.Size += int64(w.writer.writer.EstimatedSize())
+			memSize += int64(w.writer.writer.EstimatedSize())
+		} else {
+			// if kvs are still in memory, only calculate half of the total size
+			// in our tests, SST file size is about 50% of the raw kv size
+			memSize += w.batchSize / 2
```
Little-Wallace, line 247: Why half?

glorv: The SST compaction ratio is about 2, so only calculated 50% of the raw size as disk-quota size

pkg/lightning/backend/local/local.go
```diff
@@ -2435,7 +2435,12 @@ func (w *Writer) flush(ctx context.Context) error {
 func (w *Writer) Close(ctx context.Context) error {
 	defer w.kvBuffer.destroy()
 	defer w.local.localWriters.Delete(w)
-	return w.flush(ctx)
+	err := w.flush(ctx)
+	// FIXME: in theory this line is useless, but I our test with go1.15
```
amyangfei, line 2453: Is the behavior of GC different among different go runtime, from 1.13 ~ 1.16

glorv: I think GO had been optimizing its GC behaviors across these versions. But sadly, In my newest benchmark with go 1.16.3, it still has this issue. The memory usage increased up to more than 75GB within 10min. 😂

lance6716:
/retitle lightning/backend: optimize local writer concurrency and memory usage

Little-Wallace (approved):
LGTM

pkg/lightning/backend/local/local.go
```diff
@@ -258,25 +316,330 @@ func (e *File) unlock() {
...
+
+					if err := e.batchIngestSSTs(ingestMetas); err != nil {
+						e.setError(err)
+						return
+					}
+					seqLock.Lock()
+					finSeq := finishedSeq.Load()
+					if metas.seq == finSeq+1 {
+						finSeq = metas.seq
+						if len(inSyncSeqs.arr) > 0 {
```
lance6716, line 412:
```
for len > 0 {
  if xx {
    remove
  } else {
    break
  }
}
```

pkg/lightning/backend/local/local.go
```diff
@@ -258,25 +316,330 @@ func (e *File) unlock() {
...
+						// waiting for pending compaction tasks
+						flushQueue = append(flushQueue, flushSeq{ch: msg.flushCh, seq: curSeq})
+						seqLock.Unlock()
+					}
+				}
+				break
+			}
+			metasTmp = append(metasTmp, msg.meta)
+			// drain all the sst meta from the chan to make sure all the SSTs are processed before handle a flush msg.
+		inner:
```
lance6716, line 529: how about refine this two level loop into a state machine 😵

pkg/lightning/backend/local/local.go
```diff
@@ -153,7 +151,32 @@ type File struct {
...
 	// This should not be used as a "spin lock" indicator.
 	isImportingAtomic atomic.Uint32
-	mutex             sync.Mutex
+	// flush and ingest sst hold the rlock, other operation hold the wlock.
+	mutex sync.RWMutex
+
+	ctx          context.Context
+	cancel       context.CancelFunc
+	sstDir       string
+	sstMetasChan chan metaOrFlush
```
lance6716, line 160: seems the linearization point is when `metaOrFlush` enters this channel, so the order of concurrent calls of `addSST` and `flushEngineWithoutLock` is not guaranteed. Not sure if this will cause problems

glorv: The consistency guarantee is ensured by the caller. When flushEngineWithoutLock is called, all localwriter belongs to this engine are either closed or flushed, so before sending the flush message, all the related addSST calls are ensured to be done.

pkg/lightning/backend/local/local.go
```diff
@@ -218,21 +245,41 @@ func (e *File) getEngineFileSize() backend.EngineFileSize {
...
+func (e *File) rLock(state importMutexState) {
+	e.mutex.RLock()
+	e.isImportingAtomic.Store(uint32(state))
+}
+
+func (e *File) rUnlock() {
+	if e == nil {
+		return
+	}
+	e.isImportingAtomic.Store(0)
```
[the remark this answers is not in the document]
glorv: Good point.

pkg/lightning/backend/local/local.go
glorv: Is there a good example to refer to？😅

pkg/lightning/backend/local/local.go
```diff
@@ -208,8 +237,13 @@ func (e *File) getSizeProperties() (*sizeProperties, error) {
...
 }
 
+func isStateLocked(state importMutexState) bool {
+	return state&(importMutexStateClose|importMutexStateImport) != 0
+}
+
 func (e *File) isLocked() bool {
-	return e.isImportingAtomic.Load() != 0
+	// the engine is locked only in import or close state.
+	return isStateLocked(importMutexState(e.isImportingAtomic.Load()))
```
glorv, line 246: @kennytm PTAL. I'm not sure if we need to treat the `ReadLock` state (flush state) as locked or unlocked

pkg/lightning/backend/local/local.go
lance6716:
```golang
state := IDLE
for {
	switch state {
	case CLOSED:
		break
	case IDLE:
		// read a msg
		// set state to NEED_FLUSH or CACHED or CLOSED
	case NEED_FLUSH:
		// send pendingMetas or directly flush
		state = IDLE
	case CACHED:
		// read a msg
		// set state to NEED_FLUSH or CACHED or CLOSED
	}
}
```

not sure current loop could be converted easily 😵

pkg/lightning/backend/local/local.go
```diff
@@ -1942,142 +2486,107 @@ func (local *local) EngineFileSizes() (res []backend.EngineFileSize) {
...
-	if w.writer != nil {
-		if err := w.writer.writeKVs(&w.writeBatch); err != errorUnorderedSSTInsertion {
-			return err
-		}
-	}
+func (w *Writer) Close(ctx context.Context) error {
+	defer w.kvBuffer.destroy()
+	defer w.local.localWriters.Delete(w)
+	err := w.flush(ctx)
+	// FIXME: in theory this line is useless, but I our test with go1.15
```
gozssky, line 2546:
```suggestion
	// FIXME: in theory this line is useless, but I test with go1.15
```

pkg/lightning/restore/restore.go
```diff
@@ -1351,6 +1393,19 @@ func (tr *TableRestore) restoreEngines(pCtx context.Context, rc *Controller, cp
 		indexWorker := rc.indexWorkers.Apply()
 		defer rc.indexWorkers.Recycle(indexWorker)
 
+		engineCfg := &backend.EngineConfig{}
+		if rc.cfg.TikvImporter.Backend == config.BackendLocal {
+			idxCnt := len(tr.tableInfo.Core.Indices)
+			if tr.tableInfo.Core.PKIsHandle {
```
lance6716, line 1399: do we need to consider `IsCommonHandle`

pkg/lightning/restore/restore.go
```diff
@@ -1325,6 +1330,43 @@ func (tr *TableRestore) restoreTable(
 	return tr.postProcess(ctx, rc, cp, false /* force-analyze */)
 }
 
+// estimate SST files compression threshold by total row file size
+// with a higher compression threshold, the compression time increases, but the iteration time decreases.
+// Try to limit the total SST files number under 500. But size compress 32GB SST files cost about 20min,
+// we set the upper bound to 32GB to avoid too long compression time.
+// factor is the kv count per row.
```
lance6716, line 1337: should it be "number of indices + 1"? 😵

pkg/lightning/backend/local/local.go
```diff
@@ -2090,150 +2599,262 @@ const (
...
 		return errors.Trace(err)
 	}
-	sw.writer = sstable.NewWriter(f, sstable.WriterOptions{
-		TablePropertyCollectors: []func() pebble.TablePropertyCollector{
-			newRangePropertiesCollector,
-		},
-		BlockSize: 16 * 1024,
-	})
-	return nil
+	return i.reader.Close()
```
lance6716, line 2669:
```suggestion
	return errors.Trace(i.reader.Close())
```

pkg/lightning/backend/local/local.go
```diff
@@ -1942,142 +2486,107 @@ func (local *local) EngineFileSizes() (res []backend.EngineFileSize) {
...
+		return errors.Trace(err)
 	}
-	defer immWriter.cleanUp()
-
-	if err = immWriter.writeKVs(&w.writeBatch); err != nil {
-		return err
+	sort.Slice(w.writeBatch[:w.batchCount], func(i, j int) bool {
+		return bytes.Compare(w.writeBatch[i].Key, w.writeBatch[j].Key) < 0
+	})
+	writer.minKey = append(writer.minKey[:0], w.writeBatch[0].Key...)
```
lance6716, line 2561: seems we can move `writer.minKey` into `writeKVs`, just as `maxKey`

pkg/lightning/backend/local/local.go
```diff
@@ -1942,142 +2486,107 @@ func (local *local) EngineFileSizes() (res []backend.EngineFileSize) {
...
+	err = w.local.addSST(ctx, meta)
+	if err != nil {
+		return errors.Trace(err)
 	}
 
-	return immWriter.ingestInto(w.local, desc|localIngestDescriptionImmediate)
+	w.totalSize += w.batchSize
+	w.batchSize = 0
+	w.batchCount = 0
+	w.kvBuffer.reset()
```
lance6716, line 2578: did `w.writeBatch` need to be reset here?

pkg/lightning/backend/local/local.go
glorv: No. `appendRowsSorted` will directly call `writeKVs` multi times, but only the first call should set the `minKey`

pkg/lightning/backend/local/local.go
glorv: No. we can reuse all the `KvPair` in w.writeBatch, this can slightly help optimize the memory allocation. See function `appendRowsUnsorted`

pkg/lightning/backend/local/local.go
```diff
@@ -139,11 +131,19 @@ type importMutexState uint32
 
 const (
 	importMutexStateImport importMutexState = 1 << iota
-	importMutexStateFlush
 	importMutexStateClose
-	importMutexStateLocalIngest
+	// importMutexStateReadLock is a special state because in this state we lock engine with read lock
+	// and add isImportingAtomic with this value. In other state, we directly store with the state value.
+	// so this must always the last value of this enum.
+	importMutexStateReadLock
```
lance6716, line 138: don't know if we should implement a lock with these many states

such as `type lock struct {mutex, state}`, use mutex to protect the reading/changing of state. and if the current state and requesting state is not tolerable, block the invoker and add it to the waiting queue.

(reinventing the wheels 😵 )

pkg/lightning/backend/local/local.go
lance6716: And I weakly advise to take a look at this, only read from `e.sstMetasChan` in one place @glorv

lance6716 (approved):

pkg/lightning/backend/local/local.go
glorv: Actually, there are only two locked states, write lock for close/state, and read lock for flush.  This enum is merely used for debugging convenience. The `isImporting` atomic is using to simulate `tryLock` mechanism to avoid blocking? 🤔

pkg/lightning/backend/local/local.go
glorv: ok. I'll take a look at this.😂

tests/lightning_disk_quota/run.sh
```diff
@@ -45,7 +45,8 @@ for i in $(seq 3); do
     fi
 done
 
-while [ ! -e "$FINISHED_FILE" ]; do
+while [ ! -e "$FINISHED_FILE" ] && [ -e "$DISK_QUOTA_DIR" ]; do
+    // du may fail because the directory is removed by lightning
```
kennytm, line 49:
```suggestion
    # du may fail because the directory is removed by lightning
```

pkg/lightning/backend/local/local.go
kennytm: isLocked is used to determine if an engine has already been importing to TiKV when computing disk quota. It should return false when TiKV is not involved.

kennytm:
Huh.

pkg/lightning/backend/local/local.go
```diff
@@ -434,7 +820,7 @@ func NewLocalBackend(
 	g glue.Glue,
 	maxOpenFiles int,
 ) (backend.Backend, error) {
-	localFile := cfg.SortedKVDir
+	File := cfg.SortedKVDir
```
kennytm, line 823: why rename the variable. as least use a lower case name.

pkg/lightning/backend/local/local.go
```diff
@@ -628,27 +1036,62 @@ func (local *local) openEngineDB(engineUUID uuid.UUID, readOnly bool) (*pebble.D
 		// set threshold to half of the max open files to avoid trigger compaction
 		L0CompactionThreshold: math.MaxInt32,
 		L0StopWritesThreshold: math.MaxInt32,
+		LBaseMaxBytes:         16 << 40,
```
kennytm, line 1039:
```suggestion
		LBaseMaxBytes:         16 * units.TiB,
```

pkg/lightning/backend/local/local.go
```diff
@@ -628,27 +1036,62 @@ func (local *local) openEngineDB(engineUUID uuid.UUID, readOnly bool) (*pebble.D
...
 		ReadOnly:              readOnly,
 		TablePropertyCollectors: []func() pebble.TablePropertyCollector{
 			newRangePropertiesCollector,
 		},
 	}
+	// set level target file size to avoid pebble auto triggering compaction that split ingest SST files into small SST.
+	opt.Levels = []pebble.LevelOptions{
+		{
+			// 16GB
+			TargetFileSize: 16 << 30,
```
kennytm, line 1051:
```suggestion
			TargetFileSize: 16 * units.GiB,
```

pkg/lightning/backend/local/local.go
```diff
@@ -2090,150 +2588,263 @@ const (
...
+		}
+		reader, err := sstable.NewReader(f, sstable.ReaderOptions{})
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		iter, err := reader.NewIter([]byte{'t'}, []byte{'u'})
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		key, val := iter.SeekGE([]byte{'t'})
```
kennytm, line 2752: 🤔 seems unnecessary to restrict the range. we don't store non-`t` prefixed stuff in the SST.

pkg/lightning/restore/restore.go
```diff
@@ -80,6 +80,11 @@ const (
 	compactStateDoing
 )
 
+const (
+	compactionLowerThreshold = 512 << 20 // 512M
+	compactionUpperThreshold = 32 << 30  // 32GB
```
kennytm, line 85:
```suggestion
	compactionLowerThreshold = 512 * units.MiB
	compactionUpperThreshold = 32 * units.GiB
```

kennytm (commented):
rest LGTM

glorv:
> Huh.

I add the `require-LGT3` label for this PR since the change is quite big. Please approve this PR, then we can merge it. 😅

kennytm:
@glorv see above 😅

pkg/lightning/backend/local/local.go
glorv: This must be done by the IDE when I refactor the `LocalFile` struct..

kennytm: 1-fs8

pkg/lightning/backend/local/local.go
kennytm: @glorv (line 2748 is included)

pkg/lightning/backend/local/local.go
glorv: 😂😂😂

pkg/lightning/backend/local/local.go
glorv: I supposed the `NewIter` must receive a not empty bound, but found we can just pass `nil, nil` instead. 😂

PR #759, opened 2021-02-26
build: enable cgo by default

glorv:
Enable cgo by default because tidb-lightning depends on CGO to gain better performance。 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 -


glorv:
Due to https://github.com/cockroachdb/pebble/commit/c2b05f12d7f69ff37d3deb64381ffed161c24aa8 this bug, if built CGO is not enabled, lightning is likely to panic. Though we can update pebble to a newer version to avoid this bug, but enable cgo can gain better performance.

overvenus:
LGTM

ti-srebot (approved):

tests/br_table_partition/run.sh
```diff
@@ -19,7 +19,7 @@ TABLE="usertable"
 TABLE_COUNT=16
 PATH="tests/$TEST_NAME:bin:$PATH"
 
-run_sql "set @@global.tidb_enable_table_partition ='nightly'"
+run_sql "set @@global.tidb_enable_table_partition ='nightly'" || true
```
overvenus, line 22:
```suggestion
run_sql "set @@global.tidb_enable_list_partition = ON;"
```

glorv: `tidb_enable_list_partition` is only support in session scope. Must put it in each run_sql so that it can take effect.

```
mysql> set @@global.tidb_enable_list_partition = 'ON';
ERROR 1105 (HY000): Variable 'tidb_enable_list_partition' is a SESSION variable and can't be used with SET GLOBAL
```

tests/br_table_partition/run.sh
```diff
@@ -19,11 +19,14 @@ TABLE="usertable"
 TABLE_COUNT=16
 PATH="tests/$TEST_NAME:bin:$PATH"
 
-run_sql "set @@global.tidb_enable_table_partition ='nightly'"
+# `tidb_enable_list_partition` currently only support session level variable, so we must put it in the create table sql
+SESSION_VARIABLE=
+run_sql "set @@global.tidb_enable_table_partition ='nightly'" || SESSION_VARIABLE="set @@session.tidb_enable_list_partition = 'ON'; "
```
kennytm, line 24: this is only used inside prepare.sh... so it seems better this definition be placed just inside prepare.sh

kennytm (commented):
rest lgtm

tests/br_table_partition/run.sh
glorv: done

kennytm (commented):
LGTM

PR #766, opened 2021-03-01
lightning: Fix lints for lightning

YuJuncen:
CI for `tidb-lightning` doesn't include lints, after merged into this repository, we hope it can pass the CI tests too.

Refactoring following what the linter says.

### NOTEs
Those linters are disabled:
- errorlint, because pingcap/errors is incompatible with std errors.(see below)
Standard `errors` uses `err.Unwrap()` to get cause of errors. But `pingcap/errors` uses `err.Cause()` and `errors.Unwrap(err)`. If we follow this linter, using `errors.Is` to check error class, causes of errors would be omitted.
- wrapcheck, because there are too many unwrapped errors in tidb-lightning. An option is using a script to wrap them with `errors.Trace(err)`, which is tricky, and in someway not really wrapped those errors. Maybe we need farther refactoring for error returning.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

### Release Note

 - No release note


pkg/lightning/backend/backend_test.go
```diff
@@ -273,6 +273,7 @@ func (s *backendSuite) TestImportFailedRecovered(c *C) {
 	c.Assert(err, IsNil)
 }
 
+//nolint:interfacer // change test case signature might cause Check failed to find this function?
```
glorv, line 276: This should be a false positive case, the comment should be removed.

`check`'s finding rule is the same with go-test, only need to func name start with `Test`

pkg/lightning/backend/session.go
```diff
@@ -200,11 +198,18 @@ func newSession(options *SessionOptions) *session {
 	vars.SQLMode = sqlMode
 	if options.SysVars != nil {
 		for k, v := range options.SysVars {
-			vars.SetSystemVar(k, v)
+			if err := vars.SetSystemVar(k, v); err != nil {
+				log.L().Warn("new session: failed to set system var",
```
glorv, line 202:
```suggestion
				log.L().DPanic("new session: failed to set system var",
```

pkg/lightning/backend/tidb.go
```diff
@@ -87,7 +87,10 @@ func NewTiDBBackend(db *sql.DB, onDuplicate string) Backend {
 
 func (row tidbRow) ClassifyAndAppend(data *Rows, checksum *verification.KVChecksum, _ *Rows, _ *verification.KVChecksum) {
 	rows := (*data).(tidbRows)
-	*data = tidbRows(append(rows, row))
+	// I'm not sure if `rows := data.(*tidbRows); *rows = append(*rows, row)` could solve this lint.
+	// Seems there are some stories. Leave it untouched.
+	//nolint:gocritic
+	*data = append(rows, row)
```
glorv, line 93: I think we can make the two slices as return parameters to avoid this lint but I'm not sure if it's necessary.

pkg/lightning/mydump/region.go
```diff
@@ -301,7 +298,8 @@ func makeParquetFileRegion(
 	if err != nil {
 		return prevRowIdxMax, nil, errors.Trace(err)
 	}
-	numberRows, err := ReadParquetFileRowCount(ctx, store, r, dataFile.FileMeta.Path)
+	// TODO: handle the error here?
+	numberRows, _ := ReadParquetFileRowCount(ctx, store, r, dataFile.FileMeta.Path)
```
glorv, line 302:
```suggestion
	numberRows, err := ReadParquetFileRowCount(ctx, store, r, dataFile.FileMeta.Path)
	if err != nil {
		return prevRowIdxMax, nil, errors.Trace(err)
	}
```

pkg/lightning/restore/restore.go
```diff
@@ -801,42 +801,48 @@ func (rc *RestoreController) listenCheckpointUpdates() {
...
 		failpoint.Inject("FailIfImportedChunk", func(val failpoint.Value) {
-			if merger, ok := scp.merger.(*ChunkCheckpointMerger); ok && merger.Checksum.SumKVS() >= uint64(val.(int)) {
+			if merger, ok := scp.merger.(*checkpoints.ChunkCheckpointMerger); ok && merger.Checksum.SumKVS() >= uint64(val.(int)) {
 				rc.checkpointsWg.Done()
 				rc.checkpointsWg.Wait()
 				panic("forcing failure due to FailIfImportedChunk")
 			}
 		})
 
+		//nolint:scopelint // This would be either INLINED or ERASED, believe me!
```
glorv, line 813:
```suggestion
		//nolint:scopelint // This would be either INLINED or ERASED at compile time!
```

glorv (commented):
rest LGTM

pkg/lightning/backend/tidb.go
YuJuncen: I guess TiDB requires the `Row` interface too(due to log restore and BR in SQL). If that is real, I think when we want to refactor that, we'd better to make another PR.

pkg/lightning/backend/backend_test.go
YuJuncen:
```
[2021-03-09T02:40:29.984Z] ----------------------------------------------------------------------
[2021-03-09T02:40:29.984Z] PANIC: backend_test.go:276: backendSuite.TestClose
[2021-03-09T02:40:29.984Z] 
[2021-03-09T02:40:29.984Z] ... Panic: backendSuite.TestClose argument should be *check.C
[2021-03-09T02:40:29.984Z] OOPS: 53 passed, 1 PANICKED
[2021-03-09T02:40:29.984Z] --- FAIL: TestKV (0.60s)
```

Oops...

YuJuncen:
@overvenus PTAL if convenient

pkg/lightning/mydump/region.go
```diff
@@ -302,6 +299,9 @@ func makeParquetFileRegion(
 		return prevRowIdxMax, nil, errors.Trace(err)
 	}
 	numberRows, err := ReadParquetFileRowCount(ctx, store, r, dataFile.FileMeta.Path)
+	if err != nil {
+		return 0, nil, errors.Trace(err)
+	}
```
overvenus, line 304: Nice catch!

pkg/lightning/mydump/parser.go
```diff
@@ -527,6 +527,9 @@ func (parser *blockParser) LastRow() Row {
 
 // RecycleRow places the row object back into the allocation pool.
 func (parser *blockParser) RecycleRow(row Row) {
+	// We need farther benchmarking to make sure
+	//   whether send a pointer(instead of a slice) here can improve performance.
```
overvenus, line 531:
```suggestion
	// We need farther benchmarking to make sure whether send a pointer
	// (instead of a slice) here can improve performance.
```

pkg/lightning/log/log.go
```diff
@@ -75,7 +75,9 @@ var (
 
 // InitLogger initializes Lightning's and also the TiDB library's loggers.
 func InitLogger(cfg *Config, tidbLoglevel string) error {
-	logutil.InitLogger(&logutil.LogConfig{Config: pclog.Config{Level: tidbLoglevel}})
+	if err := logutil.InitLogger(&logutil.LogConfig{Config: pclog.Config{Level: tidbLoglevel}}); err != nil {
+		return err
```
overvenus, line 79:
```suggestion
		return errors.Trace(err)
```

overvenus (commented):
Left few nits, rest LGTM

pkg/lightning/checkpoints/checkpoints.go
```diff
@@ -527,7 +533,11 @@ func IsCheckpointsDBExists(ctx context.Context, cfg *config.Config) (bool, error
 			return false, errors.Trace(err)
 		}
 		defer rows.Close()
-		return rows.Next(), nil
+		result := rows.Next()
+		if err := rows.Err(); err != nil {
+			return false, err
```
glorv, line 538:
```suggestion
			return false, errors.Trace(err)
```

glorv (commented):
rest LGTM

PR #771, opened 2021-03-02
pkg/*/transaction:implement functions for tableinfo

AndreMouche:
Signed-off-by: shirly <AndreMouche@126.com>

The kv client(in `tidb`) should not process information about table info, we would like to move the table info into `kv.Transaction`(which is temporary too，we will fix it in the future)

### Check List 

Tests 

### Release Note

 - No release note


disksing (approved):

pkg/kv/session.go
```diff
@@ -180,6 +180,13 @@ func (t *transaction) GetUnionStore() kv.UnionStore {
 	return &t.kvUnionStore
 }
 
+func (t *transaction) GetTableInfo(id int64) *model.TableInfo {
```
overvenus, line 183: Please add a comment explain why it needs to implement `GetTableInfo`, thanks!

overvenus (commented):
Rest LGTm

overvenus:
LGTM

ti-srebot (approved):

PR #784, opened 2021-03-03
lightning: support restore data into tables that contains data

glorv:
- Support use lightning local/importer backend to load data into tables that already contain data.
- Support running multiple lightning in parallel to load data into tables that already contain data.

- Add a new metatable `mysql.brie_sub_tasks` in TiDB to store the meta data for each table restore task. 
- Use `show table next_row_id` for the auto_id allocator base value. This can avoid auto_id conflict with the existing data. If there are other task record for the same table in brie_sub_tasks, the rowid rang will be calculated based on the recored info. 
- Prefetch the checksum value and store it in the table checkpoint. Also add this value into local checksum when compares local vs remote checksum.
- Only do checksum and analyze after the last parallel task is finished, other tasks only put their `local checksum` info into the metatable, so the last only actually do the checksum can sum all the local checksum to get the right result.

Alternatives:
- It's not necessary to put the metatable into `mysql` db. We can create a temporary table like the checkpoints db on-demand and drop it after all lightning tasks are finished. 
  - Pros: avoid ask the `admin` permisssion.
- Store the metadata at `pd` instead of in TiDB. 
  - pro: maybe easier to manage. 
  - Drawback: this will only support newer cluster.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release Note

 - Support load data into tables that contain data.


glorv:
Need to update the `Glue` interface to support scan or transaction to better support the `show table next_row_id` result.

july2993:
- Support use lightning local/importer backend to load data into tables that already contain data.

if it failed to import how to clean up the imported data?

pkg/lightning/restore/restore.go
```diff
@@ -1291,16 +1442,76 @@ func (tr *TableRestore) restoreTable(
...
+	}
+
 	// no need to do anything if the chunks are already populated
 	if len(cp.Engines) > 0 {
 		tr.logger.Info("reusing engines and files info from checkpoint",
 			zap.Int("enginesCnt", len(cp.Engines)),
 			zap.Int("filesCnt", cp.CountChunks()),
 		)
 	} else if cp.Status < checkpoints.CheckpointStatusAllWritten {
+		versionStr, err := rc.tidbGlue.GetSQLExecutor().ObtainStringWithLog(
```
Little-Wallace, line 1465: Could we add some method `GetTiDBVersion()` ?

Little-Wallace (approved):
LGTM

pkg/lightning/restore/restore.go
```diff
@@ -2733,3 +2933,789 @@ func (cr *chunkRestore) restore(
...
+	if err := exec.QueryRow(ctx, "fetch table meta row count", query, &cnt); err != nil {
+		return errors.Trace(err)
+	}
+	if cnt > 0 {
+		log.L().Warn("there are unfinished table in table meta table, cleanup skipped.")
+		return nil
+	}
+
+	// avoid override existing metadata if the meta is already inserted.
+	stmt := fmt.Sprintf("DROP DATABASE %s;", m.schemaName)
```
kennytm, line 3652: (ditto)

pkg/lightning/restore/restore.go
```diff
@@ -2733,3 +2933,789 @@ func (cr *chunkRestore) restore(
 		return ctx.Err()
 	}
 }
+
+type metaMgrBuilder interface {
```
kennytm, line 2937: let's move these meta-manager stuff into its own `.go` file to avoid further bloating the `restore.go` file (it's over 3700 lines after this PR 😅)

kennytm (commented):
rest LGTM

kennytm (approved):

PR #798, opened 2021-03-04
*: unify version and set max compatible via NextMajorVersion

overvenus:

Unify version and set max compatible via NextMajorVersion.

### Check List 

Tests 

 - Unit test

Code changes

Related changes

### Release Note

 - No release note.


errors.toml
```diff
@@ -1,187 +1,187 @@
 # AUTOGENERATED BY github.com/pingcap/errors/errdoc-gen
 # YOU CAN CHANGE THE 'description'/'workaround' FIELDS IF THEM ARE IMPROPER.
 
-["BR:Backup:ErrBackupChecksumMismatch"]
+["BRIE:Backup:ErrBackupChecksumMismatch"]
```
kennytm, line 4: 🤔 are error codes supposed to be stable?

kennytm (commented):
Rest LGTM

errors.toml
overvenus: For compatibility, I'll revert the change.

pkg/pdutil/pd.go
```diff
@@ -607,3 +608,14 @@ func (p *PdController) Close() {
...
+
+// FetchPDVersion get pd version
+func FetchPDVersion(ctx context.Context, tls *common.TLS, pdAddr string) (*semver.Version, error) {
+	var rawVersion string
+	err := tls.WithHost(pdAddr).GetJSON(ctx, "/pd/api/v1/config/cluster-version", &rawVersion)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	return semver.NewVersion(rawVersion)
```
glorv, line 620: The raw version string fetch from pd is quoted, I think we need to trim the `"` here.
```
> curl http://172.16.4.5:2379/pd/api/v1/config/cluster-version

"4.0.8"
```

glorv (commented):
rest LGTM

ti-srebot (approved):

PR #761, opened 2021-02-26
lightning/backend: run batch split region in parallel for local backend

glorv:
### What problem does this PR solve? 
Run batch split region in parallel to speed up batch split region.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

Side effects

 - Possible performance regression
 - Increased code complexity

Related changes

 - Need to cherry-pick to the release branch

### Release Note

 - Run batch split regions in parallel for lightning local backend

PR #815, opened 2021-03-05
lightning: fix several disk-quota-related bugs

### LINKED ISSUE #812: Lightning Disk Quota is not fully respected after restart due to non-deterministic engine selection

### 1. What did you do?

1. Prepare a 1 TB data dump, 
2. Set batch size to 100 GB (so it is split into ~10 batches),
3. Set disk quota to 200 GB,
4. Set engine-concurrency to 6,
5. Run Lightning,
6. After 50 minutes (i.e. after the first disk-quota import has triggered), kill -9 Lightning.
7. Run Lightning again,

### 2. What did you expect to see?

Import continues with no errors.

### 3. What did you see instead?

After 20 minutes, Lightning crashed with "no space left on device" error.

```
[2021/03/05 09:24:42.639 +00:00] [ERROR] [restore.go:315] ["the whole procedure failed"] [takeTime=22m59.878871993s] [error="restore table `dq`.`g` failed: write /data1/data/lightning/kv/9ebefdf0-dc5c-4e88-b15a-0169065dcdaa.sst: no space left on device"]
```

### 4. What version of BR and TiDB/TiKV/PD are you using?

Lightning: nightly = f1daf8a4d6d14dcd2ce52766ae9428c435d33616,
Cluster: v4.0.11

### Misc 

This bug happens because the list of engines opened by the first and second invocation are different. So even if the disk usage from the currently active engine did not exceed 200 GB, the actual total counting the previous run already exceeds it.

kennytm:

1. Fix #812
2. Fix the bug where the checkpoints in local back-end are not advanced even after disk quota force-flushed all engines.

1. For #812, it is caused by iterating `cp.Engines` in `restoreEngines()`, which is a hash map. We read the map into a sorted array and iterate that instead, so engines are always restored in deterministic order.
2. For the checkpoint issue, we now update the checkpoint based on the Local Writer's synchronization status instead of the disk quota status (the deliverLoop cannot see `diskQuotaStateImporting` during `diskQuotaLock.RLock()`). *This assumes Pebble's `Ingest()` will persist all content to disk*.

### Check List 

Tests 

 - Integration test (TODO)
    * Testing its effect in the 1T GA test.

Code changes

Side effects

Related changes

### Release Note

 - No release note.

### DIFF
--- a/pkg/lightning/backend/backend.go
+++ b/pkg/lightning/backend/backend.go
@@ -163,7 +163,7 @@ type AbstractBackend interface {
 	ResetEngine(ctx context.Context, engineUUID uuid.UUID) error
 
 	// LocalWriter obtains a thread-local EngineWriter for writing rows into the given engine.
-	LocalWriter(ctx context.Context, engineUUID uuid.UUID, maxCacheSize int64) (EngineWriter, error)
+	LocalWriter(ctx context.Context, engineUUID uuid.UUID) (EngineWriter, error)
 }
 
 func fetchRemoteTableModelsFromTLS(ctx context.Context, tls *common.TLS, schema string) ([]*model.TableInfo, error) {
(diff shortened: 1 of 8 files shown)


kennytm:
Separated into #761, #845, #846, #847, closing.

PR #824, opened 2021-03-08
br: Add TLS support for PProf of BR

### LINKED ISSUE #817: Status address is not protected after enabling mTLS
During merging BR and TiDB-Lightning, we find BR status address is not protected after enabling mTLS. 

There is a TODO left, we need to implement it.

https://github.com/pingcap/br/blob/8b9b626a22e471eb2c39546cf9156be18fd10168/cmd/br/cmd.go#L153-L164

YuJuncen:
Fixed #817 

Get TLS config and apply it when starting proof.

### Check List 

Tests 

 - Integration test

### Release Note

 - Fixed a bug that caused, even BR started with TLS config, the pprof endpoints won't be served with TLS.

### DIFF
--- a/cmd/br/cmd.go
+++ b/cmd/br/cmd.go
@@ -158,12 +158,12 @@ func Init(cmd *cobra.Command) (err error) {
 		}
 		tlsConfig := task.TLSConfig{}
 		useTLS := !(tlsConfig.ParseFromFlags(cmd.Flags()) != nil || tlsConfig.Cert == "" || tlsConfig.Key == "")
-		startPProf := func() { utils.StartPProfListener(statusAddr) }
+		startPProf := func(addr string) { utils.StartPProfListener(addr) }
 		if useTLS {
-			startPProf = func() { utils.StartPProfListenerTLS(statusAddr, tlsConfig.Cert, tlsConfig.Key) }
+			startPProf = func(addr string) { utils.StartPProfListenerTLS(addr, tlsConfig.Cert, tlsConfig.Key) }
 		}
 		if statusAddr != "" {
-			startPProf()
+			startPProf(statusAddr)
 		} else {
 			utils.StartDynamicPProfListener(startPProf)
 		}
--- a/cmd/br/cmd.go
+++ b/cmd/br/cmd.go
@@ -167,6 +167,7 @@ func startPProf(cmd *cobra.Command) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
+	// Host isn't used here.
 	tls, err := common.NewTLS(ca, cert, key, "localhost")
 	if err != nil {
 		return errors.Trace(err)
--- a/cmd/br/cmd.go
+++ b/cmd/br/cmd.go
@@ -11,7 +11,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/pingcap/br/pkg/lightning/common"
+	tidbutils "github.com/pingcap/tidb-tools/pkg/utils"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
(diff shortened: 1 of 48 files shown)


pkg/utils/pprof.go
```diff
@@ -39,19 +38,40 @@ func StartPProfListener(statusAddr string) {
...
+		go func() {
+			if e := http.Serve(listener, nil); e != nil {
+				log.Warn("failed to serve pprof", zap.String("addr", startedPProf), zap.Error(e))
+				mu.Lock()
+				startedPProf = ""
+				mu.Unlock()
+				return
+			}
+		}()
+	}
```
overvenus, line 76: They are very similar, could you unify them?

pkg/utils/dyn_pprof_unix.go
```diff
@@ -17,16 +17,16 @@ const startPProfSignal = syscall.SIGUSR1
 var signalChan = make(chan os.Signal, 1)
 
 // StartDynamicPProfListener starts the listener that will enable pprof when received `startPProfSignal`.
-func StartDynamicPProfListener() {
+func StartDynamicPProfListener(start func(addr string)) {
```
overvenus, line 20: Prefer pass arguments explicitly instead of a closure.

YuJuncen: This function is an internal function, I think if we pass arguments here, we probably need to provide TLS config explicitly here, like `StartDynamicPProfListener(cert, key string)`. At the implementation might would be like:

```go
if cert == "" || key == "" {
  StartPProfListener("0.0.0.0:0")
} else {
  StartPProfListener("0.0.0.0:0", cert, key)
}
```

We got more args, and function overloading-like magic... Without more information, I'm afraid I would prefer the current style...

3pointer (commented):
/lgtm

pkg/utils/dyn_pprof_unix.go
overvenus: We can move the if condition into the `StartPProfListener`.

```go
func StartPProfListener(addr, cert, key string) error {
    if cert == "" || key == "" {
    } else {
    }
}
```

YuJuncen: Yep, I think those would make a `if` statement within the function, whose condition is the new parameter we added. And in my opinion this is also control coupling or another form of function overloading. 

I think make another function and let user to decide how to start the pprof listener would be more flexible. 

I'm wondering if there are some style guides preferring plain arguments or some bad things happened with closures...? If so, I will surely remove those.

YuJuncen:
@kennytm Seems there is strong coupling between this method and lightning... I'm afraid reuse it needs more refactor...

kennytm:
i mean using this to handle TLS instead of `if`/`else`.

```go
 	listener, err := net.Listen("tcp", statusAddr) 
        ...
 	listener = l.globalTLS.WrapListener(listener) 

 	go func() { 
 		err := l.server.Serve(listener) 
 		log.L().Info("stopped HTTP server", log.ShortError(err)) 
 	}()        
```

cmd/br/cmd.go
```diff
@@ -11,6 +11,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/pingcap/br/pkg/lightning/common"
```
kennytm, line 14: we should use [`github.com/pingcap/tidb-tools/pkg/utils`](https://github.com/pingcap/tidb-tools/blob/master/pkg/utils/security.go), or at least move this package out of Lightning.

YuJuncen: done

kennytm:
LGTM

ti-srebot (approved):

PR #837, opened 2021-03-10
pkg/lightning:move mock file for lightning into lightning package

AndreMouche:
Signed-off-by: shirly <AndreMouche@126.com>

move mock file for lightning into lightning package,
After this PR merged, tidb needn't import github.com/cockroachdb/pebble indirectly.

related PR:https://github.com/pingcap/br/pull/786

### Check List 

Tests 

 - Unit test
 - Integration test

### Release Note

 - No release note


overvenus (changes requested):
Please do not split mock package.

AndreMouche:
why? I think it is only used for lighting, so we could and should split it. If we should merge them together, so why we need to have an independent package for lighting?

> Please do not split mock package.

why? I think it is only used for lighting, so we could and should split it. If we should merge them together, so why we need to have an independent package for lighting?

kennytm:
This PR is caused by pingcap/tidb#23183.

The issue is like this:

1. After merging Lightning into BR, BR as a module now depends on Pebble.
2. Pebble depends on [`github.com/cockroachdb/errors`](https://github.com/cockroachdb/pebble/blob/2673cbf29137abbb8379c72020cc2a9f009eec57/go.mod#L6)
3. `github.com/cockroachdb/errors` depends on [Protobuf 1.4.2](https://github.com/cockroachdb/errors/blob/4c1636886a764323efed240d738357ea251908bb/go.mod#L12)
4. But TiDB currently requires Protobuf at `=1.3.4` due to some maintenance issue or something.

Current solution in 23183 is to do `replace protobuf => protobuf 1.3.4`. Now trying an alternate solution of making the mocks test-only.

PR #845, opened 2021-03-10
lightning/restore: fix #812, ensure engines are restored in fixed order

### LINKED ISSUE #812: Lightning Disk Quota is not fully respected after restart due to non-deterministic engine selection

kennytm:

Fix #812

(Split off from #815). 

#812 is caused by iterating `cp.Engines` in `restoreEngines()`, which is a hash map. We read the map into a sorted array and iterate that instead, so engines are always restored in deterministic order.

### Check List 

Tests 

 - Integration test
 - Manual test
     * Import a 1 TB table with checkpoint
     * Kill Lightning
     * Restart Lightning
     * Note that the same engines (0, 1, 2, ...) are used in both steps 1 and 3, instead of randomly chosen ones.

Code changes

Side effects

Related changes

    * release-4.0

### Release Note

 - When resuming from checkpoint, Lightning now keeps restoring the original engines instead of starting to process some random new ones (#815).


tests/lightning_checkpoint_engines/run.sh
```diff
@@ -81,13 +81,17 @@ for BACKEND in importer local; do
...
-  set +e
   for i in $(seq "$ENGINE_COUNT"); do
       echo "******** Importing Table Now (step $i/$ENGINE_COUNT) ********"
+      set +e
       do_run_lightning $BACKEND mysql 2> /dev/null
       [ $? -ne 0 ] || exit 1
+      set -e
+      # make sure the data engines are processed in the engine_id order.
+      # there is 75% this test failing if they are processed randomly.
+      run_sql 'SELECT *, status <= prev_engine_status AS ordering FROM (SELECT table_name, engine_id, status, lag(status) OVER (PARTITION BY table_name ORDER BY engine_id) AS prev_engine_status FROM tidb_lightning_checkpoint.engine_v5 WHERE engine_id >= 0) _'
```
glorv, line 92: So I think this check should be `status < prev_engine_status` or add another restriction that `status >= 120`

kennytm: changed to a more dedicated test case.

glorv (commented):
rest LGTM

glorv (commented):
/lgtm

ti-srebot (approved):

PR #846, opened 2021-03-10
lightning/restore: save checkpoint as soon as local writers are sync

kennytm:

(Split off from #815).

We now update the checkpoint based on the Local Writer's synchronization status instead of the disk quota status (the `deliverLoop` cannot see `diskQuotaStateImporting` during `diskQuotaLock.RLock()`). *This assumes Pebble's `Ingest()` will persist all content to disk.*

### Check List 

Tests 

 - Integration test (TODO)

Code changes

Side effects

Related changes

### Release Note

 - No release note

### DIFF
--- a/pkg/lightning/backend/local.go
+++ b/pkg/lightning/backend/local.go
@@ -1771,6 +1771,7 @@ type LocalWriter struct {
 	memtableSizeLimit int64
 	writeBatch        kvMemCache
 	writer            *sstWriter
+	isDirty           atomic.Bool
 }
 
 func (w *LocalWriter) AppendRows(ctx context.Context, tableName string, columnNames []string, ts uint64, rows Rows) error {
@@ -1781,9 +1782,14 @@ func (w *LocalWriter) AppendRows(ctx context.Context, tableName string, columnNa
 	if err := w.writeErr.Get(); err != nil {
 		return err
 	}
+	w.isDirty.Store(true)
 	w.kvsChan <- kvs
 	w.local.Ts = ts
 	return nil
+}
+
+func (w *LocalWriter) IsSynchronized() bool {
+	return !w.isDirty.Load()
 }
 
 func (w *LocalWriter) Close() error {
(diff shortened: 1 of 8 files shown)


pkg/lightning/backend/local.go
```diff
@@ -1830,34 +1882,38 @@ outside:
 		select {
 		case kvs, ok := <-w.kvsChan:
 			if !ok {
+				w.markCleanQueue()
```
glorv, line 1885: Since all kvs are consumed, the current state must be either 0 or 1, do we really need this action? I think it's redundant.

pkg/lightning/backend/local.go
```diff
@@ -1830,34 +1882,38 @@ outside:
...
+			if w.writeBatch.totalSize > w.memtableSizeLimit {
+				if w.writer == nil {
+					w.writer, err = newSSTWriter(w.genSSTPath())
+					if err != nil {
+						w.writeErr.Set(err)
+						return
+					}
+				}
+
+				if err = w.writeKVsOrIngest(0); err != nil {
```
glorv, line 1901: What if we do `w.markCleanFlush()` after this ingest and the ingest  below. If after this ingest, there is no data in the channel, this should be a consistent state that can do save the checkpoint.

glorv (commented):
I think this approach is correct but a bit complex.

pkg/lightning/backend/local.go
kennytm: > this should be a consistent state that can do save the checkpoint.

that's impossible. `writeKVsOrIngest` does one of the two things:

1. move the content of `w.writeBatch` into `w.writer` → so `w.writeBatch` will become empty but `w.writer` has content, or
2. directly ingest `w.writeBatch` into Pebble because it is not sorted after `w.writer` → a necessary condition for this is `w.writer` not being empty, thus the LocalWriter is still dirty.

glorv: What about `writeKVsOrIngest ` return a bool to represent the flush state, if it triggers a ingest, then we can do a `markCleanFlush`.

kennytm: @glorv both cases 1 and 2 are dirty.

glorv: I think there is no conflict if we do `w.dirtyState.Add(-2 * flushedKvCount)`. 
In case 1, move content into w.writer, this won't change dirtyState, so It's still dirty.
In case 2, the decrease won't contain the kvs that still in w.writer so it's still dirty too.

A clean state can be defined as `All kvs send by AppendRows are by ingested`, so if we only decrease the count after ingest by the ingested kv count, we can make sure if the count decrease to 0, the state must be clean.

gozssky:
It seems this pr can be replaced by https://github.com/pingcap/br/pull/1080.

PR #847, opened 2021-03-10
lightning: reduce memory use

kennytm:

Lightning's Local Backend may OOM because some cache are too aggressive.

(Split off from #815).

1. Make the engine's mem table size configurable (keep default at 512 MiB)
2. Make the local writer's mem table size configurable (reduce default from 512 MiB to 128 MiB)
3. Reduced capacity of `LocalWriter.kvsChan` from 1024 to 16. Originally this will make Lightning retain 1024 × (region-conc) × (encoded-kvs) in the channel buffer, which can hold over 3 GiB.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release Note

 - Added configurations `tikv-importer.engine-mem-cache-size` and `tikv-importer.local-writer-mem-cache-size` to tune between memory usage and performance.

### DIFF
--- a/pkg/lightning/config/config.go
+++ b/pkg/lightning/config/config.go
@@ -614,9 +614,6 @@ func (cfg *Config) Adjust(ctx context.Context) error {
 			}
 			cfg.TikvImporter.DiskQuota = ByteSize(storageSize.Available - reservedSize)
 		}
-		if cfg.TikvImporter.DiskQuota < cfg.TikvImporter.LocalWriterMemCacheSize {
-			cfg.TikvImporter.LocalWriterMemCacheSize = cfg.TikvImporter.DiskQuota
-		}
 	}
 
 	if cfg.TikvImporter.Backend == BackendTiDB {
--- a/pkg/lightning/config/config.go
+++ b/pkg/lightning/config/config.go
@@ -70,19 +70,9 @@ const (
 	defaultBuildStatsConcurrency      = 20
 	defaultIndexSerialScanConcurrency = 20
 	defaultChecksumTableConcurrency   = 2
-)
 
-const (
-	LocalMemoryTableSize = 512 * units.MiB
-
-	// autoDiskQuotaLocalReservedSize is the estimated size a local-backend
-	// engine may gain after calling Flush(). This is currently defined by its
-	// max MemTable size (512 MiB). It is used to compensate for the soft limit
-	// of the disk quota against the hard limit of the disk free space.
-	//
-	// With a maximum of 8 engines, this should contribute 4.0 GiB to the
-	// reserved size.
-	autoDiskQuotaLocalReservedSize uint64 = LocalMemoryTableSize
+	defaultEngineMemCacheSize      = 512 * units.MiB
+	defaultLocalWriterMemCacheSize = 128 * units.MiB
 
 	// autoDiskQuotaLocalReservedSpeed is the estimated size increase per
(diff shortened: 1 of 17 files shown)


overvenus:
Do we need to cherry pick it to release-4.0?

kennytm:
@overvenus preferred but not blocking.

pkg/lightning/restore/restore.go
```diff
@@ -2221,8 +2209,8 @@ func (tr *TableRestore) analyzeTable(ctx context.Context, g glue.SQLExecutor) er
 ////////////////////////////////////////////////////////////////
 
 var (
-	maxKVQueueSize         = 128   // Cache at most this number of rows before blocking the encode loop
-	minDeliverBytes uint64 = 65536 // 64 KB. batch at least this amount of bytes to reduce number of messages
+	maxKVQueueSize         = 128            // Cache at most this number of rows before blocking the encode loop
+	minDeliverBytes uint64 = 96 * units.KiB // 96 KB (data + index). batch at least this amount of bytes to reduce number of messages
```
glorv, line 2213: Is this change has side effect for tidb backend because we increase the batch size?🤔

pkg/lightning/restore/restore.go
```diff
@@ -1396,19 +1396,7 @@ func (t *TableRestore) restoreEngine(
...
-	// (region-concurrency + index-concurrency) * 512 MiB.
-	// This number should not exceed the disk quota.
-	// Therefore, we need to reduce that "512 MiB" to respect the disk quota:
-	localWriterMaxCacheSize := int64(rc.cfg.TikvImporter.DiskQuota) // int64(rc.cfg.App.IndexConcurrency+rc.cfg.App.RegionConcurrency)
-	if localWriterMaxCacheSize > config.LocalMemoryTableSize {
-		localWriterMaxCacheSize = config.LocalMemoryTableSize
-	}
-
-	indexWriter, err := indexEngine.LocalWriter(ctx, localWriterMaxCacheSize)
+	indexWriter, err := indexEngine.LocalWriter(ctx)
```
glorv, line 1399: I perfer to create a `indexEngineWriter` for each chunk  since the performance is much better

glorv (commented):
rest LGTM

pkg/lightning/restore/restore.go
kennytm: the TiDB backend does `SplitIntoChunks` inside `WriteRows`, so fortunately no.

pkg/lightning/config/config.go
```diff
@@ -613,6 +614,9 @@ func (cfg *Config) Adjust(ctx context.Context) error {
 			}
 			cfg.TikvImporter.DiskQuota = ByteSize(storageSize.Available - reservedSize)
 		}
+		if cfg.TikvImporter.DiskQuota < cfg.TikvImporter.LocalWriterMemCacheSize {
```
lance6716, line 617: should we solve `LocalWriterMemCacheSize` for `x = storageSize.Available - (enginesCount*x + writeAmount*autoDiskQuotaLocalReservedSpeed)`

and I'm not sure if this is to prevent any flush triggering disk quota. maybe we should also left a margin

lance6716 (commented):
rest LGTM

pkg/lightning/config/config.go
kennytm: just removed this line of code 🙃

ti-srebot (approved):

PR #850, opened 2021-03-11
pkg/lightning: replace local generated ts with tso fetch from pd

glorv:
Replace commit ts generated with local time by tso fetch from pd. This can address the issue that if local time is mush faster than tidb cluster time, after import data with local/importer backend, query may returns empty result.

- Fetch a ts from pd and store it in `RestoreController` and use this ts for all `WriteRows`
- Always fetch the newest ts from pd when do table checksum. This is useful when we support concurrent import with many lightnings and only do checksum after the last lightning finished.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - Fix the bug that lightning generated ts may be to large or small that query may return incorrect result.


pkg/lightning/restore/restore.go
```diff
@@ -241,6 +246,25 @@ func NewRestoreControllerWithPauser(
...
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		defer pdClient.Close()
+
+		physical, logical, err := pdClient.GetTS(ctx)
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		ts = oracle.ComposeTS(physical, logical)
```
kennytm, line 265: please use `pdutil.NewPdController`

kennytm (commented):
rest LGTM for now.

i don't particularly like requiring TS for TiDB backend, but let's refactor this later.

kennytm (commented):
/lgtm

glorv:
@lance6716 @Little-Wallace  PTAL

PR #853, opened 2021-03-11
br: remove alter-primary-key configuration

tangenta:

https://github.com/pingcap/tidb/pull/23270 will deprecate the `alter-primary-key` config.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

NA

Side effects

 - Breaking backward compatibility?

Related changes

NA

### Release Note

 - No release note


kennytm:
you need to change the `tests/br_incompatible_config` test case as well. it will start a TiDB using the `alter-primary-key` config.

tangenta:
I notice that `allow-auto-random` is deprecated too and it is not removed. Should I remove both of them?

kennytm:
@tangenta  yes that too.

since now alter-pk is opt-in, some DDL statements in `tests/br_incompatible_config` should be changed to use that `/*T![clustered_index] NONCLUSTERED */` syntax.

tangenta:
```sql
mysql> create table t (a int primary key nonclustered, b int unique);
Query OK, 0 rows affected (0.02 sec)

mysql> insert into t values (42, 42);
Query OK, 1 row affected (0.01 sec)

mysql> alter table t drop primary key;
ERROR 8200 (HY000): Unsupported drop primary key when alter-primary-key is false
```

This is because https://github.com/pingcap/tidb/pull/23270 is not merged. We should merge one first since these two PRs depend on each other.

kennytm:
@tangenta why 23270 depends on this PR?

tangenta:
Because TiDB depends on BR and 23270 removes the `AlterPrimaryKey` field from `Config`. The build will fail.

OK, let me bring back `AlterPrimaryKey` field and merge 23270 first.

lysu:
@tangenta please help resolve conflict

ti-srebot (approved):

Issue #849, opened 2021-03-11
backup and recovery  rawkv requests specifying the startkey and endkey manually.


yuzhibotao:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
backup rawkv using br must specify startkey and endkey manually.

2. What did you expect to see?
I want to remove startkey and endkey  in command to backup all data in rawkv.

3. What did you see instead?

4. What version of BR and TiDB/TiKV/PD are you using?
v4.0.10

5. Operation logs
   - Please upload `br.log` for BR if possible
   - Please upload `tidb-lightning.log` for TiDB-Lightning if possible
   - Please upload `tikv-importer.log` from TiKV-Importer if possible
   - Other interesting logs

6. Configuration of the cluster and the task
   - `tidb-lightning.toml` for TiDB-Lightning if possible
   - `tikv-importer.toml` for TiKV-Importer if possible
   - `topology.yml` if deployed by TiUP

7. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus if possible

YuJuncen:
If possible, could you help to test whether `--start-key "" --end-key ""` works?

YuJuncen:
After a short testing, seems it would be OK to pass empty string directly. (Even we need to modify BR to bypass some tests.)

A PR can be provided soon.

<details>
<summary>
Test script:
</summary>
test_full_rawkv() {
    check_range_start=00
    check_range_end=ff

    checksum_full=$(checksum $check_range_start $check_range_end)
    # backup current state of key-values
    run_br --pd $PD_ADDR backup raw -s "local://$TEST_DIR/rawkv-full" --format raw

    clean $check_range_start $check_range_end
    # Ensure the data is deleted
    checksum_new=$(checksum $check_range_start $check_range_end)
    if [ "$checksum_new" == "$checksum_full" ];then
        echo "failed to delete data in range"
        fail_and_exit
    fi

    run_br --pd $PD_ADDR restore raw -s "local://$TEST_DIR/rawkv-full" --format raw
    checksum_new=$(checksum $check_range_start $check_range_end)
    if [ "$checksum_new" != "$checksum_full" ];then
        echo "failed to restore"
        fail_and_exit
    fi
}
</details>

PR #861, opened 2021-03-12
raw_backup, raw_restore: allow empty start_key and end_key

YuJuncen:
Fixed https://github.com/pingcap/br/issues/849.

https://github.com/pingcap/br/issues/849#issuecomment-797239056

Besides, removed a constraint start key should not equal end key if they both are empty(In fact, the presents -inf and +inf, which isn't the same at all), and prettified some log output.

### Check List 

Tests 

 - Integration test

Related changes

### Release Note

 - Now, It's legal to run BR backup/restore raw without `--start-key` or `--end-key`.


pkg/redact/redact.go
```diff
@@ -32,5 +32,9 @@ func Key(key []byte) string {
 	if NeedRedact() {
 		return "?"
 	}
+	if len(key) == 0 {
+		// TiKV treats zero length keys as infinate(i.e. unbounded).
+		return "inf"
+	}
```
kennytm, line 38: looks unnecessary

```suggestion
```

YuJuncen: Yep, identify it was `+inf` or `-inf` is impossible in this context... Remote it is acceptable.

pkg/restore/client.go
```diff
@@ -219,7 +220,10 @@ func (rc *Client) GetFilesInRawRange(startKey []byte, endKey []byte, cf string)
 			utils.CompareEndKey(endKey, rawRange.EndKey) > 0 {
 			// Only partial of the restoring range is in the current backup-ed range. So the given range can't be fully
 			// restored.
-			return nil, errors.Annotate(berrors.ErrRestoreRangeMismatch, "the given range to restore is not fully covered by the range that was backed up")
+			return nil, errors.Annotatef(berrors.ErrRestoreRangeMismatch,
+				"the given range to restore (%s, %s) is not fully covered by the range that was backed up(%s, %s)",
```
kennytm, line 224:
```suggestion
				"the given range to restore [%s, %s) is not fully covered by the range that was backed up [%s, %s)",
```

kennytm (commented):
rest LGTM

ti-srebot (approved):

YuJuncen:
/run-integration-tests

well...

PR #862, opened 2021-03-12
log: add FilterCore and filter TiDB log

overvenus:

Add FilterCore and filter TiDB log, also it brings Lightning missing logs back, in case it does not initialize pingcap/log's global logger.

### Check List 

Tests 

 - Unit test

Related changes

### Release Note

 - No release note.


cmd/br/cmd.go
```diff
@@ -113,7 +116,13 @@ func Init(cmd *cobra.Command) (err error) {
 			// cmd.PrintErr prints to stderr, but PrintErrf prints to stdout.
 			cmd.PrintErr(fmt.Sprintf("Detail BR log in %s \n", conf.File.Filename))
 		}
-		lg, p, e := log.InitLogger(conf)
+		// FilterCore requires zap.AddCaller.
+		conf.DisableCaller = false
+		filterTiDBLog := zap.WrapCore(func(core zapcore.Core) zapcore.Core {
+			// Filter logs from TiDB.
+			return lglog.NewFilterCore(core, "github.com/pingcap/tidb/")
```
kennytm, line 123: can we filter PD too 🥺

pkg/lightning/log/filter.go
```diff
@@ -0,0 +1,49 @@
...
+// package path.
+func (f *FilterCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	for i := range f.filters {
+		// Caller.Function is a package path-qualified function name.
+		if strings.Contains(entry.Caller.Function, f.filters[i]) {
+			return nil
+		}
+	}
+	return f.Core.Write(entry, fields)
+}
```
kennytm, line 49: From the [doc's description](https://pkg.go.dev/go.uber.org/zap/zapcore#Core) IMO we should implement the filter logic in `Check()` instead of `Write()`.

`Write` can be kept unimplemented by forwarding to the inner `Core`.

Also you should implement `With()`.

overvenus: I tried `Check`, but I find `Entry.Caller` is not initialized when it is passed `Check()`.  

https://github.com/uber-go/zap/blob/v1.16.0/logger.go#L268-L277

cmd/br/cmd.go
overvenus: Sure, I will filter PD for Lightning.

pkg/lightning/log/filter.go
kennytm: Ok. Then can we keep `Check` unimplemented (i.e. forward to `zapcore.Core`)?

overvenus: I tried, we have to override `Check` and add `FilterCore` to `CheckedEntry`, because `Write()` is invoked via `CheckedEntry.core.Write()`.

https://github.com/uber-go/zap/blob/v1.16.0/zapcore/entry.go#L200-L227

kennytm: bruh.

pkg/lightning/log/filter.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	return ce
+}
+
+// Write filters entry by checking if entry's Caller.Function matches filtered
+// package path.
+func (f *FilterCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	for i := range f.filters {
+		// Caller.Function is a package path-qualified function name.
+		if strings.Contains(entry.Caller.Function, f.filters[i]) {
```
kennytm, line 44: (just wondering if it is more efficient to use a `regexp` instead of a loop + `strings.Contains`)

pkg/lightning/log/filter.go
```diff
@@ -28,6 +28,12 @@ func NewFilterCore(core zapcore.Core, filteredPackages ...string) *FilterCore {
 	}
 }
 
+// With adds structured context to the Core.
+func (f *FilterCore) With(fields []zapcore.Field) zapcore.Core {
+	f.Core = f.Core.With(fields)
+	return f
```
kennytm, line 34:
```suggestion
	return &FilterCore{
		Core:    f.Core.With(fields),
		filters: f.filters,
	}
```

`With` is not a mutable method.

pkg/lightning/log/filter.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return ce
+}
+
+// Write filters entry by checking if entry's Caller.Function matches filtered
+// package path.
+func (f *FilterCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	for i := range f.filters {
+		// Caller.Function is a package path-qualified function name.
+		if strings.Contains(entry.Caller.Function, f.filters[i]) {
```
lance6716, line 52: (we can't use logger of lightning instance when IMPORT via SQL, at that time we should use tidb's logger, right?)

pkg/lightning/log/filter_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"go.uber.org/zap/zapcore"
+
+	"github.com/pingcap/br/pkg/lightning/log"
+)
+
+var _ = Suite(&testFilterSuite{})
+
+type testFilterSuite struct{}
+
+func (s *testFilterSuite) TestFilter(c *C) {
```
lance6716, line 20: how about test a `NameError` or something, which will print stacks that contain the specified package name

lance6716 (commented):
rest LGTM

pkg/lightning/log/filter.go
overvenus: For IMPORT via SQL, we do not need to do anything, it automatically proxies to tidb's global logger.

pkg/lightning/log/filter_test.go
overvenus: Could you show an example?

pkg/lightning/log/filter.go
overvenus: I will replace `lightning/log.L().X()` to `pingcap/log.X()`.

pkg/lightning/log/filter_test.go
lance6716: maybe

```
	// stack messages won't trigger filter
	logger, buffer = log.MakeTestLogger(zap.WrapCore(func(c zapcore.Core) zapcore.Core {
		return log.NewFilterCore(c, "github.com/pingcap/check").With([]zap.Field{zap.String("a", "b")})
	}), zap.AddCaller())
	logger.Warn("the message", zap.NamedError("stack", errors.New("test")))
	c.Assert(strings.Contains(buffer.Stripped(), "github.com/pingcap/check"), IsTrue)
```

ti-srebot (approved):

PR #871, opened 2021-03-15
br flags: --no-credentials and --skip-check-path

gregwebs:

This helps with local testing of GCS and might also be useful in other contexts.

With gcs-fake-server running backup can be performed with these options:

```
br --gcs.endpoint http://localhost:4443/storage/v1 \
  --no-credentials --send-credentials-to-tikv=false --skip-check-path
```

### Check List 

Tests

 - Manual test (see above)

Code changes

 - Has exported function/method change: no
 - Has exported variable/fields change: only additions
 - Has interface methods change: no
 - Has persistent data change: no

Side effects

- ?

Related changes

### Release Note

 - Add testing options --no-credentials and --skip-check-path

### DIFF
--- a/pkg/task/common.go
+++ b/pkg/task/common.go
@@ -35,6 +35,8 @@ import (
 const (
 	// flagSendCreds specify whether to send credentials to tikv
 	flagSendCreds = "send-credentials-to-tikv"
+	// No credentials specifies that cloud credentials should not be loaded
+	flagNoCreds = "no-credentials"
 	// flagStorage is the name of storage flag.
 	flagStorage = "storage"
 	// flagPD is the name of PD url flag.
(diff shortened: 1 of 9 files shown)


pkg/task/common.go
```diff
@@ -175,6 +184,9 @@ func DefineCommonFlags(flags *pflag.FlagSet) {
 	flags.Bool(flagEnableOpenTracing, false,
 		"Set whether to enable opentracing during the backup/restore process")
 
+	flags.BoolP(flagNoCreds, "", false, "Don't load credentials")
+	flags.BoolP(flagSkipCheckPath, "", false, "Skip path verification")
```
kennytm, line 188: for now please mark these two flags "Hidden" (see line 170 above) since afaik they are only used in tests.

kennytm (commented):
Rest LGTM

pkg/task/common.go
gregwebs: okay, they are marked as hidden now

ti-srebot (approved):

Issue #898, opened 2021-03-18
S3's `check-path` is not specific enough


kennytm:
Please answer these questions before submitting your issue. Thanks!

### 1. What did you do?

1. dump some SQL files into `s3://some-bucket/some-prefix/` on AWS S3.
2. create a role, give it the `s3:ListBucket` and `s3:GetObject` permission, conditioned with `s3:prefix`=`some-prefix/`.
3. run Lightning using that role

### 2. What did you expect to see?

successfully restored.

### 3. What did you see instead?

failed due to an unnecessary `HeadBucket` call:

```
tidb lightning encountered error:  create storage failed: Bucket some-bucket is not accessible: Forbidden: Forbidden
        status code: 403, request id: ..., host id: ...: [BR:ExternalStorage:ErrStorageInvalidConfig]invalid external storage config
```

### 4. What version of BR and TiDB/TiKV/PD are you using?

master

kennytm:
Note: affects all users of `br/pkg/storage`, including BR itself (can workaround with `--skip-check-path`), TiCDC and Dumpling (both have SkipCheckPath hard-coded to false).

handlerww:
Reproduce:
1 Create IAM policy, select S3 permission with GetObject and ListBucket, add a request condition about s3:prefix.

![image](https://user-images.githubusercontent.com/2058782/111681583-5dfb8480-885e-11eb-8293-e5db3f9f4774.png)

2. Create a IAM role, attach the policy created before to this role

3. Set IAM role of EC2 to a certain role
    
![image](https://user-images.githubusercontent.com/2058782/111680492-4079eb00-885d-11eb-92ff-3cca0c5c083c.png)

lonng:
@glorv Seems this bug had been fixed in #899, can we close this issue?

glorv:
closed in favor of https://github.com/pingcap/dumpling/pull/286 and https://github.com/pingcap/ticdc/pull/1890

PR #899, opened 2021-03-18
lightning: replace storage.Create by storage.New

kennytm:

1. `storage.Create` has been deprecated... so let's replace the remaining two calls from Lightning.
2. Lightning originally used `storage.Create(sendCreds=true)`, but 
    * the entire cloud access happens within Lightning, so we don't need to send credentials to anyone
    * this introduced `skipCheckPath=false`, which leads to #898. 

Changed `storage.Create` to `storage.New` with sendCreds=false and skipCheckPath=true.

### Check List 

Tests 

Code changes

Side effects

Related changes

### Release Note

 - When using Lightning to import from S3, we no longer require `s3:ListBucket` permission on the entire bucket, only the data source prefix itself.


glorv:
So shall we also remote the `storage.Create` source code since it's not recommanded.

kennytm:
@glorv Yes after it has been removed from Dumpling.

ti-srebot (approved):

PR #905, opened 2021-03-22
pkg/lightning: let batch split keys also consider the raft entry limit

glorv:
Avoid error like following:
```
[2021/03/19 19:29:35.040 +08:00] [WARN] [localhelper.go:184] ["split regions"] [error="batch split regions failed: split region failed: err=message:\"raft entry is too large, region 173229565, entry size 15024314\" raft_entry_too_large:<region_id:173229565 entry_size:15024314 > : [BR:Restore:ErrRestoreSplitFailed]fail to split region"] ["retry time"=1] [region_id=173229565]
```

In case when each index key size is too large e.g. several kb, the total size of 4096 keys may exceed the raft msg size 8MB which causes `BatchSplitRegion` to fail. 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - Fix the bug that batch split region fails due to total key size exceeds raft entry limit


pkg/lightning/backend/local/localhelper.go
```diff
@@ -47,6 +48,9 @@ const (
 var (
 	// the max keys count in a batch to split one region
 	maxBatchSplitKeys = 4096
+	// the max total key size in a split region batch.
+	// it shouldn't exceed tikv raft max entry size(default is 8MB).
+	maxBatchSplitSize = 8 * units.MB
```
Little-Wallace, line 53: This is too large because there may be some meta message appended by TiKV.

Little-Wallace (commented):
rest lgtm

Little-Wallace (approved):
LGTM

Issue #904, opened 2021-03-22
br backup checksum will fail immediately when target s3 disconnected


Tammyxia:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

- During br backup checksum, disconnecting target s3 storage.
- Check br backup status, find checksum fail immediately(has not time toleration for target disconnecting).

2. What did you expect to see?

3. What did you see instead?
4. 
![image](https://user-images.githubusercontent.com/78345569/111930363-b4352580-8af3-11eb-9fca-c1ef64a0860b.png)

4. What version of BR and TiDB/TiKV/PD are you using?

5. Operation logs
   - Please upload `br.log` for BR if possible
   - Please upload `tidb-lightning.log` for TiDB-Lightning if possible
   - Please upload `tikv-importer.log` from TiKV-Importer if possible
   - Other interesting logs

6. Configuration of the cluster and the task
   - `tidb-lightning.toml` for TiDB-Lightning if possible
   - `tikv-importer.toml` for TiKV-Importer if possible
   - `topology.yml` if deployed by TiUP

7. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus if possible

PR #906, opened 2021-03-22
storage: Add S3 retry time 

YuJuncen:
solve #904 

Added the min backoff time of S3 client to 1s, max retry time to 7 times.
Hence with exponential backoff algorithm, we can wait for about (1+2+4+8+16+32+64) = 2^7-1 secs.

### Check List 

Tests 

 - Unit test

### Release Note

 - The backoff time for S3 errors was increased.


pkg/backup/client.go
```diff
@@ -216,6 +218,19 @@ func (bc *Client) SaveBackupMeta(ctx context.Context, backupMeta *backuppb.Backu
...
+			"process will sleep for 3s and notify the shell to kill s3 service.")
+		if pid, ok := v.(int); ok {
+			proc, err := os.FindProcess(pid)
+			if err != nil {
+				log.Warn("failed to find shell to notify, skipping notify", zap.Error(err))
+			} else if err := proc.Signal(syscall.SIGUSR1); err != nil {
+				log.Warn("failed to notify the shell, skipping notify", zap.Error(err))
+			}
+		}
+		time.Sleep(3 * time.Second)
```
kennytm, line 232: i prefer touching a file than sending a signal.

1. SIGUSR1 is unix-specific so it can't be compiled on windows
2. The integrated test relies on a sleep(1) loop, so this SIGUSR1 won't immediately take effect. Thus there's no advantage compared with communicating via existence of a file (which is cross-platform).

YuJuncen: done

kennytm (commented):
/lgtm

ti-srebot (approved):

Issue #790, opened 2021-03-03
Tracking issue of BR clustered index compatibility


overvenus:
Compatibility of BR and clustered index issues: 

- [x] 1. Restore a clustered index backup to TiDB 4.0.x causes data corruption.

<details> <summary> Steps </summary>

```
# Execute on TiDB 5.0
MySQL [(none)]> SET tidb_enable_clustered_index = 1;
MySQL [(none)]> CREATE TABLE composite_primary_key (
 key_a INT NOT NULL,
 key_b INT NOT NULL,
 b CHAR(100),
 PRIMARY KEY (key_a, key_b)
);
MySQL [(none)]> INSERT INTO composite_primary_key VALUES (1, 1, 'aaa'), (2, 2, 'bbb');

# Make a backup
tiup br:nightly backup full -s "local:///backup/clustered"
# Restore the backup on TiDB 4.0.11
tiup br:v4.0.11 restore full -s "local:///backup/clustered" --check-requirements=false

# Execute on TiDB 4.0.11
MySQL [(none)]> select * from test.composite_primary_key;
ERROR 1105 (HY000): [components/tidb_query/src/batch/executors/table_scan_executor.rs:338]: Data is corrupted, missing data for NOT NULL column (offset = 0)
```

</details>

- [x] 2. Restore a non-clustered index backup to TiDB 5.0 causes data corruption.

<details> <summary> Steps </summary>

```
# Execute on TiDB 4.0.11
MySQL [(none)]> CREATE TABLE composite_primary_key (
 key_a INT NOT NULL,
 key_b INT NOT NULL,
 b CHAR(100),
 PRIMARY KEY (key_a, key_b)
);
MySQL [(none)]> INSERT INTO composite_primary_key VALUES (1, 1, 'aaa'), (2, 2, 'bbb');

# Make a backup

# Execute on TiDB 5.0
MySQL [(none)]> SET tidb_enable_clustered_index = 1;
MySQL [(none)]> CREATE TABLE composite_primary_key (
 key_a INT NOT NULL,
 key_b INT NOT NULL,
 b CHAR(100),
 PRIMARY KEY (key_a, key_b)
);

# Restore the backup on TiDB 5.0
tiup br:v4.0.11 restore full -s "local:///backup/notclustered" --check-requirements=false

MySQL [(none)]> select * from test.composite_primary_key;
ERROR 1105 (HY000): invalid data type: unsupported data type `128`
```

</details>

[...]

PR #929, opened 2021-03-25
br/restore: add version check for backup schema_version

YuJuncen:
a special fix for v4.0 versions of #790 

Because #685 cannot be cherry-picked to release-4.0, 4.0.x BR would corrupt 4.0 clusters without even a warn message when the backup archive come from a table with non-int clustered primary key. This PR would add a check to the backup archive version, and intercept restoring from incompatible backup archives.

CheckClusterVersion was changed to check version in different way.
And added a check to cluster version stored in backupmeta during restore. (after reading backup meta)

**Currently, this check only checks major versions (because of clustered index).**

### Check List 

Tests 

 - Unit test

Code changes

 signature change for `CheckClusterVersion`.
``` diff
- CheckClusterVersion(ctx context.Context, client pd.Client) error
+ CheckClusterVersion(ctx context.Context, client pd.Client, checker VerChecker) error
```

### Release Note

 - BR would check cluster version of backup now.


pkg/version/version.go
```diff
@@ -6,15 +6,17 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
+	"github.com/pingcap/log"
+	pd "github.com/tikv/pd/client"
+	"go.uber.org/zap"
+
```
overvenus, line 15:
```suggestion
```

pkg/version/version.go
```diff
@@ -6,15 +6,17 @@ import (
...
 	"github.com/pingcap/kvproto/pkg/metapb"
```
overvenus, line 17:
```suggestion
```

pkg/version/version.go
```diff
@@ -199,3 +224,20 @@ func CheckTiDBVersion(versionStr string, requiredMinVersion, requiredMaxVersion
 	}
 	return CheckVersion("TiDB", *version, requiredMinVersion, requiredMaxVersion)
 }
+
+// NormalizeBackupVersion normalizes the version string from backupmeta.
+func NormalizeBackupVersion(version string) *semver.Version {
```
overvenus, line 229: Seem `version.removeVAndHash` can be replaced by this funcation.

overvenus (commented):
rest LGTM

pkg/version/version.go
YuJuncen: You mean merge those two functions? Seems `removeVAndHash` focus on the version string made by the make script, and `NormalizeBackupVersion` focus on the version string returned from TiKV. Maybe it's hard to just replace the former with the latter. 🤔

YuJuncen:
/run-integration-tests

port in use

pkg/version/version.go
```diff
@@ -99,44 +99,67 @@ func CheckClusterVersion(ctx context.Context, client pd.Client) error {
...
+	return nil
+}
 
-		if tikvVersion.Major != BRVersion.Major {
-			return errors.Annotatef(berrors.ErrVersionMismatch, "TiKV node %s version %s and BR %s major version mismatch, please use the same version of BR",
-				s.Address, tikvVersionString, build.ReleaseVersion)
+// CheckVersionForBackup checks the version for backup and
+func CheckVersionForBackup(backupVersion *semver.Version) VerChecker {
+	return func(store *metapb.Store, ver *semver.Version) error {
+		if backupVersion.Major != ver.Major {
```
3pointer, line 116: How about change it to `backupVersion.Major < ver.Major`?

YuJuncen: sounds good

YuJuncen:
listen tcp 0.0.0.0:10080: bind: address already in use

/run-integration-tests

3pointer (commented):
/lgtm

YuJuncen:
[2021-05-08T11:29:42.533Z] [2021/05/08 19:29:40.505 +08:00] [INFO] [restore_raw.go:110] ["all files are filtered out from the backup archive, nothing to restore"]

/run-integration-tests

PR #948, opened 2021-03-27
backup: retry connection reset by peer error

3pointer:
sometimes s3 might return a new connection error, such as `Error: msg:"Io(Custom { kind: Other, error: \"failed to put object Error during dispatch: connection error: Connection reset by peer (os error 104)\" })"  on storeID: 1: [BR:KV:ErrKVUnknown]unknown tikv error` and we need handle it.

Add a new error check

### Check List 

Tests 

Related changes

### Release Note

 - Fix the issue that br cannot retry on s3 connect reset by peer error.


pkg/utils/retry.go
```diff
@@ -47,13 +47,15 @@ func WithRetry(
...
-	// If s3 stop or not start.
-	closedOrRefused := strings.Contains(msgLower, "server closed") || strings.Contains(msgLower, "connection refused")
+	// If failed to read/write to S3/GCS.
+	failed := strings.Contains(msgLower, "failed to put object") ||
+		strings.Contains(msgLower, "failed to get object") || strings.Contains(msgLower, "invalid http request")
+	// If S3/GCS stop or not start.
+	closedOrRefused := strings.Contains(msgLower, "server closed") || strings.Contains(msgLower, "writing a body to connection") ||
+		strings.Contains(msgLower, "connection refused") ||  strings.Contains(msgLower, "connection reset by peer")
 	// Those conditions are retryable.
 	return failed && closedOrRefused
```
YuJuncen, line 60: I think maybe `closedOrRefused` conditions could always be treated at retryable? 🤔 (As those are enough for identify a server error.)

```suggestion
	return closedOrRefused
```

3pointer: sure, Fix it.

YuJuncen:
HTTP errors from hyper can be retried:
- "channel closed"
- "error trying to connect"
- "connection closed before message completed"
- "body write aborted"
- ...(TBD)

pkg/utils/retry.go
```diff
@@ -47,13 +47,11 @@ func WithRetry(
...
 	msgLower := strings.ToLower(msg)
-	// If failed to read/write to S3.
-	failed := strings.Contains(msgLower, "failed to put object") || strings.Contains(msgLower, "failed to get object")
-	// If s3 stop or not start.
-	closedOrRefused := strings.Contains(msgLower, "server closed") || strings.Contains(msgLower, "connection refused")
-	// Those conditions are retryable.
-	return failed && closedOrRefused
+	// UNSAFE! TODO: Add a error type for retryable connection error.
+	// If S3/GCS stop or not start.
+	return strings.Contains(msgLower, "server closed") || strings.Contains(msgLower, "writing a body to connection") ||
```
YuJuncen, line 55: Maybe we can use an array here and test all GCS error messages here.

BTW, maybe `"error during dispatch"` would be a more generic pattern? 🤔

YuJuncen:
`value *errors.withStack = listen tcp 0.0.0.0:11950: bind: address already in use ("listen tcp 0.0.0.0:11950: bind: address already in use")`

/run-integration-tests

Issue #801, opened 2021-03-04
Check whether table has data before the restoration.


3pointer:
### Describe your feature request related problem:

problem is like #797
When the table already has some data before we do a full restore job and the job will be failed. The response error is `checksum mismatched`.

### Describe the feature you'd like:

We'd better add a pre-check before this restoration. because the whole restoration may take a long time to finish. if the checksum result is not reliable. we should let the user know it at the beginning.

### Describe alternatives you've considered:

This check should only use in the full restoration.

feniljain:
Hey, I would like to take up this issue.

PR #954, opened 2021-03-28
restore: Adding table empty check before starting full restore operation

feniljain:
https://github.com/pingcap/br/issues/801

A check for total bytes is kept on the tables and checked how many are filled, and the filled tables are returned, where further execution is stopped if there are any filled tables present.

### Check List 

Tests 

 -  Follow instructions from README.md to have the setup ready, take backup and do not drop databases and run the br restore command

Code changes

Side effects

 - Increased code complexity

### Release Note

 - No release note

### DIFF
--- a/pkg/task/restore.go
+++ b/pkg/task/restore.go
@@ -218,7 +218,10 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 		return errors.Annotate(berrors.ErrRestoreModeMismatch, "cannot do transactional restore from raw kv data")
 	}
 
-	files, tables, dbs := filterRestoreFiles(client, cfg)
+	files, tables, dbs, filledTables := filterRestoreFiles(client, cfg)
+	if len(filledTables) > 0 {
+		return errors.Annotate(berrors.ErrRestoreInvalidBackup, "tables already present")
+	}
 	if len(dbs) == 0 && len(tables) != 0 {
 		return errors.Annotate(berrors.ErrRestoreInvalidBackup, "contain tables but no databases")
 	}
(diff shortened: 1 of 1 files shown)


pkg/task/restore.go
```diff
@@ -247,11 +250,6 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 	}
 	ddlJobs := restore.FilterDDLJobs(client.GetDDLJobs(), tables)
 
-	err = client.PreCheckTableTiFlashReplica(ctx, tables)
-	if err != nil {
-		return errors.Trace(err)
-	}
-
```
kennytm, line 254: these are accidentally removed?

pkg/task/restore.go
```diff
@@ -420,10 +418,14 @@ func dropToBlackhole(
...
 	client *restore.Client,
 	cfg *RestoreConfig,
-) (files []*backuppb.File, tables []*utils.Table, dbs []*utils.Database) {
+) (files []*backuppb.File, tables []*utils.Table, dbs []*utils.Database, filledTables []*utils.Table) {
 	for _, db := range client.GetDatabases() {
 		createdDatabase := false
 		for _, table := range db.Tables {
+			if table.TotalBytes > 0 {
+				filledTables = append(filledTables, table)
+			}
```
kennytm, line 427: um, so,

when we perform restore, there are two "tables" to deal with:

1. the table inside the backup archive, or the "source"
2. the live table in the target database, or the "destination"

what BR restore does is copy all data from the "source" into the "destination". 

**here, `table.TotalBytes` refer to the size of the "source", not the size of "destination"**.

in #801 the meaning of "table has data" refers to the destination, i.e. checking if the table in the target database is empty. 

if the destination is not empty, we cannot ensure `checksum(source) == checksum(destination)`.

i suggest you do a simple check before submitting a fix:

1. BR backup a table `t` into `local:///tmp/sample/`
2. Do not truncate `t`, keep it non-empty.
3. BR restore from `local:///tmp/sample/`
4. If #801 is active, it should throw a warning/error because table `t` is not empty.
5. Now truncate table `t`
6. BR restore from `local:///tmp/sample/` again.
7. There should be no errors, and table `t` is filled with the old data.

feniljain: I am extremely sorry for misunderstanding the issue completely. Thank you again for taking your time and clarifying the issue in detail, indeed they were different tables I was looking at. For the steps you mentioned I tried them out and there seems to be no error whatsoever in any step, so I guess the issue was fixed already. With your go I will close this PR hence worth.

3pointer: I think this issue is not fixed yet. 
Before step.3. you can try to insert some data into table. `t`. and step.4. will throw an error which is `checksum mismatch...`.

PR #959, opened 2021-03-29
restore: set sst size in SSTFileMeta correctly

3pointer:
Upload size correctly to avoid the empty regions after the restoration.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release Note

 -


kennytm:
"upload" means "update"? 😅

3pointer:
> "upload" means "update"? 😅

🤔️ eh..., yes, maybe update/set is better.

PR #967, opened 2021-03-30
 check table is empty before do import

july2993:
### What problem does this PR solve? 
if the user DO NOT know lighting can only init an empty table, it may
corrupt the table, also, will fail to pass checksum after import.

### What is changed and how it works?
check table is empty before do import data.
if it is not empty, print the related error to force the user to clean up the table if the user wants to import the table.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
  
if the table is empty run success, if the table is not empty runs fail like this:
```
[2021/03/30 17:07:31.064 +08:00] [INFO] [restore.go:612] ["restore all schema completed"] [takeTime=15.574742ms] []
[2021/03/30 17:07:31.066 +08:00] [ERROR] [restore.go:327] ["run failed"] [step=2] [error="table `test`.`test1` not empty, please cleanup the table first"]
Error: table `test`.`test1` not empty, please cleanup the table first
[2021/03/30 17:07:31.066 +08:00] [ERROR] [restore.go:338] ["the whole procedure failed"] [takeTime=22.20738ms] [error="table `test`.`test1` not empty, please cleanup the table first"]
[...]

PR #985, opened 2021-04-06
Fix check empty table in checkpoint enabled case

july2993:
ref https://github.com/pingcap/br/pull/967
fix the case when checkpoint enabled
after `checkpoint.Initialize()`, it will insert a `CheckpointStatusLoaded` status checkpoint for the missing checkpoint tables.
so it do not work for the case when checkpoint enabled.

make Get() of mysql driver checkpoint return a CheckpointStatusMissing
status instead of return error like file driver checkpoint

check if the checkpoint status is CheckpointStatusMissing before we
Initialize() the checkpointsDB which will add a CheckpointStatusLoaded
status checkpoint for the table if it is missing.

### Check List 

Tests 

 - Unit test
   if the table is empty run success, if the table is not empty runs fail(with a CheckpointStatusAnalyzed status checkpoint still success)

 - No code

Code changes

Side effects

Related changes

### Release Note

 - No release note


july2993:
/run-all-tests
fail br raw kv?

pkg/lightning/checkpoints/checkpoints.go
```diff
@@ -770,7 +770,7 @@ func (cpdb *MySQLCheckpointsDB) Get(ctx context.Context, tableName string) (*Tab
 		tableRow := tx.QueryRowContext(c, tableQuery, tableName)
 
 		var status uint8
-		if err := tableRow.Scan(&status, &cp.AllocBase, &cp.TableID); err != nil {
+		if err := tableRow.Scan(&status, &cp.AllocBase, &cp.TableID); err != nil && err != sql.ErrNoRows {
```
glorv, line 773: Why need to skip the `ErrNoRows` error. I think it's better to check this error explicitly if caller need to skip it.

glorv (commented):
rest LGTM

pkg/lightning/checkpoints/checkpoints.go
july2993: https://github.com/pingcap/br/blob/master/pkg/lightning/checkpoints/checkpoints_file_test.go#L217

for the file driver, it will return a Missing `CheckpointStatusMissing` status checkpoint, we should keep the same behavior.

Also, the users should not care about the different implementations of the checkpoint DB.

glorv: Since only the check empty tables logic depends on the get operation before checkpoint db is inited, I think you can make both MySQL and file checkpoint return nil or something like NotFound error to represent that this checkpoint data do not exist. In other normal logic, I don't think it's a good idea to just ignore it.

july2993: > Since only the check empty tables logic depends on the get operation before checkpoint db is inited, I think you can make both MySQL and file checkpoint return nil or something like NotFound error to represent that this checkpoint data do not exist. In other normal logic, I don't think it's a good idea to just ignore it.

then what's the usage of `CheckpointStatusMissing` status?

tests/lightning_local_backend/run.sh
```diff
@@ -97,6 +97,9 @@ for ckpt in mysql file; do
   grep -Fq "No table has lost intermediate files according to given config" $TEST_DIR/lightning_ctl.output
 
   # when position of chunk file doesn't equal to offset, intermediate file should exist
+  run_sql 'DROP DATABASE IF EXISTS cpeng;'
```
glorv, line 100: Why do clean up here since checkpoint is enabled

july2993: this pr make https://github.com/pingcap/tidb-lightning/pull/493/files#diff-ca1e86afc37a9b71183fb70c7fd2d65ea68a24a60e55a664d2f379be63fc7f9dR2336
the fail point doing nothing actually.
so the first run will succeed and the checkpoint is removed.

Screen Shot 2021-04-07 at 7 28 38 PM

tests/lightning_local_backend/run.sh
```diff
@@ -111,6 +114,9 @@ for ckpt in mysql file; do
 
   # after index engine is imported, local file could handle lost
   set +e
+  run_sql 'DROP DATABASE IF EXISTS cpeng;'
```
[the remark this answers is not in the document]
july2993: already remove this part

tests/lightning_local_backend/run.sh
july2993: if not remove it will fail to check the empty table since it's a new import.
I have no clue why the failpoint is changed to doing nothing so just keep clean up here like before now.

july2993:
@glorv @kennytm PTAL

glorv (commented):
/lgtm

3pointer (approved):

PR #986, opened 2021-04-07
pkg/lightning: only retry unfinished ranges when retry import engine

glorv:
In our multi-lightning tests, sometimes the import index engine phase encounters error retry after running for hours. In this condition, a lot of ranges have already finished ingesting, so we can avoid retrying these finished ranges in local backend when retry importing engine to make import retry faster.

- Save finished ranges into `LocalEngine` after ingest succeeded. And always calculate the retry ranged based on current finished ranges.

TODO:
Shall we also save the finished ranges info into pebble like the `engineMeta` as a checkpoint?

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - Only retry unfinished ranges when retry import full engine.


glorv:
@kennytm @Little-Wallace PTAL

pkg/lightning/backend/local/local.go
```diff
@@ -902,11 +903,19 @@ func splitRangeBySizeProps(fullRange Range, sizeProps *sizeProperties, sizeLimit
...
+		}
 		curSize += p.Size
 		curKeys += p.Keys
 		if int64(curSize) >= sizeLimit || int64(curKeys) >= keysLimit {
-			ranges = append(ranges, Range{start: curKey, end: p.Key})
-			curKey = p.Key
+			// in case the sizeLimit or keysLimit is too small
+			endKey := p.Key
+			if bytes.Equal(curKey, endKey) {
+				endKey = nextKey(endKey)
```
lance6716, line 915: could we just return

pkg/lightning/backend/local/local.go
```diff
@@ -1309,18 +1326,18 @@ func (local *local) writeAndIngestByRanges(ctx context.Context, engineFile *File
...
 
-func (r *syncdRanges) add(g Range) {
+func (r *syncedRanges) add(g Range) {
 	r.Lock()
 	r.ranges = append(r.ranges, g)
 	r.Unlock()
 }
 
-func (r *syncdRanges) take() []Range {
+func (r *syncedRanges) take() []Range {
```
lance6716, line 1340: not used now

pkg/lightning/backend/local/local.go
```diff
@@ -1380,25 +1405,95 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
...
+func (engine *File) unfinishedRanges(ranges []Range) []Range {
+	engine.finishedRanges.Lock()
+	defer engine.finishedRanges.Unlock()
+
+	engine.finishedRanges.ranges = dedupAndMergeRanges(engine.finishedRanges.ranges)
+
+	return filterOverlapRange(ranges, engine.finishedRanges.ranges)
+}
+
+func dedupAndMergeRanges(ranges []Range) []Range {
```
lance6716, line 1429: can it deduplicate, and handle one range totally covers another?

glorv: Yes，If several range overlap with each other, the result will be the smallest start and the largest end

lance6716: the "dedup" makes me think of "[[1,2], [1,2]] -> [[1,2]]" 😂

pkg/lightning/backend/local/local.go
glorv: If the region-split-size is set to a very small value that one kv can exceed this threshold, then the logic can exact cover this threshold, though such small threshold threshold is only used in tests

pkg/lightning/backend/local/local.go
glorv: The `TestDedupAndMergeRanges` unit test has test cases that covers this condition

glorv: OK, I rename the function name to `sortAndMergeRanges`😅

pkg/lightning/backend/local/local.go
```diff
@@ -854,19 +855,19 @@ func (local *local) WriteToTiKV(
 		zap.Int64("buf_size", bytesBuf.totalSize()),
 		zap.Stringer("takeTime", time.Since(begin)))
 
-	var remainRange *Range
+	finishedRange := regionRange
 	if iter.Valid() && iter.Next() {
 		firstKey := append([]byte{}, iter.Key()...)
-		remainRange = &Range{start: firstKey, end: regionRange.end}
+		finishedRange = Range{start: regionRange.start, end: firstKey}
```
Little-Wallace, line 861: Shall it be `regionRange` or `start` passed by method parameter?

glorv: The `start`, `end` passed from method parameter is the Range calculated by lightning, and it's range may covers multiple regions. The`regionRange` is the actual range the intersect  target region and [start, end). So the finishedRange's start should be the regionRange's start

pkg/lightning/backend/local/local.go
lance6716: I see, `return` will cause the range merged into next one.

lance6716 (approved):

PR #1003, opened 2021-04-13
log: delete lightning's redact package

lichunzhu:
lightning and br's redact package are duplicated.

replace lightning's redact package with br's.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release Note

 - No release note


pkg/lightning/backend/tidb/tidb.go
```diff
@@ -59,7 +60,7 @@ type tidbRows []tidbRow
 // MarshalLogArray implements the zapcore.ArrayMarshaler interface
 func (rows tidbRows) MarshalLogArray(encoder zapcore.ArrayEncoder) error {
 	for _, r := range rows {
-		encoder.AppendString(string(r))
+		encoder.AppendString(redact.String(string(r)))
```
kennytm, line 63: there's a slight difference between this and the original `ZapRedactArray`. 

when redaction is on, the marshal output in `ZapRedactArray` was always `?`, while here we get `?,?,?,?,?`, and thus "leaking" the number of columns. not sure if it is considered sensitive though.

pkg/logutil/logging.go
```diff
@@ -186,3 +186,11 @@ func Keys(keys [][]byte) zap.Field {
 func ShortError(err error) zap.Field {
 	return zap.String("error", err.Error())
 }
+
+// Any constructs a field that carries an interface{}.
```
kennytm, line 190:
```suggestion
// Any constructs a redacted field that carries an interface{}.
```

maybe name it `RedactAny`.

kennytm (commented):
rest LGTM

pkg/lightning/backend/tidb/tidb.go
lichunzhu: Table structure and number of columns seem not sensitive. PTAL @overvenus

PR #1014, opened 2021-04-14
test: fix the unstable unit test 

3pointer:
TestLightning `Runserver` checks error is `nil` in another goroutine.and the result is unstable because Teardown(`ctx.cancel()`) may happen before `Runserver` exit.

just like https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/9631/pipeline

remove this check. because the error is either `nil` or `context.cancel`.

### Check List 

Tests 

 - Unit test

Related changes

### Release Note

 - No release note.

### DIFF
--- a/pkg/lightning/lightning_test.go
+++ b/pkg/lightning/lightning_test.go
@@ -143,7 +143,7 @@ func (s *lightningServerSuite) TestRunServer(c *C) {
 	resp.Body.Close()
 
 	go func() {
-		s.lightning.RunServer()
+		_ = s.lightning.RunServer()
 	}()
 	time.Sleep(100 * time.Millisecond)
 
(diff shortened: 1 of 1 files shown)


pkg/lightning/lightning_test.go
```diff
@@ -143,8 +143,7 @@ func (s *lightningServerSuite) TestRunServer(c *C) {
 	resp.Body.Close()
 
 	go func() {
-		err := s.lightning.RunServer()
-		c.Assert(err, IsNil)
+		s.lightning.RunServer()
```
YuJuncen, line 146:
```suggestion
		_ = s.lightning.RunServer()
```

Or the linter may get unhappy...

PR #1020, opened 2021-04-15
lightning: optimize lightning memory and perf

glorv:
Optimize lightning performance with better memory allocation.

- Reuse the bytes buffer in session encode 
- Avoid doing sort.Slice if local-writer's `writeBatch` is already sorted
- Pre-allocate a larger writeBatch to avoid slice reallocate too many times.

Benchmark result:
Each source file in the following two dataset are about 256MiB in size, and the target table contains 4 indexes(1 int primary key and 3 secondary keys).
| branch | data size | cost |
| --- | --- | --- |
| master([921e0da](https://github.com/pingcap/br/commit/921e0da638d880498e93afb0b945d72f84049115)) | 100GiB | 16m48s |
| opt-local-writer | 100GiB | 13m3s |
| this pr | 100GiB | 11m37s |
| opt-local-writer | 500GiB | 1h2m | 
| this pr | 500GiB | 56m40s | 

This PR is based on #753.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - Optimize local writer memory allocation.


pkg/lightning/backend/kv/sql2kv.go
```diff
@@ -405,62 +410,82 @@ func (kvcodec *tableKVEncoder) Encode(
...
-			indexKVs = append(indexKVs, kv)
+			indexKVs.pairs = append(indexKVs.pairs, kv)
 			indexChecksum.UpdateOne(kv)
 		}
 	}
 
+	// the related buf is shared, so we only need to set it into one of the kvs so it can be released
+	if kvs.bytesBuf != nil {
+		dataKVs.bytesBuf = kvs.bytesBuf
+		dataKVs.memBuf = kvs.memBuf
```
Little-Wallace, line 446: Shall we set `kvs.bytesBuf = nil; kvs.memBuf = nil;` ? so that the reference of buf can be release.

pkg/lightning/backend/kv/session.go
```diff
@@ -47,18 +51,77 @@ func (*invalidIterator) Valid() bool {
...
-		Val: append([]byte{}, v...),
+	kvPairs := mb.kvPairs
+	size := len(k) + len(v)
+	if mb.buf == nil || mb.buf.cap-mb.buf.idx < size {
+		if mb.buf != nil {
+			kvPairs.bytesBuf = mb.buf
+		}
+		mb.AllocateBuf(size)
+	}
+	kvPairs.pairs = append(kvPairs.pairs, common.KvPair{
```
Little-Wallace, line 120: Do we need to catch `bytesBuf` to common.KvPair when `buf` is enough?

Little-Wallace: If the current `buf` is not enough large, it seem that we discard it directly.  Would it be wasteful? I think we shall split them into different kind of buf. such as `32B, 128B, 512B,  1KB,  2KB, 4KB, 8KB,  16KB, 32KB, 64KB` and so on.

Little-Wallace:
rest LGTM

pkg/lightning/backend/kv/session.go
glorv: > Do we need to catch bytesBuf to common.KvPair when buf is enough?

No. Only the last KvPairs that reference to the `buf` should take the responsibility to recycle the buffer because only this message are consumed, then the buffer is not referenced anymore and can be recycled.

> If the current buf is not enough large, it seem that we discard it directly. Would it be wasteful? I think we shall split them into different kind of buf. such as 32B, 128B, 512B, 1KB, 2KB, 4KB, 8KB, 16KB, 32KB, 64KB and so on.

yes, the current logic is to simply design. Since the default buf size is 1MiB, and 
message batch size is 64KiB, it should be large enough for most case, so in most case it should be ok. And since go is a GC language , we can depend on go GC and memory allocator to help us manage the buf allocation strategy to avoid too complicated memory management.

pkg/lightning/backend/kv/sql2kv.go
glorv: Done. In theory go GC should properly handle this but I still add the set nil action since it's not harmful and maybe help go GC better handle some situation because we depend on CGO to manage memory.

Little-Wallace (approved):
LGTM

3pointer (approved):

PR #1036, opened 2021-04-19
backup: debounce fine-grained backup

YuJuncen:
Sometimes, TiKV make empty response stream from backing up request. (This seems buggy, but not sure.)

In this condition, fine-grained backup would crazily send `backup` RPC, if there are enough resource, thing wouldn't be so bad: when PD make progress, real leader can be found, and everything should back to normal, but for our CI, the busiest node of our community, un, ya see. 

*Note the timestamp. There are about 100k+ or even more logs like this.*

image

Now, if fine-grained backup make no progress, it would backoff for 10 seconds.
The backoff time is an experience value, by analyzing the local log. Let's change it if necessary.

### Check List 

Tests 

### Release Note

 - Fix a bug that caused BR send too many useless RPCs to TiKV.


pkg/backup/client.go
```diff
@@ -866,7 +870,14 @@ func (bc *Client) handleFineGrained(
 		return 0, errors.Annotatef(err, "failed to send fine-grained backup [%s, %s)",
 			redact.Key(req.StartKey), redact.Key(req.EndKey))
 	}
-	return max, nil
+
+	// If no progress, backoff 10s for debouncing.
+	// 10s is the default interval of stores sending a heartbeat to the PD.
+	// And is the average new leader election timeout, which would be a reasonable back off time.
+	if !hasProgress && backoffMill < 10000 {
```
3pointer, line 877: `if !hasProgress`

Issue #1042, opened 2021-04-20
Backup too slow when backup lots of tables


3pointer:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
BR Backup a large number(60K) of tables (with 400K regions)

2. What did you expect to see?
Backup finished within a meaningful time.

3. What did you see instead?
Some prepare work costs too much time when BR has lots of tables to backup. such as  [getRegionCount](https://github.com/pingcap/br/blob/master/pkg/task/backup.go#L343). we'd better find another way to get region count.

4. What version of BR and TiDB/TiKV/PD are you using?
master

PR #1047, opened 2021-04-21
backup: optimize progress update logic

3pointer:
Solve https://github.com/pingcap/br/issues/1042

if backup ranges is more than 100, we can use range as the unit of progress.

### Check List 

Tests 

Related changes

### Release Note

 - Fix the issue that backup costs too much time to count progress units when backup lots of ranges.

### DIFF
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -71,9 +71,9 @@ const (
 	backupFineGrainedMaxBackoff = 80000
 	backupRetryTimes            = 5
 	// RangeUnit represents the progress updated counter when a range finished.
-	RangeUint ProgressUnit = "range"
+	RangeUnit ProgressUnit = "range"
 	// RegionUnit represents the progress updated counter when a region finished.
-	RegionUint ProgressUnit = "region"
+	RegionUnit ProgressUnit = "region"
 )
 
 // Client is a client instructs TiKV how to do a backup.
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -354,10 +354,10 @@ func BuildBackupRangeAndSchema(
 					return nil, nil, errors.Trace(err)
 				}
 				tableInfo.AutoRandID = globalAutoRandID
-				logger.Info("change table AutoRandID",
+				logger.Debug("change table AutoRandID",
 					zap.Int64("AutoRandID", globalAutoRandID))
 			}
-			logger.Info("change table AutoIncID",
+			logger.Debug("change table AutoIncID",
 				zap.Int64("AutoIncID", globalAutoID))
 
 			// remove all non-public indices
--- a/pkg/backup/client.go
+++ b/pkg/backup/client.go
@@ -39,7 +39,6 @@ import (
 
 	"github.com/pingcap/br/pkg/conn"
 	berrors "github.com/pingcap/br/pkg/errors"
-	"github.com/pingcap/br/pkg/glue"
 	"github.com/pingcap/br/pkg/logutil"
 	"github.com/pingcap/br/pkg/redact"
 	"github.com/pingcap/br/pkg/rtree"
@@ -433,7 +432,7 @@ func (bc *Client) BackupRanges(
 	ranges []rtree.Range,
 	req backuppb.BackupRequest,
 	concurrency uint,
-	updateCh glue.Progress,
+	progressCallBack func(bool),
 ) ([]*backuppb.File, error) {
 	if span := opentracing.SpanFromContext(ctx); span != nil && span.Tracer() != nil {
 		span1 := span.Tracer().StartSpan("Client.BackupRanges", opentracing.ChildOf(span.Context()))
@@ -467,7 +466,7 @@ func (bc *Client) BackupRanges(
 		for _, r := range ranges {
 			sk, ek := r.StartKey, r.EndKey
 			workerPool.ApplyOnErrorGroup(eg, func() error {
-				files, err := bc.BackupRange(ectx, sk, ek, req, updateCh)
+				files, err := bc.BackupRange(ectx, sk, ek, req, progressCallBack)
 				if err == nil {
 					filesCh <- files
 				}
@@ -501,7 +500,7 @@ func (bc *Client) BackupRange(
 	ctx context.Context,
 	startKey, endKey []byte,
 	req backuppb.BackupRequest,
-	updateCh glue.Progress,
+	progressCallBack func(bool),
 ) (files []*backuppb.File, err error) {
 	start := time.Now()
 	defer func() {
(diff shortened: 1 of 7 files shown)


pkg/backup/client.go
```diff
@@ -433,7 +432,7 @@ func (bc *Client) BackupRanges(
 	ranges []rtree.Range,
 	req backuppb.BackupRequest,
 	concurrency uint,
-	updateCh glue.Progress,
+	progressCallBack func(bool),
```
kennytm, line 435: please use an enum instead of a bool here

3pointer: sure, I'll change it

pkg/task/backup.go
```diff
@@ -336,30 +336,52 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
...
-	// The number of regions need to backup
-	approximateRegions := 0
-	for _, r := range ranges {
-		var regionCount int
-		regionCount, err = mgr.GetRegionCount(ctx, r.StartKey, r.EndKey)
-		if err != nil {
-			return errors.Trace(err)
+	summary.CollectInt("backup total ranges", len(ranges))
+
+	var RangeUpdateCh glue.Progress
```
YuJuncen, line 341: Eh, why use PascalCase here...?

3pointer: fixed...

pkg/backup/client.go
```diff
@@ -69,6 +69,16 @@ const (
...
 )
 
+// ProgressUnit represents the unit of progress.
+type ProgressUnit string
+
+const (
+	// RangeUnit represents the progress updated counter when a range finished.
+	RangeUint ProgressUnit = "range"
+	// RegionUnit represents the progress updated counter when a region finished.
+	RegionUint ProgressUnit = "region"
```
kennytm, line 76:
```suggestion
	RangeUnit ProgressUnit = "range"
	// RegionUnit represents the progress updated counter when a region finished.
	RegionUnit ProgressUnit = "region"
```

tests/br_300_small_tables/run.sh
```diff
@@ -29,7 +32,22 @@ done
 
 # backup db
 echo "backup start..."
+
+export GO_FAILPOINTS="github.com/pingcap/br/pkg/task/progress-call-back=return(\"$PROGRESS_FILE\")"
 run_br backup db --db "$DB" -s "local://$TEST_DIR/$DB" --pd $PD_ADDR
+export GO_FAILPOINTS=""
+
+# check if we have 300 range, because we have 300 tables
+if [[ "$(wc -l <$PROGRESS_FILE)" == "1" ]] && [[ $(grep -q "range:300" $PROGRESS_FILE) ]];
```
YuJuncen, line 41: Seems the return code of `grep` is enough. And seems each table has an index...?

```suggestion
if [[ "$(wc -l <$PROGRESS_FILE)" == "1" ]] && grep -q "range:600" $PROGRESS_FILE;
```

3pointer: ok, I removed the number ...

Consider PKIsHandle in release-4.0, if we compare the number, it will make the case more complicated...

YuJuncen: Yep, and seems clustered indices also influence this...

3pointer:
[2021-04-23T10:01:26.876Z] ERROR 2003 (HY000): Can't connect to MySQL server on '127.0.0.1' (111)

/run-all-tests

tests/br_db/run.sh
```diff
@@ -36,7 +39,20 @@ run_sql "CREATE TABLE $DB.usertable2 ( \
 run_sql "INSERT INTO $DB.usertable2 VALUES (\"c\", \"d\");"
 # backup db
 echo "backup start..."
+export GO_FAILPOINTS="github.com/pingcap/br/pkg/task/progress-call-back=return(\"$PROGRESS_FILE\")"
 run_br --pd $PD_ADDR backup db --db "$DB" -s "local://$TEST_DIR/$DB" --ratelimit 5 --concurrency 4
+export GO_FAILPOINTS=""
+
+# check if we use the region unit
+if [[ "$(wc -l <$PROGRESS_FILE)" == "1" ]] && [[ !$(grep -q "region" $PROGRESS_FILE) ]];
```
YuJuncen, line 47: I'm afraid `[[ !$(grep -q "region" $PROGRESS_FILE) ]]` would be `true` in any condition:

Firstly, the expression would be evaluated to: `[[ ! ]]`, since `grep -q` wouldn't print anything to stdout. And its exit code would be ignored in this context.

Then, with single argument, `[[` would test whether the argument is empty string(i.e. `[[ -n ! ]]`), and '!' isn't empty, so the result of test would always be `true`.

pkg/task/backup.go
```diff
@@ -336,30 +339,78 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
...
-	// The number of regions need to backup
-	approximateRegions := 0
-	for _, r := range ranges {
-		var regionCount int
-		regionCount, err = mgr.GetRegionCount(ctx, r.StartKey, r.EndKey)
-		if err != nil {
-			return errors.Trace(err)
+	summary.CollectInt("backup total ranges", len(ranges))
+
+	var rangeUpdateCh glue.Progress
```
glorv, line 344: I think we can still only use one `glue.Progress` since only one type of msg is actually sent.

tests/br_db/run.sh
3pointer: Ok, fixed.

Now it can be tested. https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/10502/pipeline

glorv:
Shall we provide a benchmark result? 🤔

rest LGTM

YuJuncen:
> Shall we provide a benchmark result? 🤔
> 
> rest LGTM

Maybe run `br_300_small_tables` both in master and this PR, then record the time cost is enough?

3pointer:
> Shall we provide a benchmark result? 🤔
> 
> rest LGTM

I tested 10k tables via 4.0.9 cluster.

it costs 3 minutes before backup.
```
[2021/04/25 16:58:15.957 +08:00] [INFO] [collector.go:60] ["Database backup Success summary: total backup ranges: 20000, total success: 20000, total failed: 0, total take(Database backup time): 8.179047095s, total take(real time): 3m24.428890176s, total kv: 200000, total size(MB): 23.94, avg speed(MB/s): 2.93"] ["backup checksum"=574.461008ms] ["backup fast checksum"=17.282718266s] ["backup total regions"=20000] [BackupTS=424500498946588673] [Size=76490910]
```

and with this PR it costs only a few seconds before backup.
```
[2021/04/25 16:37:15.408 +08:00] [INFO] [collector.go:62] ["Database backup success summary"] [total-ranges=20000] [ranges-succeed=20000] [ranges-failed=0] [backup-checksum=597.893735ms] [backup-fast-checksum=3.08850778s] [backup-total-ranges=20000] [BackupTS=424500216892751874] [Size=72032247] [total-take=18.761364956s] [total-kv=200000] [data-size=25.1MB] [average-speed=2.927MB/s]
```

Issue #872, opened 2021-03-15
br  can not  backup and resotore tidb   system databases  like mysql.user


wangbinhe3db:
## Question
i want to backup and restore all tidb data ,  but  br can not   backup and resotore tidb system databases 
and i see this in code 
![image](https://user-images.githubusercontent.com/78577992/111099647-ea692700-8580-11eb-9367-e6dd3b98bda0.png)
br  can not  backup and restore tidb   system databases  like mysql.user
i want to know the reason ?

wangbinhe3db:
\cc  @kennytm

kennytm:
@wangbinhe3db we explicitly opt out of system tables. Because if the user wrongly restore on a cluster with different schema version, BR can easily brick the entire cluster. See #679 for the intended direction how we'd like to handle system tables.

SunRunAway:
1. `mysql.user` may have the account information which customer may forget in their brain.
2. TiDB allow a customer to create a table in the `mysql` schema.

I suggest we should backup the system tables, and can opt out when restoring.

@kennytm @IANTHEREAL

kennytm:
@SunRunAway please discuss the plan about system tables in #679. i'm closing this issue as duplicate.

Issue #679, opened 2021-01-06
BR failed to backup stats


3pointer:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
Upgrade TiDB cluster from v4.0.0 to v4.0.9. then use BR v4.0.9 to backup.

2. What did you expect to see?
backup success.

3. What did you see instead?
```
[ERROR] [backup.go:25] ["failed to backup"] [error="[types:1406]Data Too Long, field len 5, data len 6"] [errorVerbose="[types:1406]Data Too Long, field len 5, data len 6\n
github.com/pingcap/errors.AddStack\n
\t	github.com/pingcap/errors@v0.11.5-0.20201126102027-b0a155152ca3/errors.go:174\n
github.com/pingcap/errors.(*Error).GenWithStack\n
\t	github.com/pingcap/errors@v0.11.5-0.20201126102027-b0a155152ca3/normalize.go:147\n
github.com/pingcap/tidb/types.ProduceStrWithSpecifiedTp\n
\t	github.com/pingcap/tidb@v1.1.0-beta.0.20201214152324-ce2f365189d3/types/datum.go:995\n
github.com/pingcap/tidb/types.(*Datum).convertToString\n
\t	github.com/pingcap/tidb@v1.1.0-beta.0.20201214152324-ce2f365189d3/types/datum.go:942\n
github.com/pingcap/tidb/types.(*Datum).ConvertTo\n
\t	github.com/pingcap/tidb@v1.1.0-beta.0.20201214152324-ce2f365189d3/types/datum.go:821\n
github.com/pingcap/tidb/statistics/handle.(*Handle).histogramFromStorage\n
\t	github.com/pingcap/tidb@v1.1.0-beta.0.20201214152324-ce2f365189d3/statistics/handle/handle.go:667\n
github.com/pingcap/tidb/statistics/handle.(*Handle).columnStatsFromStorage\n
\t	github.com/pingcap/tidb@v1.1.0-beta.0.20201214152324-ce2f365189d3/statistics/handle/handle.go:460\n
github.com/pingcap/tidb/statistics/handle.(*Handle).tableStatsFromStorage\n
\t	github.com/pingcap/tidb@v1.1.0-beta.0.20201214152324-ce2f365189d3/statistics/handle/handle.go:539\n
github.com/pingcap/tidb/statistics/handle.(*Handle).tableStatsToJSON\n
[...]

PR #1048, opened 2021-04-22
backup,restore: support backing up / restore system databases

YuJuncen:
#872 and #679.

But this doesn't allow restore stats -- because the table IDs of the stats backed up are old table IDs, we need more efforts to rewrite them to make them take effect after the restoration.

This PR basically implemented https://github.com/pingcap/br/issues/679#issuecomment-762592254, with some difference:

- When the target table not exists, `RENAME TABLE temp.xxx TO mysql.xxx` would be used, for effectively restoring user tables created in `mysql` schema.
- The user interface is the `-f` (the table filter) flag. Users must specify each table they want to restore by the table filter syntax (e.g. `br restore full -f '*.*' -f '!mysql.*' -f 'mysql.usertable'` to restore `mysql.usertable`).

### Check List 

Tests 

 - Integration test

Related changes

    - The document points out that table filter would always ignore system tables, that isn't real after this merged.
    - The usage of this feature.

### Release Note

 - BR now support backing up user tables created in the `mysql` schema.

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -21,7 +21,6 @@ var (
 	ErrInvalidArgument = errors.Normalize("invalid argument", errors.RFCCodeText("BR:Common:ErrInvalidArgument"))
 	ErrVersionMismatch = errors.Normalize("version mismatch", errors.RFCCodeText("BR:Common:ErrVersionMismatch"))
 	ErrFailedToConnect = errors.Normalize("failed to make gRPC channels", errors.RFCCodeText("BR:Common:ErrFailedToConnect"))
-	ErrFailedToExecute = errors.Normalize("failed to execute SQL", errors.RFCCodeText("BR:Common:ErrFailedToExecute"))
 
 	ErrPDUpdateFailed    = errors.Normalize("failed to update PD", errors.RFCCodeText("BR:PD:ErrPDUpdateFailed"))
 	ErrPDLeaderNotFound  = errors.Normalize("PD leader not found", errors.RFCCodeText("BR:PD:ErrPDLeaderNotFound"))
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -43,7 +43,7 @@ var (
 	ErrRestoreInvalidRange     = errors.Normalize("invalid restore range", errors.RFCCodeText("BR:Restore:ErrRestoreInvalidRange"))
 	ErrRestoreWriteAndIngest   = errors.Normalize("failed to write and ingest", errors.RFCCodeText("BR:Restore:ErrRestoreWriteAndIngest"))
 	ErrRestoreSchemaNotExists  = errors.Normalize("schema not exists", errors.RFCCodeText("BR:Restore:ErrRestoreSchemaNotExists"))
-	ErrUnsupportedSystemTable  = errors.Normalize("the system table isn't supported to restore here", errors.RFCCodeText("BR:Restore:ErrUnsupportedSysTable"))
+	ErrUnsupportedSystemTable  = errors.Normalize("the system table isn't supported for restoring yet", errors.RFCCodeText("BR:Restore:ErrUnsupportedSysTable"))
 
 	// TODO maybe it belongs to PiTR.
 	ErrRestoreRTsConstrain = errors.Normalize("resolved ts constrain violation", errors.RFCCodeText("BR:Restore:ErrRestoreResolvedTsConstrain"))
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -21,6 +21,7 @@ var (
 	ErrInvalidArgument = errors.Normalize("invalid argument", errors.RFCCodeText("BR:Common:ErrInvalidArgument"))
 	ErrVersionMismatch = errors.Normalize("version mismatch", errors.RFCCodeText("BR:Common:ErrVersionMismatch"))
 	ErrFailedToConnect = errors.Normalize("failed to make gRPC channels", errors.RFCCodeText("BR:Common:ErrFailedToConnect"))
+	ErrFailedToExecute = errors.Normalize("failed to execute SQL", errors.RFCCodeText("BR:Common:ErrFailedToExecute"))
 
 	ErrPDUpdateFailed    = errors.Normalize("failed to update PD", errors.RFCCodeText("BR:PD:ErrPDUpdateFailed"))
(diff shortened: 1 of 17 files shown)


pkg/restore/systable_restore.go
```diff
@@ -0,0 +1,163 @@
...
+	for _, table := range originDatabase.Tables {
+		tableName := table.Info.Name
+		if f.MatchTable(sysDB, tableName.O) {
+			rc.replaceTemporaryTableToSystable(ctx, tableName.L, db)
+		}
+		tablesRestored = append(tablesRestored, tableName.L)
+	}
+	if err := rc.afterSystemTablesReplaced(ctx, tablesRestored); err != nil {
+		for _, e := range multierr.Errors(err) {
+			logutil.WarnTerm("error during winding up the restoration of system table", zap.String("database", sysDB), logutil.ShortError(e))
```
kennytm, line 57: what is meant by "wind up the restoration" 🤔

kennytm (commented):
rest LGTM

pkg/restore/systable_restore.go
YuJuncen: > From New Oxford American Dictionary, `wind up'.
> gradually or finally bring an activity to a conclusion: *the experiments had to be wound up because the funding stopped.*

I was trying to find a word to describe the works for making the system config active after replacing tuples into system tables, but seems this verb is too exotic in some way... 🥲 Maybe just "activating new config" could be fine.

kennytm: or maybe just "error reconfigurating the system tables"

pkg/restore/systable_restore.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	if needFlushStat {
+		// The newly created tables have different table IDs with original tables,
+		// 	hence the old statistics are invalid.
+		//
+		// TODO:
+		// 	Plan A) rewrite the IDs in the `rc.statsHandler.Update(rc.dom.InfoSchema())`.
+		//  Plan B) give the user a SQL to let him update it manually.
+		//  If plan A applied, we can deprecate the origin interface for backing up statistics.
+		err = multierr.Append(err, errors.Annotatef(berrors.ErrUnsupportedSystemTable,
```
kennytm, line 113: hold on, this error cannot be emitted *after* that `REPLACE INTO` stuff has taken place, this will break the stats of other existing tables.

YuJuncen: Yep. 🤔

YuJuncen: Done

pkg/errors/errors.go
```diff
@@ -21,6 +21,7 @@ var (
 	ErrInvalidArgument = errors.Normalize("invalid argument", errors.RFCCodeText("BR:Common:ErrInvalidArgument"))
 	ErrVersionMismatch = errors.Normalize("version mismatch", errors.RFCCodeText("BR:Common:ErrVersionMismatch"))
 	ErrFailedToConnect = errors.Normalize("failed to make gRPC channels", errors.RFCCodeText("BR:Common:ErrFailedToConnect"))
+	ErrFailedToExecute = errors.Normalize("failed to execute SQL", errors.RFCCodeText("BR:Common:ErrFailedToExecute"))
```
kennytm, line 24: given that the SQL can be anything this error looks no better than ErrUnknown 🙃

YuJuncen: Removed 🤔

YuJuncen:
`ERROR 2003 (HY000): Can't connect to MySQL server on '127.0.0.1' (111)` 🤔

/run-integration-tests

SunRunAway:
Will global variables can be restored after this PR?

YuJuncen:
> Will global variables can be restored after this PR?

Not tested, but in theory they could.

SunRunAway:
> > Will global variables can be restored after this PR?
> 
> Not tested, but in theory they could.

Should you add this into the objective and and let the testing cover it?

SunRunAway:
> > Will global variables can be restored after this PR?
> 
> Not tested, but in theory they could.

@IANTHEREAL Please make it clear to customers.

YuJuncen:
> > > Will global variables can be restored after this PR?
> > 
> > 
> > Not tested, but in theory they could.
> 
> Should you add this into the objective and and let the testing cover it?

Done, a single value test shows that GLOBAL_VARIABLES can be restored. However, note this PR is a workaround for backup / restoring user tables created in `mysql`. I think it would be better to not promise any ability of backup / restoring system tables.

pkg/restore/systable_restore.go
```diff
@@ -0,0 +1,185 @@
...
+	return ok
+}
+
+// RestoreSystemSchemas restores the system schema(i.e. the `mysql` schema).
+// Detail see https://github.com/pingcap/br/issues/679#issuecomment-762592254.
+func (rc *Client) RestoreSystemSchemas(ctx context.Context, f filter.Filter) {
+	sysDB := mysql.SystemDB
+
+	temporaryDB := utils.TemporaryDBName(sysDB)
+	defer rc.cleanTemporaryDatabase(ctx, sysDB)
```
3pointer, line 43: Did this `cleanTemporaryDatabase` happen even we failed to replaceSystemTable?

SunRunAway:
> > > > Will global variables can be restored after this PR?
> 
> > > 
> 
> > > 
> 
> > > Not tested, but in theory they could.
> 
> > 
> 
> > Should you add this into the objective and and let the testing cover it?
> 
> 
> 
> Done, a single value test shows that GLOBAL_VARIABLES can be restored. However, note this PR is a workaround for backup / restoring user tables created in `mysql`. I think it would be better to not promise any ability of backup / restoring system tables.

We should make it clear. Because once you released it, users will rely on this feature.

YuJuncen:
@SunRunAway I added a warning over restoring system tables in the [document](https://github.com/pingcap/docs-cn/pull/6194/files#diff-443aab9d2cb868c0aaf5de811b5f95f770b22d6994579ce4b9ed21d8c433918dR397-R404).

cc @3pointer Please check whether it is necessary.

IANTHEREAL:
If the system table cannot be restored correctly (referring to the system table similar to mysql.tidb, not the user data table under the mysql database), then there is no need to provide recovery methods to avoid misuse by users and affect their business

SunRunAway:
If the system table cannot be restored correctly, why you backup them, why you implement this feature by this pull request?

PR #1066, opened 2021-04-27
add rate limit for local backend, parse limit args

recall704:

add rate limit for local backend

set rate limit for  `io.Writer`

### DIFF
--- a/go.mod1
+++ b/go.mod1
@@ -54,4 +54,7 @@ require (
 	sourcegraph.com/sourcegraph/appdash v0.0.0-20190731080439-ebfcffb1b5c0
 )
 
-replace cloud.google.com/go/storage => github.com/3pointer/google-cloud-go/storage v1.6.1-0.20210108125931-b59bfa0720b2
+replace (
+	cloud.google.com/go/storage => github.com/3pointer/google-cloud-go/storage v1.6.1-0.20210108125931-b59bfa0720b2
+	github.com/pingcap/kvproto v0.0.0-20210308063835-39b884695fb8 => github.com/recall704/kvproto v0.0.0-20210414071537-7bdfcba5f5d5
+)
--- a/pkg/storage/flags.go
+++ b/pkg/storage/flags.go
@@ -9,12 +9,16 @@ import (
 
 // DefineFlags adds flags to the flag set corresponding to all backend options.
 func DefineFlags(flags *pflag.FlagSet) {
+	DefineLocalFlags(flags)
 	defineS3Flags(flags)
 	defineGCSFlags(flags)
 }
 
 // ParseFromFlags obtains the backend options from the flag set.
 func (options *BackendOptions) ParseFromFlags(flags *pflag.FlagSet) error {
+	if err := options.Local.parseFromFlags(flags); err != nil {
+		return errors.Trace(err)
+	}
 	if err := options.S3.parseFromFlags(flags); err != nil {
 		return errors.Trace(err)
 	}
--- a/pkg/storage/local.go
+++ b/pkg/storage/local.go
@@ -9,7 +9,11 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/fujiwara/shapeio"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/kvproto/pkg/backup"
+	backuppb "github.com/pingcap/kvproto/pkg/backup"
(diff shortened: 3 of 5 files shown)


go.mod1
```diff
@@ -54,4 +54,7 @@ require (
 	sourcegraph.com/sourcegraph/appdash v0.0.0-20190731080439-ebfcffb1b5c0
 )
 
-replace cloud.google.com/go/storage => github.com/3pointer/google-cloud-go/storage v1.6.1-0.20210108125931-b59bfa0720b2
+replace (
+	cloud.google.com/go/storage => github.com/3pointer/google-cloud-go/storage v1.6.1-0.20210108125931-b59bfa0720b2
+	github.com/pingcap/kvproto v0.0.0-20210308063835-39b884695fb8 => github.com/recall704/kvproto v0.0.0-20210414071537-7bdfcba5f5d5
```
recall704, line 59: https://github.com/pingcap/kvproto/pull/756 PR required

kennytm:
Instead of adding a `rate_limit` field to local storage only, I suggest adapting to the existing `--ratelimit` parameter.

* Not good: `-o 'local:///data/output?rate_limit=100'`
* Good: `--ratelimit 100 -o 'local:///data/output'`

The rate limiter should introduce a wrapper of `ExternalStorage` similar to the `withCompression` type.

recall704:
> Instead of adding a `rate_limit` field to local storage only, I suggest adapting to the existing `--ratelimit` parameter.
> 
> * Not good: `-o 'local:///data/output?rate_limit=100'`
> * Good: `--ratelimit 100 -o 'local:///data/output'`
> 
> The rate limiter should introduce a wrapper of `ExternalStorage` similar to the `withCompression` type.

```go

type withRatelimit struct {
	ExternalStorage
	Ratelimit uint64 // Byte/sec
}

// WithRatelimit
func WithRatelimit(inner ExternalStorage, ratelimit uint64) ExternalStorage {
	if ratelimit == 0 {
		return inner
	}
	return &withRatelimit{ExternalStorage: inner, Ratelimit: ratelimit}
}

func (r *withRatelimit) Create(ctx context.Context, name string) (ExternalFileWriter, error) {
	var (
		writer ExternalFileWriter
		err    error
	)
	if localStorage, ok := r.ExternalStorage.(*LocalStorage); ok {
		file, err := os.Create(filepath.Join(localStorage.base, name))
		if err != nil {
			return nil, errors.Trace(err)
		}
		buf := bufio.NewWriter(file)
		if r.Ratelimit > 0 {
			w := shapeio.NewWriterWithContext(buf, ctx)
			w.SetRateLimit(float64(r.Ratelimit))
			return newFlushStorageWriter(w, buf, file), nil
		}
		return newFlushStorageWriter(buf, buf, file), nil
	}
	writer, err = r.ExternalStorage.Create(ctx, name)
	if err != nil {
		return nil, errors.Trace(err)
	}
	return writer, nil
}
```

I try to use `withRatelimit` to set rate limit, but I have to overwrite `Create` method, any idea?

thanks

kennytm:
@recall704 sorry for late reply, there's a week-long holiday in China and Japan 🙃.

> I try to use withRatelimit to set rate limit, but I have to overwrite Create method, any idea?

you could just return a wrapper type that implements `ExternalFileWriter`:

```go
type withRateLimitWriter struct {
    ExternalFileWriter
    Ratelimit uint64
}

func (r *withRatelimit) Create(ctx context.Context, name string) (ExternalFileWriter, error) {
    inner, err := r.ExternalStorage.Create(ctx, name)
    if err != nil {
        return nil, err
    }
    return &withRateLimitWriter{
        ExternalFileWriter: inner,
        Ratelimit:          r.Ratelimit,
    }, nil
}
```

and then apply the Ratelimit in its `Write` method.

```go
func (rw *withRateLimitWriter) Write(ctx context.Context, p []byte) (int, error) {
    // do rate limiting here.
}
```

There's no need to figure out how to fit this into `shapeio`'s Reader/Writer API, since [that package](https://github.com/fujiwara/shapeio/blob/v1.0.0/shapeio.go) is just using [`golang.org/x/time/rate`](https://pkg.go.dev/golang.org/x/time/rate) anyway. I suggest you use `rate.Limiter` directly instead of `shapeio` (allowing multiple writers from `Create` sharing the same limiter as a global rate limit).

PR #1068, opened 2021-04-28
err_msg: clarify some ambiguous error message

zwj-coder:
Some error message in BR is ambiguous and misleading. i.e. if tikv has no right to read from & write to backup storage, error message cannot clarify whether error occur on br side or tikv side. Users may think why I got "permission denied"? I'm running BR as root.

Some error messages are changes, and will clarify what is going on when error occur.

### Check List 

Tests 

 - Integration test

Code changes

### Release Note

 - No realease note

### DIFF
--- a/errors.toml
+++ b/errors.toml
@@ -98,7 +98,7 @@ rewrite rule not found
 
 ["BR:KV:ErrKVUnknown"]
 error = '''
-unknown tikv error
+unknown error occur on tikv, please ensure tikv is permitted to read from & write to the storage.
 '''
 
 ["BR:KV:ErrNotTiKVStorage"]
--- a/tests/br_full/run.sh
+++ b/tests/br_full/run.sh
@@ -53,6 +53,25 @@ run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB-lz4" --concurrency 4 
 export GO_FAILPOINTS=""
 size_lz4=$(du -d 0 $TEST_DIR/$DB-lz4 | awk '{print $1}')
 
+echo "backup with tikv error start..." 
+export GO_FAILPOINTS="github.com/pingcap/br/pkg/backup/tikv-rw-error=return(\"tikv read or write error\")"
+error_msg=$(run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB-tikverr" --concurrency 4 |  awk '{match($0,/.+(\[BR:KV:ErrKVUnknown\]unknown error occur on tikv, please ensure tikv is permitted to read from & write to the storage.)/,a);if (a[1] != "") print a[1]}')
+if [ "$error_msg" = "" ]
+then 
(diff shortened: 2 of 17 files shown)


errors.toml
```diff
@@ -98,7 +98,7 @@ rewrite rule not found
 
 ["BR:KV:ErrKVUnknown"]
 error = '''
-unknown tikv error
+unknown error occur on tikv, please ensure tikv is permitted to read from & write to backup storage.
```
YuJuncen, line 101: Seems the error can occur in restore too, maybe 'the storage' would be enough.

(BTW, I'm afraid other errors might be treated as `BR:KV:ErrKVUnknown` too, at such condition, maybe this message becomes misleading. Thankfully, seems currently there isn't.)

tests/br_full/run.sh
```diff
@@ -53,6 +53,17 @@ run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB-lz4" --concurrency 4
 export GO_FAILPOINTS=""
 size_lz4=$(du -d 0 $TEST_DIR/$DB-lz4 | awk '{print $1}')
 
+echo "backup with tikv error start..." 
+export GO_FAILPOINTS="github.com/pingcap/br/pkg/backup/tikv-rw-error=return(\"tikv read or write error\")"
+run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB-tikverr" --concurrency 4 &
+pid=$!
+export GO_FAILPOINTS=""
+sleep 15
```
YuJuncen, line 61: Consider add a check for whether the new error message was printed here?

YuJuncen (commented):
rest lgtm

tests/br_full/run.sh
zwj-coder: ok

tests/br_full/run.sh
```diff
@@ -53,6 +53,25 @@ run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB-lz4" --concurrency 4
 export GO_FAILPOINTS=""
 size_lz4=$(du -d 0 $TEST_DIR/$DB-lz4 | awk '{print $1}')
 
+echo "backup with tikv error start..." 
+export GO_FAILPOINTS="github.com/pingcap/br/pkg/backup/tikv-rw-error=return(\"tikv read or write error\")"
+error_msg=$(run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB-tikverr" --concurrency 4 |  awk '{match($0,/.+(\[BR:KV:ErrKVUnknown\]unknown error occur on tikv, please ensure tikv is permitted to read from & write to the storage.)/,a);if (a[1] != "") print a[1]}')
+if [ "$error_msg" = ""]
```
YuJuncen, line 59: Seems this would cause `unary operator expected`.
```suggestion
if [ "$error_msg" = "" ]
```

Issue #1063, opened 2021-04-26
Delete not applied correctly when restoring cdclog


dveeden:
1. What did you do?

```
tidb 5.7.25-TiDB-v5.0.1 > create table cdctest (id int primary key, c1 varchar(255));
Query OK, 0 rows affected (0.64 sec)

tidb 5.7.25-TiDB-v5.0.1 > insert into cdctest values (1, 'test 1');
Query OK, 1 row affected (0.12 sec)

tidb 5.7.25-TiDB-v5.0.1 > insert into cdctest values (2, 'test 2');
Query OK, 1 row affected (0.09 sec)

tidb 5.7.25-TiDB-v5.0.1 > insert into cdctest values (3, 'test 3');
Query OK, 1 row affected (0.12 sec)

tidb 5.7.25-TiDB-v5.0.1 > update cdctest set c1='update 3' where id=3;
Query OK, 1 row affected (0.14 sec)
Rows matched: 1  Changed: 1  Warnings: 0

tidb 5.7.25-TiDB-v5.0.1 > delete from cdctest where id=2;
Query OK, 1 row affected (0.13 sec)

tidb 5.7.25-TiDB-v5.0.1 > begin; select @@tidb_current_ts; rollback;
Query OK, 0 rows affected (0.00 sec)

+--------------------+
| @@tidb_current_ts  |
+--------------------+
| 424528095772147713 |
+--------------------+
1 row in set (0.03 sec)
```

```
tiup br restore cdclog --pd=192.168.122.50:2379 --storage=/tmp/cdclog --start-ts=424527916452806657 --end-ts=424528095772147713 --log-level=debug
```

```
tidb 5.7.25-TiDB-v5.0.1 > select * from cdctest;
+----+----------+
| id | c1       |
+----+----------+
|  1 | test 1   |
|  2 | NULL     |
|  3 | update 3 |
+----+----------+
3 rows in set (0.00 sec)
```

2. What did you expect to see?

```
tidb 5.7.25-TiDB-v5.0.1 > select * from cdctest;
+----+----------+
| id | c1       |
+----+----------+
|  1 | test 1   |
|  3 | update 3 |
+----+----------+
2 rows in set (0.00 sec)
```

3. What did you see instead?

```
tidb 5.7.25-TiDB-v5.0.1 > select * from cdctest;
+----+----------+
| id | c1       |
+----+----------+
|  1 | test 1   |
|  2 | NULL     |
|  3 | update 3 |
+----+----------+
[...]

PR #1069, opened 2021-04-28
cdclog: fix lost delete

3pointer:
solve #1063 

give the right op for each key.

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release Note

 - Fix the issue that log restore lost delete key during restore.

### DIFF
--- a/pkg/kv/kv.go
+++ b/pkg/kv/kv.go
@@ -19,6 +19,7 @@ import (
 	"sort"
 
 	"github.com/pingcap/errors"
+	sst "github.com/pingcap/kvproto/pkg/import_sstpb"
 	"github.com/pingcap/log"
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/mysql"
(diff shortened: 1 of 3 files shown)


kennytm:
please gofmt.

pkg/kv/kv.go
```diff
@@ -57,6 +57,8 @@ type Iter interface {
 	Value() []byte
 	// Close close this iter.
 	Close() error
+	// KeyIsDelete
+	KeyIsDelete() bool
```
glorv, line 61: Shall we replace this with something like `OpType()` so if we need to add more types of keys e.g. `Merge`, we needn't add another method.

3pointer: @kennytm I want to know what's your thoughts.

kennytm: +1 for glorv.

3pointer: Done, PTAL~

pkg/restore/ingester.go
```diff
@@ -387,12 +387,12 @@ func (i *Ingester) writeToTiKV(
 	regionMaxSize := i.regionSplitSize * 4 / 3
 
 	for iter.Seek(regionRange.Start); iter.Valid() && bytes.Compare(iter.Key(), regionRange.End) <= 0; iter.Next() {
-		size += int64(len(iter.Key()) + len(iter.Value()))
```
YuJuncen, line 390: Hold on, why remove this 🤔

glorv: +1

glorv:
reset LGTM

glorv:
undo LGTM

Issue #1055, opened 2021-04-25
Improve the human-readable of errors and logs


IANTHEREAL:
### Describe your feature request related problem:

At present, when there is an error in the BR backup, the error message that the user sees is not enough to explain the problem and help diagnose

### Describe the feature you'd like:

Let error and log information help users diagnose and solve problems quickly

zwj-coder:
Keep improving errors and logs in [#1076 ](https://github.com/pingcap/br/pull/1076)

PR #1076, opened 2021-04-29
err_msg: clarify some ambiguous error message

zwj-coder:
Keeping improve BR errors and logs [#1055 ](https://github.com/pingcap/br/issues/1055)

Some error message in BR is ambiguous and misleading. 
i.e. 
- If tikv has no permission to read from & write to backup storage, error message cannot clarify whether error occur on br side or tikv side. Users may think why I got "permission denied"? I'm running BR as root!
- If user want to restore tables or databases that have not been backup, error should occur.

Clarify the summary info including:
- `data-size => total kv size`
- `size => backup data size(after compressed)`

1. Adjust the backup and restore summary:
- the `data-size` field has been changed to `total-kv-size`
- the `size` field in backup summary has been changed to `backup data size(after compressed)` 
- the `size` field in restore summary has been changed to `restore data size(after decompressed)`

2. Some error messages are changed, and will clarify what is going on when error occur.

- Original error message is as following
```
[2021/05/18 16:37:28.517 +08:00] [ERROR] [push.go:147] ["backup occur unknown error"] [error="Io(Os { code: 13, kind: PermissionDenied...})"] [stack="github.com/pingcap/br/pkg/backup.(*pushDown).pushBackup\n\t/home/zwj/log_improve/br/pkg/backup/push.go:147\ngithub.com/pingcap/br/pkg/backup.(*Client).BackupRange\n\t/home/zwj/log_improve/br/pkg/backup/client.go:540\ngithub.com/pingcap/br/pkg/backup.(*Client).BackupRanges.func2.1\n\t/home/zwj/log_improve/br/pkg/backup/client.go:476\ngithub.com/pingcap/br/pkg/utils.(*WorkerPool).ApplyOnErrorGroup.func1\n\t/home/zwj/log_improve/br/pkg/utils/worker.go:63\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/root/go/pkg/mod/golang.org/x/sync@v0.0.0-20201020160332-67f06af15bc9/errgroup/errgroup.go:57"]
```

- For now, the error message output is as following: 
```
[2021/05/19 15:43:43.305 +08:00] [ERROR] [push.go:175] [error="[BR:KV:ErrKVStorage]tikv storage occur I/O error: I/O permission denied error occurs on TiKV Node(store id: 1; Address: 127.0.0.1:20162)"] ["work around"="please ensure tikv has permission to read from & write to the storage."] [stack="github.com/pingcap/br/pkg/backup.(*pushDown).pushBackup\n\t/home/zwj/log_improve/br/pkg/backup/push.go:175\ngithub.com/pingcap/br/pkg/backup.(*Client).BackupRange\n\t/home/zwj/log_improve/br/pkg/backup/client.go:540\ngithub.com/pingcap/br/pkg/backup.(*Client).BackupRanges.func2.1\n\t/home/zwj/log_improve/br/pkg/backup/client.go:476\ngithub.com/pingcap/br/pkg/utils.(*WorkerPool).ApplyOnErrorGroup.func1\n\t/home/zwj/log_improve/br/pkg/utils/worker.go:63\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/root/go/pkg/mod/golang.org/x/sync@v0.0.0-20201020160332-67f06af15bc9/errgroup/errgroup.go:57"]
```

3. When user want to restore tables or databases that have not been backup, error should occur.

### Check List 

Tests 

1. If user want to restore a table or database that has not been backup, an error should occur to notice the user
image
image
2. If user restore a empty table, the summary should notice the user there is nothing to restore.
image
3. Summary message have improve, data-size => total-kv-size and size => backup data size(after compressed)
image

 - Integration test

Code changes

### Release Note

 - No realease note


YuJuncen:
Eh, by the way, I have an idea about how to handle this, which might make this PR better.

At here:

https://github.com/pingcap/br/blob/b1ed365120742e378c1076c6cba94faad85bdd3b/pkg/backup/push.go#L140-L141

The 'unknown error' omits many context, which make it confusing -- users may think this error occurs in the internal of PR. How about add the store into the context? like:

```go
 return res, errors.Annotatef(berrors.ErrKVUnknown, "error occurs in the TiKV Node %s(@%s): %v", 
    store.GetID(), 
    store.GetAddress(), 
    errPb)
```

NOTEs:

Unfortunately, currently, our `pushDown` make multi workers and receive response from them by a channel(`respCh`), which omits the store info:

https://github.com/pingcap/br/blob/b1ed365120742e378c1076c6cba94faad85bdd3b/pkg/backup/push.go#L25-L29

You may would like to create a type that contains both response and store info, and replace this type to `respCh`,  then send the store info like this:
```go
push.respCh <- RespAndStore{
						Resp:  resp,
						Store: store,
					}
```
... at here:
https://github.com/pingcap/br/blob/b1ed365120742e378c1076c6cba94faad85bdd3b/pkg/backup/push.go#L79-L83

zwj-coder:
> Eh, by the way, I have an idea about how to handle this, which might make this PR better. (also cc @3pointer, PTAL)
> 
> At here:
> 
> https://github.com/pingcap/br/blob/b1ed365120742e378c1076c6cba94faad85bdd3b/pkg/backup/push.go#L140-L141
> 
> The 'unknown error' omits many context, which make it confusing -- users may think this error occurs in the internal of PR. How about add the store into the context? like:
> 
> ```go
>  return res, errors.Annotatef(berrors.ErrKVUnknown, "error occurs in the TiKV Node %s(@%s): %v", 
>     store.GetID(), 
>     redact.String(store.GetAddress()), // The net address might be sensitive, so we might need to redact them.
>     errPb)
> ```
> 
> I think after this, users can notice that this error occurs in TiKV, and we can keep the description of `ErrKVUnknown` untouched, in case of misleading in non-permission relative cases.
> 
> NOTEs:
> 
> Unfortunately, currently, our `pushDown` make multi workers and receive response from them by a channel(`respCh`), which omits the store info:
> 
> https://github.com/pingcap/br/blob/b1ed365120742e378c1076c6cba94faad85bdd3b/pkg/backup/push.go#L25-L29
> 
> You may would like to create a type that contains both response and store info, and replace this type to `respCh`, then send the store info like this:
> 
> ```go
> push.respCh <- RespAndStore{
> 	Resp:  resp,
> 	Store: store,
> }
> ```
> 
> ... at here:
> https://github.com/pingcap/br/blob/b1ed365120742e378c1076c6cba94faad85bdd3b/pkg/backup/push.go#L79-L83

I think since the error stack contains both error message from tikv and br, if the error occur at the tikv side, it's tikv's responsibility to report detailed error message.

YuJuncen:
@zwj-coder Yep, but I think BR can figure out on WHICH TiKV error happens , IMO this could be helpful for debugging. 🤔

zwj-coder:
@YuJuncen I agree, but if the tikv cluster is configured in the same way, one of them have permission problem probably means rest of them also have same problem, so we could add this store help information for debugging usage.

YuJuncen:
@zwj-coder Well, that is true.

3pointer:
/lgtm

BTW Is `internal error` better than `unknown error` here?🤔️

zwj-coder:
@3pointer TiKV has define several error types but rather coarse, internal error only indicate the error occur on TiKV, unknown error means the error TiKV has not defined occurs, maybe we should refine error type in TiKV more fine-grained.

IANTHEREAL:
Good Job! There are still some details that need to be improved

For
```
 ["unknown error occurs on TiKV Node(store id: 4; Address: 127.0.0.1:20161)"] 
 [error="[BR:KV:ErrKVUnknown]unknown error occur on tikv"] 
 ["error message from TiKV"="tikv read or write error"] 
 ["work around"="please ensure tikv is permitted to read from & write to the storage."] 
```

The error information is redundant, causing trouble to reading. 
```
 ["unknown error occurs on TiKV Node(store id: 4; Address: 127.0.0.1:20161)"] 
 [error="[BR:KV:ErrKVUnknown]unknown error occur on tikv"] 
```

`tikv read or write error` is not `unknown error`, it woule make users confused, the information is confusing and redundant, and does not meet the error code definition specification

I suggest to optimize the error message to

```
 [error="[BR:KV:StorageError]error occur on TiKV Node(store id: 4; Address: 127.0.0.1:20161)""] 
 ["error message from TiKV"="tikv read or write error"] 
 ["work around"="please ensure tikv is permitted to read from & write to the storage."] 
```
You can make adjustments according to the actual situation of the code

zwj-coder:
@IANTHEREAL  PTAL

zwj-coder:
@IANTHEREAL PTAL

pkg/task/backup.go
```diff
@@ -434,7 +434,7 @@ func RunBackup(c context.Context, g glue.Glue, cmdName string, cfg *BackupConfig
 		return errors.Trace(err)
 	}
 
-	g.Record("Size", utils.ArchiveSize(&backupMeta))
```
3pointer, line 437: if delete this line. The SQL backup won't be able to record the size.

pkg/task/restore.go
```diff
@@ -210,7 +239,7 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 	if err != nil {
 		return errors.Trace(err)
 	}
-	g.Record("Size", utils.ArchiveSize(backupMeta))
```
3pointer, line 213: ditto

pkg/utils/permission.go
```diff
@@ -0,0 +1,20 @@
+package utils
+
+import "strings"
+
+var (
+	notFoundMsg         = "notfound"
```
3pointer, line 6: notfound is too general, I'm afraid some other cloud errors will reach here too. it's better to add another key word. such as `io`

pkg/task/common.go
```diff
@@ -297,11 +303,13 @@ func (cfg *Config) ParseFromFlags(flags *pflag.FlagSet) error {
 		if len(db) == 0 {
 			return errors.Annotate(berrors.ErrInvalidArgument, "empty database name is not allowed")
 		}
+		cfg.Schemas[db] = struct{}{}
 		if tblFlag := flags.Lookup(flagTable); tblFlag != nil {
 			tbl := tblFlag.Value.String()
 			if len(tbl) == 0 {
 				return errors.Annotate(berrors.ErrInvalidArgument, "empty table name is not allowed")
 			}
+			cfg.Tables[tbl] = struct{}{}
```
3pointer, line 312: we could have the same tables in different dbs.
It's better to change the key from `tbl` to `db.tbl`.

pkg/task/restore.go
```diff
@@ -150,6 +150,35 @@ func (cfg *RestoreConfig) adjustRestoreConfig() {
 	}
 }
 
+// CheckRestoreDBAndTable is used to check whether the restore dbs or tables have been backup
+func CheckRestoreDBAndTable(client *restore.Client, cfg *RestoreConfig) error {
+	// if the databases or tables restore should be in backup data
```
3pointer, line 155:
```
if len(cfg.Schemas) == 0 && len(cfg.Tables) == 0 {   
       // we can avoid building the tablesMap and schemasMap in some cases. 
       return nil
}
```

PR #1080, opened 2021-04-30
lightning: save chunk checkpoint timely

glorv:
Try to save chunk checkpoint timely.

- Return a chunk status object when close chunk but it's not flushed. Then before each chunk restorer started, check and save all the pending chunk checkpoint if available.
- Try to save the chunk checkpoint before and after each deliver loop.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release Note

 - check and save chunk checkpoint timely.


pkg/lightning/backend/local/local.go
```diff
@@ -165,6 +165,13 @@ type File struct {
 	sstIngester    sstIngester
 	finishedRanges syncedRanges
 
+	// sst seq lock
+	seqLock sync.Mutex
+	// seq number for incoming sst meta
+	nextSeq atomic.Int32
```
kennytm, line 171: seems unnecessary to make `nextSeq` atomic as it is only used inside `e.seqLock`

glorv: yes, fixed.

pkg/lightning/backend/importer/importer_test.go
```diff
@@ -115,8 +115,9 @@ func (s *importerSuite) TestWriteRows(c *C) {
 	c.Assert(err, IsNil)
 	err = writer.WriteRows(s.ctx, nil, s.kvPairs)
 	c.Assert(err, IsNil)
-	err = writer.Close(s.ctx)
+	st, err := writer.Close(s.ctx)
```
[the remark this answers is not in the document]
glorv: checked in L120, though it should be nil for importer backend

pkg/lightning/backend/local/local.go
```diff
@@ -2519,23 +2550,37 @@ func (w *Writer) flush(ctx context.Context) error {
...
 	// FIXME: in theory this line is useless, but In our benchmark with go1.15
 	// this can resolve the memory consistently increasing issue.
 	// maybe this is a bug related to go GC mechanism.
 	w.writeBatch = nil
-	return err
+	return flushStatus{local: w.local, seq: w.lastMetaSeq}, err
+}
+
+func (w *Writer) IsSynced() bool {
+	return w.batchCount == 0 && w.lastMetaSeq <= w.local.finishedMetaSeq.Load()
```
Little-Wallace, line 2583: Does it need to be called in a mutex?

Little-Wallace (commented):
rest LGTM

Little-Wallace (approved):
LGTM

gozssky:
LGTM

gozssky (approved):

kennytm (approved):

PR #1089, opened 2021-05-07
lightning/restore: support ingset multi ssts for same range

glorv:
Ingest multi SST files (different CF) together to gain better consistency. This can resolve the problem that lighting import or BR restore may cause TiFlash panic.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release Note

 - No Release Note


pkg/restore/import.go
```diff
@@ -155,16 +177,32 @@ func (ic *importClient) GetImportClient(
 	return client, errors.Trace(err)
 }
 
+func (ic *importClient) SupportMultiIngest(ctx context.Context, stores []uint64) (bool, error) {
```
Little-Wallace, line 180: Why storeId is an array?

Little-Wallace (approved):
LGTM

pkg/lightning/backend/local/local.go
```diff
@@ -868,9 +871,39 @@ func NewLocalBackend(
...
+	for _, s := range stores {
+		client, err := local.getImportClient(ctx, s.Id)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		_, err = client.MultiIngest(ctx, &sst.MultiIngestRequest{})
+		if err != nil {
+			if s, ok := status.FromError(err); ok {
+				if s.Code() == codes.Unimplemented {
+					local.supportMultiIngest = false
```
solotzg, line 895: Does tiflash need to implement such interfaces? @Little-Wallace

Little-Wallace: No

solotzg: It seems that `supportMultiIngest` is true only when all stores have implemented related interfaces.

glorv: MultiIngest is only called by region leader, so tiflash needn't implement this interface. We should skip check tiflash here

pkg/restore/client.go
```diff
@@ -541,6 +541,38 @@ func (rc *Client) setSpeedLimit(ctx context.Context) error {
...
+	}
+	idx := 1
+	for idx < len(files) {
+		if !isFilesBelongToSameRange(files[idx-1].Name, files[idx].Name) {
+			break
+		}
+		idx++
+	}
+
+	return files[idx:], files[idx:]
```
3pointer, line 573: files[:idx], files[idx:] 🤔️？

glorv: 😂

Issue #870, opened 2021-03-15
When br is restoring cdclog, if there is a ddl operation in the TS interval, the ddl operation will not be restored


tongtongyin:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
When cdc outputs the ddl operation, name the suffix of the ddl file name as uint64max-TS of the first ddl operation (it can be directly named as the TS of the last ddl), and then when collecting ddl files during br recovery, add this ts is regarded as the TS of the last ddl, and all ddl smaller than startTs are filtered out, so some ddl files that should not be filtered are filtered out, so some ddl operations are not restored when they are restored.

2. What did you expect to see?

3. What did you see instead?

4. What version of BR and TiDB/TiKV/PD are you using?

5. Operation logs
   - Please upload `br.log` for BR if possible
   - Please upload `tidb-lightning.log` for TiDB-Lightning if possible
   - Please upload `tikv-importer.log` from TiKV-Importer if possible
   - Other interesting logs

6. Configuration of the cluster and the task
   - `tidb-lightning.toml` for TiDB-Lightning if possible
   - `tikv-importer.toml` for TiKV-Importer if possible
   - `topology.yml` if deployed by TiUP

7. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus if possible

3pointer:
The reason we choose `uint64max-TS` is [here](https://github.com/pingcap/br/blob/master/pkg/restore/log_client.go#L190-L194)， And I didn't get it why `all ddls smaller than startTs are filtered out is wrong`? could you provide an example?

dveeden:
```
tidb 5.7.25-TiDB-v5.0.1 > begin; select @@tidb_current_ts; rollback;
Query OK, 0 rows affected (0.01 sec)

+--------------------+
| @@tidb_current_ts  |
+--------------------+
| 424528599009984514 |
+--------------------+
1 row in set (0.00 sec)

Query OK, 0 rows affected (0.00 sec)

tidb 5.7.25-TiDB-v5.0.1 > create table cdcinsert (id int primary key);
Query OK, 0 rows affected (0.60 sec)

tidb 5.7.25-TiDB-v5.0.1 > insert into cdcinsert values (1),(2),(3);
Query OK, 3 rows affected (0.10 sec)
Records: 3  Duplicates: 0  Warnings: 0

[...]

PR #1094, opened 2021-05-10
cdclog: fix the lost ddl files

3pointer:
Solve #870 

ddl file use the first event's ts to build filename.
so the filter should be changed to `the first event's ts < endTS`

### Check List 

Tests 

 - Unit test

Related changes

### Release Note

 - Fix the issue that cdclog lost some DDLs during restore.


overvenus (approved):

kennytm:
/run-integration-test 

```
[2021-06-16T08:15:58.454Z] TEST: [br_full_ddl] fail due to stats are not equal
```

glorv (approved):

kennytm:
/run-integration-test 

```
[2021-07-05T03:14:55.223Z] TEST: [br_full_ddl] fail due to stats are not equal
```

kennytm:
Still failing with the same error. I think the fix is wrong.

PR #1095, opened 2021-05-10
restore: fix restore system table bug && add some unrecoverable tables

3pointer:

fix restore system table bug && add some unrecoverable tables
### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release Note

 - Fix the issue that restore from `mysql` schema would failed.

### DIFF
--- a/pkg/restore/systable_restore.go
+++ b/pkg/restore/systable_restore.go
@@ -29,6 +29,17 @@ var statsTables = map[string]struct{}{
 	"stats_top_n":      {},
 }
 
+var unRecoverableTable = map[string]struct{}{
+	// some variables in tidb (e.g. gc_safe_point) cannot be recovered.
+	"tidb":             {},
+	"global_variables": {},
+}
+
+func isUnrecoverableTable(tableName string) bool {
+	_, ok := unRecoverableTable[tableName]
+	return ok
+}
+
 func isStatsTable(tableName string) bool {
 	_, ok := statsTables[tableName]
 	return ok
@@ -61,15 +72,15 @@ func (rc *Client) RestoreSystemSchemas(ctx context.Context, f filter.Filter) {
 	tablesRestored := make([]string, 0, len(originDatabase.Tables))
 	for _, table := range originDatabase.Tables {
 		tableName := table.Info.Name
-		if f.MatchTable(sysDB, tableName.O) {
+		if f.MatchTable(temporaryDB.O, tableName.O) {
 			if err := rc.replaceTemporaryTableToSystable(ctx, tableName.L, db); err != nil {
 				logutil.WarnTerm("error during merging temporary tables into system tables",
 					logutil.ShortError(err),
 					zap.Stringer("table", tableName),
 				)
 			}
+			tablesRestored = append(tablesRestored, tableName.L)
 		}
-		tablesRestored = append(tablesRestored, tableName.L)
 	}
 	if err := rc.afterSystemTablesReplaced(ctx, tablesRestored); err != nil {
 		for _, e := range multierr.Errors(err) {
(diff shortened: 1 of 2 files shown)


pkg/restore/systable_restore.go
```diff
@@ -61,15 +72,15 @@ func (rc *Client) RestoreSystemSchemas(ctx context.Context, f filter.Filter) {
 	tablesRestored := make([]string, 0, len(originDatabase.Tables))
 	for _, table := range originDatabase.Tables {
 		tableName := table.Info.Name
-		if f.MatchTable(sysDB, tableName.O) {
+		if f.MatchTable(temporaryDB.O, tableName.O) {
```
YuJuncen, line 75: Seems after this, users must use `-f __TiDB_BR_Temporary_mysql.xxx` to specify which tables to be restored. 🤔

tests/br_systables/run.sh
```diff
@@ -37,7 +37,8 @@ check() {
     run_sql "SELECT count(*) from mysql.foo;" | grep 11
     run_sql "SELECT count(*) from mysql.usertable;" | grep 1000
     run_sql "SHOW TABLES IN mysql;" | grep -v bar
```
YuJuncen, line 39:
```suggestion
    run_sql "SHOW TABLES IN mysql;" | awk '/bar/{exit 1}'
```

According to the log, the table `mysql.bar` has been restored, but there are other tables too, so `grep` generated a false negative result. 🤔

(the log: `[2021-05-13T06:53:25.176Z] [2021/05/13 14:53:24.980 +08:00] [INFO] [systable_restore.go:148] ["successfully restore system database"] [table=bar] [database=mysql] [sql="RENAME TABLE `__tidb_br_temporary_mysql`.`bar` TO `mysql`.`bar`;"]`, from [here](https://ci.pingcap.net/blue/rest/organizations/jenkins/pipelines/br_ghpr_unit_and_integration_test/runs/1797/nodes/74/steps/1073/log/?start=0))

pkg/restore/systable_restore.go
3pointer: yes... but for now we will filter `!__TiDB_BR_Temporary_mysql.xxx` in[ L45](__TiDB_BR_Temporary_mysql.xxx)...

YuJuncen: Once the user specifies any table filter, seems that rule would be overridden. So it would be OK? 🤔

3pointer: I didn't get it. 

In L45, we filtered out other DBs only keep the `temporaryDB` which means we can not meet `sysDB` here.

3pointer: which means when we do 
`br backup db --db mysql xxx`
and then `br restore db --db mysql` we will get nothing to restore...

3pointer:
ignore compatibility-test, since the docker images didn't update to latest master.

3pointer:
@YuJuncen PTAL

pkg/restore/systable_restore.go
YuJuncen: Well, I think we'd use original database name everywhere. L45 is buggy😢, the current version is fine.

tests/br_systables/run.sh
```diff
@@ -36,8 +36,9 @@ rollback_modify() {
 check() {
     run_sql "SELECT count(*) from mysql.foo;" | grep 11
     run_sql "SELECT count(*) from mysql.usertable;" | grep 1000
-    run_sql "SHOW TABLES IN mysql;" | grep -v bar
-    run_sql "SELECT VARIABLE_VALUE FROM mysql.tidb WHERE VARIABLE_NAME = 'tikv_gc_life_time'" | grep "1h"
+    run_sql "SHOW TABLES IN mysql;" | awk '/bar/{exit 1}'
+    # we cannot let user overwrite `mysql.tidb` through br in any time.
+    run_sql "SELECT VARIABLE_VALUE FROM mysql.tidb WHERE VARIABLE_NAME = 'tikv_gc_life_time'" | grep -v "1h"
```
YuJuncen, line 41:
```suggestion
    run_sql "SELECT VARIABLE_VALUE FROM mysql.tidb WHERE VARIABLE_NAME = 'tikv_gc_life_time'" | awk '/1h/{exit 1}'
```

Since `run_sql` could make extra lines like `***** row *****` each time, I think it would be better to replace all `grep -v <pattern>` with `awk '/<pattern>/{exit 1}'` here for preventing false negative?

YuJuncen (commented):
Rest LGTM

PR #1101, opened 2021-05-11
lightning: make every engine use a unique commitTS

### LINKED ISSUE #1108: lightning: let each engine use a unique commit ts when ingest into tikv

Currently, the whole lightning task uses the same commit ts when ingest key-values to TiKV. So if kv from different conflict with each other, we don't whether this is because there are duplicated data or just because we ingest the same multi-times due to retry.

gozssky:

close #1108 

To support duplicate detection for local backend, we need to ensure a same commitTS is used for the same engine to avoid ingesting kv pairs more than once with different commitTS. And we should make every engine use a unique commitTS so that we can detect cross-engine duplicates

Do not merge this pr before #1080 . 

- Make every engine use a unique commitTS which is obtained from PD
- Attach a commitTS to every engine and save it in engineMeta.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release Note

- lightning: make every engine use a unique commitTS


pkg/lightning/checkpoints/checkpoints.go
```diff
@@ -106,6 +106,7 @@ const (
 			table_name varchar(261) NOT NULL,
 			engine_id int NOT NULL,
 			status tinyint unsigned DEFAULT 30,
+			commit_ts bigint NOT NULL DEFAULT 0,
```
overvenus, line 109: Bigint ranges from -9223372036854775808 to 9223372036854775807, it may cause TSO(uint64) overflow.

pkg/lightning/checkpoints/checkpoints.go
```diff
@@ -106,6 +106,7 @@ const (
 			table_name varchar(261) NOT NULL,
 			engine_id int NOT NULL,
 			status tinyint unsigned DEFAULT 30,
+			commit_ts bigint unsigned NOT NULL DEFAULT 0,
```
overvenus, line 109: Could you add a test where commit_ts = uint64::MAX?

gozssky: I found it is not easy to test with uint64::MAX. It seems the [go-sqlmock](https://github.com/DATA-DOG/go-sqlmock) we used in test doesn't allow large uint64 value. It may be caused by golang/go#9373. I test that converting uint64 in ExecContext  to a string is ok. But I don't known if there is a potential risk. Do you have any suggestions?

pkg/lightning/backend/backend.go
```diff
@@ -143,6 +143,10 @@ type AbstractBackend interface {
 
 	OpenEngine(ctx context.Context, config *EngineConfig, engineUUID uuid.UUID) error
 
+	SetEngineTSIfNotExists(ctx context.Context, engineUUID uuid.UUID, ts uint64) error
```
kennytm, line 146: suggestion: 

1. turn this function into `AllocateTSIfNotExists(ctx context.Context, engineUUID uuid.UUID) error`, which is the same as `allocateCommitTS` followed by `SetEngineTSIfNotExists`
2. local and importer backends can share the same `allocateCommitTS` implementation, and both contain the `pdAddr` field
3. tidb and noop backends just return nil.
4. get rid of `rc.isImporterBackend()`.

kennytm (commented):
rest LGTM

pkg/lightning/backend/backend.go
gozssky: done in https://github.com/pingcap/br/pull/1101/commits/7716e864c6abfecf7b266d0f655da97b220c461a

kennytm (approved):

pkg/lightning/restore/restore.go
```diff
@@ -2823,12 +2859,26 @@ func (cr *chunkRestore) deliverLoop(
...
+				time.Sleep(time.Millisecond)
+			}
 			defer rc.diskQuotaLock.RUnlock()
 
-			// try to update chunk checkpoint, this can help save checkpoint after importing when disk-quota is triggered
-			if !dataSynced {
-				dataSynced = cr.maybeSaveCheckpoint(rc, t, engineID, cr.chunk, dataEngine, indexEngine)
+			// When enabled DiskQuota, `enforceDiskQuota` may force the backend to import the content of a large engine
+			// into the target and then reset the engine to empty. When this happen, we need to use a new commitTS.
+			if err = rc.backend.AllocateTSIfNotExists(ctx, t.tableName, engineID); err != nil {
```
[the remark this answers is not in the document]
gozssky: Sounds very reasonable. Maybe we can do it inside each backend's implementation and remove `AllocateTSIfNotExists`.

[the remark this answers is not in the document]
gozssky: done in https://github.com/pingcap/br/pull/1101/commits/0bb61ab43c1e93934fc0d1b9769c19ef66233b23

glorv (approved):

glorv:
@kennytm PTAL again since some interface has changed

Little-Wallace (approved):
LGTM

PR #1103, opened 2021-05-13
*: update tidb

disksing:
Signed-off-by: disksing <i@disksing.com>

update tidb dependency

- go get -u tidb
- update mockChecksumKVClient to meet new interface
- use kv.PriorityLow instead of tikvstore.PriorityLow

### Check List 

Tests 
 - Unit test

### Release note
- No release note


tisonkun:
@disksing iiuc `Release Note` should be `Release note` because it checks exactly matched.

disksing:
> @disksing iiuc `Release Note` should be `Release note` because it checks exactly matched.

Interesting. `Release Note` It's from the PR template.

PR #1105, opened 2021-05-13
lightning: fix encode kvs size greater than 4.0g caused pebble panic

3pointer:
`panic: pebble: batch too large: >= 4.0 G`

sends kv batch every 96k.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release Note

 - Fix the issue that lightning panic due to batch kv greater than 4.0g.

### DIFF
--- a/pkg/lightning/restore/restore.go
+++ b/pkg/lightning/restore/restore.go
@@ -2638,6 +2638,9 @@ func (cr *chunkRestore) encodeLoop(
 			}
 			kvPacket = append(kvPacket, deliveredKVs{kvs: kvs, columns: columnNames, offset: newOffset, rowID: rowID})
 			kvSize += kvs.Size()
+			failpoint.Inject("mock-kv-size", func(val failpoint.Value) {
+				kvSize += uint64(val.(int))
+			})
 			// pebble cannot allow > 4.0G kv in one batch.
 			// we will meet pebble panic when import sql file and each kv has the size larger than 4G / maxKvPairsCnt.
 			// so add this check.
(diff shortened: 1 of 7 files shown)


pkg/lightning/restore/restore.go
```diff
@@ -2636,6 +2637,17 @@ func (cr *chunkRestore) encodeLoop(
...
 			}
 			kvPacket = append(kvPacket, deliveredKVs{kvs: kvs, columns: columnNames, offset: newOffset, rowID: rowID})
+			kvSize += kvs.Size()
+			failpoint.Inject("mock-kv-size", func(val failpoint.Value) {
+				kvSize += uint64(val.(int))
+			})
+			// pebble cannot allow > 4.0G kv in one batch.
+			// we will meet pebble panic when import sql file and each kv has the size larger than 4G / maxKvPairsCnt.
+			// so add this check.
+			if kvSize > minDeliverBytes {
```
glorv, line 2647:
```suggestion
			if kvSize >= minDeliverBytes || len(kvPacket) >= maxKvPairsCnt || newOffset == cr.chunk.Chunk.EndOffset {
				canDeliver = true
				kvSize = 0
			}
```

glorv: Please merge the two if together

gozssky: kvPacket contains both `dataKVs` and `indexKVs` here. Since `dataKVs` and `indexKVs` are wrote to different engines, how about moving this check to `deliverLoop`?

Little-Wallace (approved):
LGTM

PR #1115, opened 2021-05-17
lightning: avoid ignore error when import data fails

glorv:
Let lightning return the actual error instead of return `context.Canceled`

- Retry split regions instead of direct return an error in `BatchSplitRegions`
- check parent's context cancel event to avoid direct returning when chunk routine fails.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release note

 - No release note


Little-Wallace (approved):
LGTM

3pointer (commented):
/lgtm

PR #1120, opened 2021-05-19
backup/checksum: add cluster index support for ranges

Leavrth:

ranges only support single integer primary key.

Let ranges be NotNullRange if table use the common handle

### Check List 

Tests 

 - [x] Unit test
 - [ ] Integration test
 - [x] Manual test

Code changes

Related changes

### Release note

 - No release note


pkg/backup/client.go
```diff
@@ -266,8 +266,18 @@ func BuildTableRanges(tbl *model.TableInfo) ([]kv.KeyRange, error) {
 }
 
 func appendRanges(tbl *model.TableInfo, tblID int64) ([]kv.KeyRange, error) {
-	ranges := ranger.FullIntRange(false)
-	kvRanges := distsql.TableRangesToKVRanges(tblID, ranges, nil)
+	var ranges []*ranger.Range
+	if tbl.IsCommonHandle {
```
3pointer, line 270: can we re-use the [TableHandleRangesToKVRanges](https://github.com/pingcap/tidb/blob/8fb29eb17e88cabd3968d6034d1e976823f2f79e/distsql/request_builder.go#L321) in TiDB

Little-Wallace: Do we need to judge wether the type of pk is integer?

Little-Wallace (approved):
LGTM

PR #1139, opened 2021-05-24
pkg: use the GoTimeToTS() from oracle pkg

JmPotato:

`GoTimeToTS()` from [varsutil.go#L469](https://github.com/pingcap/tidb/blob/c8bc701170165e82cc7893ab68523f9a2059a7e6/sessionctx/variable/varsutil.go#L469) is exactly the same as `GoTimeToTS()` in `oracle` pkg which is the original version and the right place it should belong to.

Use `oracle.GoTimeToTS()` instead of `variable.GoTimeToTS()`, and the latter will be removed in pingcap/tidb#24813.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

 - pingcap/tidb#24766

### Release note

 - No release note.

### DIFF
--- a/pkg/task/backup.go
+++ b/pkg/task/backup.go
@@ -19,8 +19,8 @@ import (
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/mysql"
 	"github.com/pingcap/tidb/sessionctx/stmtctx"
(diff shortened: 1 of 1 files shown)


Yisaer (approved):
/lgtm

Little-Wallace (approved):
LGTM

PR #1146, opened 2021-05-26
tests: delete instability test case temporary

Leavrth:

outage-at-finegrained in `br_tikv_outage` is instability which somtimes cause CI failed.

delete it

### Release note

 -No release note

### DIFF
--- a/tests/br_tikv_outage/run.sh
+++ b/tests/br_tikv_outage/run.sh
@@ -54,7 +54,8 @@ hint_backup_start=$TEST_DIR/hint_backup_start
 hint_get_backup_client=$TEST_DIR/hint_get_backup_client
 
 
-cases=${cases:-'outage outage-after-request outage-at-finegrained shutdown scale-out'}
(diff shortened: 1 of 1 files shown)


tests/br_tikv_outage/run.sh
```diff
@@ -54,7 +54,8 @@ hint_backup_start=$TEST_DIR/hint_backup_start
 hint_get_backup_client=$TEST_DIR/hint_get_backup_client
 
 
-cases=${cases:-'outage outage-after-request outage-at-finegrained shutdown scale-out'}
+# cases=${cases:-'outage outage-after-request outage-at-finegrained shutdown scale-out'}
+cases=${cases:-'outage outage-after-request shutdown scale-out'}
```
lonng, line 58: It's a bad idea to delete some case, we should file an issue to record that if we want to disable it temporarily. And we should add the issue link at above of the L57 to tracking it.

Leavrth: Got it. I will close the pr.

PR #1151, opened 2021-05-27
tests: add row count check for br clustered index tests

Leavrth:

Key range in backup and checksum might be wrong at the same time(They are still the same) when the table use clustered index, which means checksum cannot find this error.

add row_count check for the test.

### Check List 

### Release note

 -No release note


3pointer (commented):
/lgtm

lonng (commented):
/lgtm

Issue #1158, opened 2021-05-31
tests: br_tikv_outage sometimes fail


Leavrth:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?

In CI test https://ci.pingcap.net/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/3238/pipeline

2. What did you expect to see?

br_tikv_outage expect to kill a tikv when backup is running.

3. What did you see instead?

In this case, backup is finished before a tikv is killed.

4. What version of BR and TiDB/TiKV/PD are you using?

master

5. Operation logs

backup is finished:
`[2021-05-28T03:47:21.135Z] [2021/05/28 11:47:20.931 +08:00] [INFO] [collector.go:232] ["Full backup success summary"] [total-ranges=52] [ranges-succeed=52] [ranges-failed=0] [backup-checksum=20.198435441s] [backup-fast-checksum=8.184174ms] [backup-total-ranges=52] [backup-total-regions=52] [total-take=26.347638342s] [total-kv=41259] [total-kv-size=22.79MB] [average-speed=3.779MB/s] ["backup data size(after compressed)"=15.95MB] [BackupTS=425243080030683137]`

tikv restart:
`[2021-05-28T03:47:31.105Z] + start_tikv 3`

so this line will be done immediately because backup is already finished.
https://github.com/pingcap/br/blob/master/tests/br_tikv_outage/run.sh#L73

Then the test will do restore operation.
`[2021-05-28T03:47:31.105Z] + bin/br.test -test.coverprofile=/tmp/backup_restore_test/cov.br_tikv_outage.1713.out.log DEVEL restore full -s 'local:///tmp/backup_restore_test/backup{test:br_tikv_outage|with:outage}' -L debug --ca /tmp/backup_restore_test/certs/ca.pem --cert /tmp/backup_restore_test/certs/br.pem --key /tmp/backup_restore_test/certs/br.key`

But in tikv3 log, storage has not been started yet when restore is running.
`[2021-05-28T03:47:36.032Z] [2021/05/28 11:47:30.051 +08:00] [INFO] [mod.rs:215] ["Storage started."]`

So this test is failed.

PR #1160, opened 2021-05-31
tests/br: fix bug in issue #1158

Leavrth:

fix bug in issue #1158

for #1158 add sleep 1 to ensure tikv start up completely before restore

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

 -No release note

### DIFF
--- a/tests/br_tikv_outage/run.sh
+++ b/tests/br_tikv_outage/run.sh
@@ -71,6 +71,8 @@ github.com/pingcap/br/pkg/conn/hint-get-backup-client=1*return(\"$hint_get_backu
     backup_pid=$!
     single_point_fault $failure
     wait $backup_pid
+    # let tikv start up completely if backup is finished before tikv restarts
+    sleep 1
     case $failure in
     scale-out | shutdown | outage-at-finegrained ) stop_services
         start_services ;;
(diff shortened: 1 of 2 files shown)


YuJuncen:
Maybe the utility function `ensure_tikv` would be suitable for this? (Eh, maybe needing some modify, say, ensure every TiKV nodes are up via interfaces of PD.)

https://github.com/pingcap/br/blob/701909b29ffa5b964d437e15fa9acfbf046255cb/tests/_utils/run_services#L144-L154

BTW, according to the log provided in #1158, seems TiKV 3 started after 5 seconds since restoration starting. I guess maybe 1s of sleeping isn't enough?

Leavrth:
> Maybe the utility function `ensure_tikv` would be suitable for this? (Eh, maybe needing some modify, say, ensure every TiKV nodes are up and connected via interfaces of PD.)

Got it, I will modify the code and have a test.

>  Well, seems there is clock drifting between the host and the container.

Yes, there seems 2 different clocks in bash and in BR

3pointer (approved):

tests/br_tikv_outage/run.sh
```diff
@@ -71,6 +71,8 @@ github.com/pingcap/br/pkg/conn/hint-get-backup-client=1*return(\"$hint_get_backu
     backup_pid=$!
     single_point_fault $failure
     wait $backup_pid
+    # let tikv start up completely if backup is finished before tikv restarts
+    ensure_tikv
```
glorv, line 75: shall we move this line into the case block to make it clearer which failures depend on this

Leavrth: I add the line into kv_outage function where tikv restart. It seems be better.

glorv (approved):

3pointer (approved):

Issue #1201, opened 2021-06-10
Restore system table mysql.user failed


fubinzh:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
a. Add a user to TiDB 
b. Do full backup with "br backup ful ..."
c. Delete the user from TiDB
d. Do br full restore with below command:
    ./br restore full --pd "172.16.6.183:2379" -f "mysql*.*" -f "oomtest*.*" --storage "local:///minio/fubin/tidb510_br510" --log-file system_full_backup2.restore.log

2. What did you expect to see?
System table mysql.user should be recovered correctly.

3. What did you see instead?
mysql.user table not recovered.

== error below is seen in system_full_backup2.restore.log: ==
[2021/06/10 14:37:28.639 +08:00] [INFO] [systable_restore.go:174] ["table existing, using replace into for restore"] [table=user] [schema=mysql]
[2021/06/10 14:37:28.639 +08:00] [WARN] [systable_restore.go:140] ["failed to execute SQL restore system database"] [table=user] [database=mysql] [sql="REPLACE INTO `mysql`.`user` SELECT * FROM `__tidb_br_temporary_mysql`.`user`;"] [error="[schema:1146]Table '__tidb_br_temporary_mysql.user' doesn't exist"] [errorVerbose="[schema:1146]Table '__tidb_br_temporary_mysql.user' doesn't exist

==detailed steps ==
[root@centos76_vm dev]# mysql -h172.16.6.151 -uroot -P4000 -E -e "CREATE USER 'toolsqa'@'%' IDENTIFIED BY 'changeme'; GRANT ALL PRIVILEGES ON *.* TO 'toolsqa'@'%'; FLUSH PRIVILEGES;"
[root@centos76_vm dev]# mysql -h172.16.6.151 -uroot -P4000 -E -e "select user from mysql.user;"        
*************************** 1. row ***************************
user: root
*************************** 2. row ***************************
user: toolsqa

[root@centos76_vm dev]# ./br backup full --pd "172.16.6.183:2379" --storage "local:///minio/fubin/tidb510_br510" --log-file system_full_backup2.log
...
[...]

Issue #1197, opened 2021-06-10
Can't restore system tables separately without restoring user tables


fubinzh:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

a. Run "br backup full" to create full backup, including system tables and user tables.
b. Run `br restore full --filter "mysql*.*"` to restore system tables only, (do not restore user tables)

2. What did you expect to see?
System tables mysql.* should be restored correctly.

3. What did you see instead?

br failed to restore system tables. "Nothing to restore" seen in log.

4. What version of BR and TiDB/TiKV/PD are you using?

br 5.1.0 dev build.
[root@centos76_vm dev]# ./br -V
Release Version: v5.0.0-master-dirty
Git Commit Hash: b9c4ea60eb131cf306805e6d149dda93c24ec7b1
Git Branch: backupmetav2
Go Version: go1.16.5
UTC Build Time: 2021-06-09 12:33:55
Race Enabled: false

5. Operation logs
   - Please upload `br.log` for BR if possible
   - Please upload `tidb-lightning.log` for TiDB-Lightning if possible
   - Please upload `tikv-importer.log` from TiKV-Importer if possible
   - Other interesting logs

[root@centos76_vm dev]# ./br restore full --pd "172.16.6.183:2379" --filter "mysql*.*" --storage "local:///minio/fubin/tidb510_br510" --log-file full_restore_2_system.log
[...]

PR #1207, opened 2021-06-11
restore: Fix system tables are filtered bug

zwj-coder:
Fix issue [#1197](https://github.com/pingcap/br/issues/1197) and [#1201](https://github.com/pingcap/br/issues/1201)
When user only want to restore the system table, i.e. -f "mysql*.*", because the system database was rename to a temporary name, that name will be filtered causing no system table is restored 

When filter restore databases and tables, revert the temporary name to its original name 

### Check List 

Tests 

 - Integration test
 
Code changes

Related changes

### Release note

 - No Release Note


tests/br_systables/run.sh
```diff
@@ -47,8 +65,28 @@ check() {
...
+    run_sql "SELECT user FROM mysql.user WHERE user='newuser';" | grep 'newuser'
+}
+
 modify_systables
 run_br backup full -s "local://$backup_dir"
 rollback_modify
 run_br restore full -f '*.*' -f '!mysql.bar' -s "local://$backup_dir"
 check
+
+run_br restore full -f 'mysql.bar' -s "local://$backup_dir"
```
3pointer, line 79: better to check `nothing to restore` here 🤔️?

pkg/utils/schema.go
```diff
@@ -104,3 +104,10 @@ func IsSysDB(dbLowerName string) bool {
 func TemporaryDBName(db string) model.CIStr {
 	return model.NewCIStr("__TiDB_BR_Temporary_" + db)
 }
+
+// GetSysDBName get the original name of system DB
+func GetSysDBName(tempDB model.CIStr) (string, bool) {
+	ok := strings.Contains(tempDB.O, "__TiDB_BR_Temporary_")
+	i := strings.LastIndex(tempDB.O, "_")
+	return tempDB.O[i+1:], ok
```
kennytm, line 112: it should be checking if `ok := strings.HasPrefix(tempDB.O, "__TiDB_BR_Temporary_")`, right? And the `i+1` is the length of that temporary string.

zwj-coder: fixed this method

3pointer (approved):

kennytm (approved):

PR #1209, opened 2021-06-11
lightning: fix deliver kvs too few at a time

gozssky:

In `chunkRestore.encodeLoop`, when `kvSize >= minDeliverBytes`, it will trigger a deliver. But `minDeliverBytes` (`96KB`) is too small, which probably impact performance. I think a larger limit should be more reasonable.
https://github.com/pingcap/br/blob/221bed67d5517873ce412e93150397b7ea20d1ef/pkg/lightning/restore/restore.go#L3028-L3034

Introduce a new `maxDeliverBytes` which value is `3840MB`.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### Release note

 - lightning: fix deliver kvs too few at a time


gozssky:
After test, `minDeliverBytes` is large enough  and it doesn't impact performance. So close this pr.

PR #1240, opened 2021-06-17
lightning: use region info to split region

glorv:
Optimize region split by skip split regions if there is only one split key and the region size is smaller than region-split-size

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note

 - No release note


pkg/lightning/backend/backend.go
```diff
@@ -143,7 +146,7 @@ type AbstractBackend interface {
 
 	OpenEngine(ctx context.Context, config *EngineConfig, engineUUID uuid.UUID) error
 
-	CloseEngine(ctx context.Context, engineUUID uuid.UUID) error
+	CloseEngine(ctx context.Context, config *EngineConfig, engineUUID uuid.UUID) error
```
kennytm, line 149: can the table info be stored in `OpenEngine()`, so we don't need to modify `CloseEngine()`

glorv: Not really. When lightning restart at import phase, only CloseEngine is called, in this situation, there is no tableInfo available

Little-Wallace (approved):
LGTM

kennytm (approved):

PR #1243, opened 2021-06-17
restore: fix [files lost] in log info

Leavrth:

1. files will be empty, which cause to log info print empty content.
https://github.com/pingcap/br/blob/f0c1eb5188549489440cb9ae2eb371883cc7dc87/pkg/restore/client.go#L594

2. `lightning_checkpoint_engine_order` : there is an engine cleaned up lead to count 2 other than 3 which failed.
https://ci.pingcap.net/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/4877/pipeline/84/

1. add temp variable

2. record all the existing engines into the file, and count the number of different engines

### Check List 

Tests 

 - Unit test

Related changes

### Release note

 -No release note

### DIFF
--- a/pkg/restore/client.go
+++ b/pkg/restore/client.go
@@ -611,7 +611,8 @@ func (rc *Client) RestoreFiles(
 	}
 
 	var rangeFiles []*backuppb.File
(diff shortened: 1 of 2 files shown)


3pointer (approved):
LGTM

glorv (approved):

Issue #1050, opened 2021-04-25
`br_tikv_outage` sometimes got stuck


YuJuncen:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
run `br_tikv_outage`

2. What did you expect to see?
test passed.

3. What did you see instead?
Sometimes, if one TiKV leave permanently, there would be some regions cannot elect the new leader, then the backup progress get stuck.

4. What version of BR and TiDB/TiKV/PD are you using?
Current master.

5. Operation logs
    - [BR](https://github.com/pingcap/br/files/6371225/outage.log)
    - [pd](https://github.com/pingcap/br/files/6371226/pd.log)
    - [tikv1](https://github.com/pingcap/br/files/6371227/tikv1.log)
    - [tikv2](https://github.com/pingcap/br/files/6371228/tikv2.log)
    - [tikv3 (The killed one)](https://github.com/pingcap/br/files/6371229/tikv3.log)

YuJuncen:
Seems this operator didn't finish:

`[2021/04/23 06:33:08.975 +00:00] [INFO] [operator_controller.go:430] ["add operator"] [region-id=257] [operator="\"add-rule-peer {add peer: store [1]} (kind:region,replica, region:257(67,6), createAt:2021-04-23 06:33:08.975017839 +0000 UTC m=+263.830270796, startAt:0001-01-01 00:00:00 +0000 UTC, currentStep:0, steps:[add learner peer 281 on store 1, promote learner peer 281 on store 1 to voter])\""]`

YuJuncen:
When this happens, BR would stuck for about 30mins. If this isn't acceptable for CI, might we need to add a timeout for this test case.

Leavrth:
See the last lines in BR logs, find that BR try backup `region_id=257` over and over again.
We can also find (ctrl+f `region_id=257`) in tikv1 logs that 
`
[2021/04/23 06:32:58.362 +00:00] [INFO] [apply.rs:1256] ["execute admin command"] [command="cmd_type: ChangePeer change_peer { change_type: RemoveNode peer { id: 258 store_id: 1 } }"] [index=8] [term=7] [peer_id=258] [region_id=257]
[2021/04/23 06:32:58.363 +00:00] [INFO] [apply.rs:1626] ["exec ConfChange"] [epoch="conf_ver: 5 version: 67"] [type=RemoveNode] [peer_id=258] [region_id=257]
[...]

PR #1256, opened 2021-06-18
tests: fix br_tikv_outage learner bug

Leavrth:

fix issue #1050 
In the case `outage-after-request`, there is a voter peer on tikv1 removed, and then a learner peer on tikv1 added.
But then quickly in the next case `outage-at-finegrained` (don't restart tikv while doing backup), tikv3 is killed.
So there is only 1 peer and 1 learner for region 257, lead to test got stuck.

It seems the case `outage-after-request` have a bad influence on the next case, so this PR try to exchange the sequence of these cases.

### Check List 
Related changes

### Release note

 -No release note

### DIFF
--- a/tests/br_tikv_outage/run.sh
+++ b/tests/br_tikv_outage/run.sh
@@ -58,7 +58,7 @@ hint_finegrained=$TEST_DIR/hint_finegrained
 hint_backup_start=$TEST_DIR/hint_backup_start
 hint_get_backup_client=$TEST_DIR/hint_get_backup_client
 
-# NOTE : cases `outage-at-finegrained shutdown scale-out` should be former to avoid issue
+# NOTE : cases `outage-at-finegrained shutdown scale-out` should be first to avoid issue
 #       https://github.com/pingcap/br/issues/1050
 cases=${cases:-'outage-at-finegrained shutdown scale-out outage outage-after-request'}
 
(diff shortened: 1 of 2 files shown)


Leavrth:
@YuJuncen PTAL

tests/br_tikv_outage/run.sh
```diff
@@ -58,8 +58,9 @@ hint_finegrained=$TEST_DIR/hint_finegrained
 hint_backup_start=$TEST_DIR/hint_backup_start
 hint_get_backup_client=$TEST_DIR/hint_get_backup_client
 
-
-cases=${cases:-'outage outage-after-request outage-at-finegrained shutdown scale-out'}
+# NOTE : cases `outage-at-finegrained shutdown scale-out` should be former to avoid issue
```
kennytm, line 61:
```suggestion
# NOTE : cases `outage-at-finegrained shutdown scale-out` should be first to avoid issue
```

kennytm (approved):

glorv:
This sounds weird to me, I remember the scripts will recover the cluster to normal state before run each case, thus cases won't have side effect for each other. Are we sure this is the root cause? This case had already been fixed for several times 😭, is this the last time

Leavrth:
> This sounds weird to me, I remember the scripts will recover the cluster to normal state before run each case, thus cases won't have side effect for each other. Are we sure this is the root cause? This case had already been fixed for several times 😭, is this the last time

This test has many bugs which all can make test failed, And I think this is the last bug. 
In fact, the recover in scripts only make sure that the cluster is already initialized before running each case. (This is fixed in other PR #1160 )
As shown above, even though this killed node has been restarted and initialized, there are still some peers in this node hasn't been synchronized with other peers in other nodes, which make these peers keep `learner state`.
At this point, if we kill another node, these `learner state` peers might be failed to synchronize itself, which makes the test timeout.

glorv (approved):

PR #1259, opened 2021-06-21
systable: add more sys tables to blocklist 

3pointer:
… policy of each table

Add more sys tables to blocklist until we define the correct behavior of restore each sys table.

Add more sys tables to blocklist

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

 - No Release note

### DIFF
--- a/pkg/restore/systable_restore.go
+++ b/pkg/restore/systable_restore.go
@@ -33,6 +33,23 @@ var unRecoverableTable = map[string]struct{}{
 	// some variables in tidb (e.g. gc_safe_point) cannot be recovered.
 	"tidb":             {},
 	"global_variables": {},
+
+	// all user related tables cannot be recovered for now.
+	"columns_priv":  {},
+	"db":            {},
+	"default_roles": {},
+	"global_grants": {},
+	"global_priv":   {},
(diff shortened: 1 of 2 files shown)


pkg/restore/systable_restore.go
```diff
@@ -33,6 +33,23 @@ var unRecoverableTable = map[string]struct{}{
...
+	"role_edges":    {},
+	"tables_priv":   {},
+	"user":          {},
+
+	// gc info don't need to recover.
+	"gc_delete_range":      {},
+	"gc_delete_range_done": {},
+
+	// schema_index_usage has table id need to be rewrite.
+	"schema_index_usage": {},
```
kennytm, line 52: should this be considered a kind of `statsTables` then 🙃

kennytm (approved):

pkg/restore/systable_restore.go
3pointer: Maybe, but it's just record the last use time of one index. I think the priority of support this table is lower than `statsTables`. because it doesn't have any side effects.

glorv (approved):

PR #1272, opened 2021-06-23
lightning: fix parquet parser for decimal type

glorv:
Fix parse parquet file when the corresponding type is decimal with `FIXED_LEN_BYTE_ARRAY` or `BINARY` 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release note

 - Fix parquet parse when parse decimal type


pkg/lightning/mydump/parquet_parser.go
```diff
@@ -272,67 +372,130 @@ func (pp *ParquetParser) ReadRow() error {
...
 	case reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
 		d.SetUint64(v.Uint())
 	case reflect.Int8, reflect.Int16:
 		d.SetInt64(v.Int())
 	case reflect.Int32, reflect.Int64:
-		setDatumByInt(d, v.Int(), meta)
+		return setDatumByInt(d, v.Int(), meta)
 	case reflect.String:
-		d.SetString(v.String(), "")
+		setDatumByString(d, v.String(), meta)
```
3pointer, line 394: missing return？

pkg/lightning/mydump/parquet_parser.go
```diff
@@ -272,67 +372,130 @@ func (pp *ParquetParser) ReadRow() error {
...
 		if v.IsNil() {
 			d.SetNull()
 		} else {
-			setDatumValue(d, v.Elem(), meta)
+			return setDatumValue(d, v.Elem(), meta)
 		}
 	default:
 		log.L().Fatal("unknown value", zap.Stringer("kind", v.Kind()),
 			zap.String("type", v.Type().Name()), zap.Reflect("value", v.Interface()))
+		return errors.Errorf("unknown value: %v", v)
```
gozssky, line 406: unreadable `return` after `Fatal`

pkg/lightning/mydump/parquet_parser.go
```diff
@@ -272,67 +372,130 @@ func (pp *ParquetParser) ReadRow() error {
...
+			nsec = (v % 1e6) * 1e3
+		} else if units.MILLIS != nil {
+			sec = v / 1e3
+			nsec = (v % 1e3) * 1e6
+		} else {
+			// nano
+			sec = v / 1e9
+			nsec = v % 1e9
+		}
+		// TODO: how to deal with TimeZone
```
gozssky, line 491: Is there more we need to do about time zones?

pkg/lightning/mydump/parquet_parser.go
```diff
@@ -272,67 +372,130 @@ func (pp *ParquetParser) ReadRow() error {
...
+	val := fmt.Sprintf("%0*d", scale, intValue)
+	dotIndex := len(val) - scale
+	var res strings.Builder
+	if negative {
+		res.WriteByte('-')
+	}
+	if dotIndex == 0 {
+		res.WriteByte('0')
+	} else {
+		res.Write([]byte(val[:dotIndex]))
```
kennytm, line 443:
```suggestion
		res.WriteString(val[:dotIndex])
```

pkg/lightning/mydump/parquet_parser.go
```diff
@@ -272,67 +372,130 @@ func (pp *ParquetParser) ReadRow() error {
...
 	}
+	if scale > 0 {
+		res.WriteByte('.')
+		res.Write([]byte(val[dotIndex:]))
```
kennytm, line 447:
```suggestion
		res.WriteString(val[dotIndex:])
```

kennytm (commented):
rest LGTM

gozssky (commented):
rest LGTM

pkg/lightning/mydump/parquet_parser.go
glorv: The time/timestamp value read from parquet is either UTC or Local timezone (based on the `IsAdjustedToUTC` setting). Since currently lightning don't support set timezone (always set to target cluster's timezone). Not sure how should we deal with the unknown timezone offset if `IsAdjustedToUTC = false` (at least the default is true). @kennytm  PTAL

pkg/lightning/mydump/parquet_parser_test.go
```diff
@@ -215,3 +208,58 @@ func (s testParquetParserSuite) TestParquetVariousTypes(c *C) {
...
+		expectedValues := expectedRes[i]
+		row := parser.LastRow().Row
+		c.Assert(len(expectedValues), Equals, len(row))
+		for j := 0; j < len(row); j++ {
+			switch v := expectedValues[j].(type) {
+			case int64:
+				c.Assert(v, Equals, row[j].GetInt64())
+			case string:
+				c.Assert(v, Equals, row[j].GetString())
+			}
```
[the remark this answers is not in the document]
glorv: done

pkg/lightning/mydump/parquet_parser.go
```diff
@@ -272,67 +376,130 @@ func (pp *ParquetParser) ReadRow() error {
...
+			nsec = (v % 1e6) * 1e3
+		} else if units.MILLIS != nil {
+			sec = v / 1e3
+			nsec = (v % 1e3) * 1e6
+		} else {
+			// nano
+			sec = v / 1e9
+			nsec = v % 1e9
+		}
+		// TODO: how to deal with TimeZone if `IsAdjustedToUTC = false`
```
glorv, line 495: @kennytm PTAL here. Not sure how to deal with it and I think there is not much use case when `IsAdjustedToUTC = false`

glorv: related document: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#instant-semantics-timestamps-normalized-to-utc

pkg/lightning/mydump/parquet_parser.go
kennytm: Make the produced string `2006-01-02 15:04:05.999999Z` when isAdjustedToUTC is true, and `2006-01-02 15:04:05.999999` when it is false.

(why do we only support 3 fractional digits instead of 6?)

lonng:
@glorv It's better file an issue to record this bug.

pkg/lightning/mydump/parquet_parser.go
glorv: done

pkg/lightning/mydump/parquet_parser.go
```diff
@@ -479,42 +479,40 @@ func setDatumByInt(d *types.Datum, v int64, meta *parquet.SchemaElement) error {
...
+	} else {
+		// nano
+		sec = v / 1e9
+		nsec = v % 1e9
+	}
+	t := time.Unix(sec, nsec)
+	if utc {
+		return t.UTC().Format(utcFormat)
+	}
+	return t.Format(format)
```
gozssky, line 513:
```golang	
// IsAdjustedToUTC == false
fmt.Println(time.Unix(86400, 0).Format("2006-01-02 15:04:05.999999")) // 1970-01-02 08:00:00
// IsAdjustedToUTC == true
fmt.Println(time.Unix(86400, 0).UTC().Format("2006-01-02 15:04:05.999999Z")) // 1970-01-02 00:00:00Z
```
When `IsAdjustedToUTC` is false, it seems the reference timestamp is still `1970-01-01 00:00:00 UTC`. Is it expected?

glorv: From the parquet document, with `IsAdjustedToUTC == false` it should be `1970-01-02 00:00:00`, so the only different should be whether to add the suffix `Z`, thanks @gozssky

glorv: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#local-semantics-timestamps-not-normalized-to-utc

gozssky (approved):

3pointer (approved):

kennytm (approved):

Issue #1270, opened 2021-06-23
br restore statistics not correct if restore with -f parameters


fubinzh:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
  1. Import some data (5GB) to table test.oomtest, and there is no user created tables in mysql schema
  2. Do full backup
  3. Do br restore, and specify -f "mysql.*" parameter to restore system tables only

2. What did you expect to see?
- no tables should be restored, as there is no user created tables in mysql schema, and restore statistics should be accurate.

3. What did you see instead?
- Restore statistics not accurate,  7.426GB seems the backup size, not the restored data size.
- Also, it seems that restore statistics is not correctly when restore part of the backup. (for example, there are 2 tables in backup, and restore only one tables.)

[2021/06/23 01:56:29.793 -04:00] [INFO] [collector.go:66] ["Full restore success summary"] [total-ranges=44] [ranges-succeed=44] [ranges-failed=0] [split-region=601.358689ms] [restore-checksum=43.793886ms] [restore-ranges=19] [total-take=6.156156669s] ["restore data size(after decompressed)"=7.426GB] [total-kv=3345] [total-kv-size=287kB] [average-speed=584.5kB/s]

[root@localhost v5.1.0]# ./br restore full --pd "172.16.6.226:2379" --storage "s3://nfs/fubin/br0623?access-key=minioadmin&secret-access-key=minioadmin&endpoint=http%3a%2f%2fminio.pingcap.net%3a9000&force-path-style=true" --send-credentials-to-tikv=true -f "mysql*.*" --log-file restore.system.2.log
Detail BR log in restore.system.2.log
Full restore <---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|> 100.00%[2021/06/23 01:56:29.111 -04:00] [WARN] [systable_restore.go:94] ["error during merging temporary tables into system tables"] [error="[BR:Restore:ErrUnsupportedSysTable]restoring stats via `mysql` schema isn't support yet: the table ID is out-of-date and may corrupt existing statistics"] [table=stats_meta]
[2021/06/23 01:56:29.111 -04:00] [WARN] [systable_restore.go:94] ["error during merging temporary tables into system tables"] [error="[BR:Restore:ErrUnsupportedSysTable]restoring stats via `mysql` schema isn't support yet: the table ID is out-of-date and may corrupt existing statistics"] [table=stats_buckets]
[...]

PR #1285, opened 2021-06-25
restore: Fix restore data size

zwj-coder:
close issue [#1270](https://github.com/pingcap/br/issues/1270)
The previous summary field `restore data size(after compressed)` is not correct and will count the files that will not be restored. This pr solve this problem.

Add filter in size account

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### Release note

 - No Release Note


pkg/utils/schema.go
```diff
@@ -79,10 +79,20 @@ func LoadBackupTables(ctx context.Context, reader *metautil.MetaReader) (map[str
 }
 
 // ArchiveSize returns the total size of the backup archive.
-func ArchiveSize(meta *backuppb.BackupMeta) uint64 {
+func ArchiveSize(meta *backuppb.BackupMeta, files []*backuppb.File) uint64 {
 	total := uint64(meta.Size())
+	ContainsFile := func(files []*backuppb.File, file *backuppb.File) bool {
```
3pointer, line 84: how about use a files map?

zwj-coder: I think the number of files will not be so much, so O(n) seems enough?

zwj-coder: Since rocksdb using level compaction, the number of SST files will be restricted.

pkg/utils/schema.go
```diff
@@ -79,10 +79,20 @@ func LoadBackupTables(ctx context.Context, reader *metautil.MetaReader) (map[str
...
+		for i := 0; i < len(files); i++ {
+			if files[i].GetName() == file.GetName() {
+				return true
+			}
+		}
+		return false
+	}
 	for _, file := range meta.Files {
```
YuJuncen, line 92: Maybe we should consider backupmeta v2 here? (i.e. use `metautil.readDataFiles`, eh, which is private for now. 🤔)

pkg/metautil/metafile.go
```diff
@@ -638,9 +657,6 @@ func (writer *MetaWriter) ArchiveSize() uint64 {
 	for _, file := range writer.backupMeta.Files {
 		total += file.Size_
 	}
-	for _, size := range writer.metafileSizes {
```
YuJuncen, line 641: I'm wondering why remove this...? Seems the `for file in writer.backupMeta.Files` for backupmeta V1, and this for loop only for V2. It seems when using V1, `writer.metafileSizes` wouldn't even be accessed.

YuJuncen (commented):
Rest LGTM, and if possible, could you add an unit test for `reader.ArchiveSize` method?

pkg/metautil/metafile.go
zwj-coder: @YuJuncen  Field `metafileSizes` seems the size of the inner index files, which should not be counted, I think only the size of the data files should be counted. If we count the index file size in backup, how we count their size when we only restore parts of the backup data.

YuJuncen: For the data files, `metafaleSizes` was accumulated here.
https://github.com/pingcap/br/blob/52c2bbae806207cd0c96091b89eef66d2e6703e0/pkg/metautil/metafile.go#L614-L615

Where the `writer.metafiles.size` would be collected by this:
https://github.com/pingcap/br/blob/52c2bbae806207cd0c96091b89eef66d2e6703e0/pkg/metautil/metafile.go#L394-L399

Let `op` be `AppendDataFile`, the `size` should be the file size:
https://github.com/pingcap/br/blob/52c2bbae806207cd0c96091b89eef66d2e6703e0/pkg/metautil/metafile.go#L351-L357

So, maybe only the `metafileSizes["metafile"]` is redundant? Also cc @3pointer.

zwj-coder: I think there is another issue that in meta v1, we only count the data file size, and in meta v2 we count data file size, schema size and Ddls size. It may cause inconsistency behavior.

YuJuncen: Well, so I guess `metafileSizes["datafile"]` is enougth?

pkg/metautil/metafile.go
```diff
@@ -156,6 +156,25 @@ func (reader *MetaReader) readDataFiles(ctx context.Context, output func(*backup
 	return walkLeafMetaFile(ctx, reader.storage, reader.backupMeta.FileIndex, outputFn)
 }
 
+// ArchiveSize return the size of Archive data
+func (reader *MetaReader) ArchiveSize(ctx context.Context, files []*backuppb.File) (uint64, error) {
+	total := uint64(0)
```
3pointer, line 161: can we use the files to calculate totalSize?
```
for _, f := range files {
    total += f.size
}
```

pkg/metautil/metafile.go
```diff
@@ -212,14 +224,17 @@ func (reader *MetaReader) ReadSchemasFiles(ctx context.Context, output chan<- *T
...
 	outputFn := func(file *backuppb.File) {
 		tableID := tablecodec.DecodeTableID(file.GetStartKey())
 		if tableID == 0 {
 			log.Panic("tableID must not equal to 0", logutil.File(file))
 		}
 		fileMap[tableID] = append(fileMap[tableID], file)
+		fileSizeMap[file.Name] = file.Size_
 	}
 	err := reader.readDataFiles(ctx, outputFn)
+	reader.FileSizeMap = fileSizeMap
```
YuJuncen, line 237: I think we'd better initialize this field at `NewMetaReader`, for preventing possible access to `nil` map.

pkg/metautil/metafile.go
```diff
@@ -634,12 +649,14 @@ func (writer *MetaWriter) flushMetasV2(ctx context.Context, op AppendOp) error {
...
+	total := uint64(0)
 	for _, file := range writer.backupMeta.Files {
 		total += file.Size_
 	}
-	for _, size := range writer.metafileSizes {
-		total += uint64(size)
+	for name, size := range writer.metafileSizes {
+		if name == "datafile" {
+			total += uint64(size)
+		}
```
YuJuncen, line 659:
```suggestion
	total += uint64(writer.metafileSizes["datafile"])
```

tests/br_300_small_tables/run.sh
```diff
@@ -56,6 +76,24 @@ while [ $i -le $TABLES_COUNT ]; do
     i=$(($i+1))
 done
 
+rm -rf $RESTORE_LOG
+echo "restore 1/300 of the table start..."
+run_br restore table --db $DB  --table "sbtest100" --log-file $RESTORE_LOG -s "local://$TEST_DIR/$DB" --pd $PD_ADDR --no-schema
+restore_size=`tail -n 2 ${RESTORE_LOG} | grep "restore data size" | grep -oP '\[\K[^\]]+' | grep "restore data size" | awk -F '=' '{print $2}' | grep -oP '\d*\.\d+'`
+echo ${restore_size}
+
+diff=`echo "${backup_size}-${restore_size}*${TABLES_COUNT}" | bc`
```
YuJuncen, line 85: Just a hint, the current version is also fine.

In bash, You can use [Shell Arithmetic](https://www.gnu.org/software/bash/manual/html_node/Arithmetic-Expansion.html#Arithmetic-Expansion) for cleaner arithmetic expressions, like:

```suggestion
diff=$((backup_size-restore_size*TABLES_COUNT))
```

YuJuncen (commented):
rest LGTM

YuJuncen (approved):
LGTM

3pointer (approved):

PR #1297, opened 2021-06-29
lightning/mydump: support mydumper.csv.terminator

kennytm:

Allow user to specify a custom CSV line terminator (corresponding to LOAD DATA's LINES TERMINATED BY option.)

* Added a config `mydumper.csv.terminator`. When it is not empty, this will be used as the line separator instead of `[\r\n]`.
* The parser has been refactored to use a more tokenizer-based approach to get rid of all the `peekBytes` stuff randomly appearing somewhere.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note

 - Lightning now supports customized line ending other than `\r`/`\n` in CSV files.


overvenus (approved):
LGTM

pkg/lightning/mydump/csv_parser.go
```diff
@@ -297,10 +372,13 @@ outside:
 			parser.fieldIndexes = append(parser.fieldIndexes, len(parser.recordBuffer))
 			break outside
 		default:
-			if err = processDefault(firstByte); err != nil {
-				return nil, err
+			if prevToken == csvTokenDelimiter {
+				parser.logSyntaxError()
+				return nil, errors.AddStack(errUnexpectedQuoteField)
 			}
+			parser.appendCSVTokenToRecordBuffer(firstToken)
```
glorv, line 379: If we enter here, we only allow to read a valid unquote field, but this logic will allow read csv with broken syntax. e.g.
```
delimiter = "|+|"
sep = ","
input = "abc|1|+||+|\r\n"

this input should raise an error because delimiter appear in the middle of an unquoted string
```

pkg/lightning/mydump/csv_parser.go
```diff
@@ -297,10 +372,13 @@ outside:
...
 		}
+		prevToken = firstToken
```
glorv, line 381: This assign seems wrong if firstToken is delimiter

pkg/lightning/config/config.go
```diff
@@ -535,13 +536,20 @@ func (cfg *Config) Adjust(ctx context.Context) error {
 		return errors.New("invalid config: `mydumper.csv.separator` and `mydumper.csv.delimiter` must not be prefix of each other")
 	}
 
+	if len(csv.Terminator) > 0 && cfg.Mydumper.StrictFormat {
```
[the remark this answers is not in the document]
kennytm: @glorv Terminator and StrictFormat are currently mutually exclusive because the latter is hard-coded to use `[\r\n]` for now. There's no problem using TrimLastSep and Terminator together.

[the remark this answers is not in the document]
glorv: But this will trim the last two seps, is there any scenario that actually need this? BTW, do we want to finally replace `TrimLastSep` with `Terminator` or they just two different feature?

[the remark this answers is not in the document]
kennytm: the original purpose of TrimLastSep was to support importing [TPC-H generated data](https://docs.pingcap.com/tidb/stable/migrate-from-csv-using-tidb-lightning#tpc-h-dbgen) in the form

```
a|b|c|
d|e|f|
```

, which we can use `terminator = "|\n"` instead. 

i don't know if there are any other use cases 😅

pkg/lightning/mydump/csv_parser.go
kennytm: if `firstToken` is csvTokenDelimiter at line 348 and no errors in lines 353-359, this assignment will set `prevToken` to csvTokenDelimiter and go on to next step. not sure what is wrong.

glorv: I see. In this case, prevToken should be csvTokenCloseDelimiter so there is no different.

glorv (approved):

PR #1304, opened 2021-07-01
lightning: save status checkpoint synchronously

gozssky:

We save status checkpoint after import finished, and then clean up the engine. 
https://github.com/pingcap/br/blob/9933ebd56bedfa8bd5d06e632741e4ebecafab8b/pkg/lightning/restore/restore.go#L2748-L2752
But `saveStatusCheckpoint`  saves checkpoint asynchronously, an engine may be cleaned up successfully but exit before update checkpoint. After restart, `verifyLocalFile` will fail.
https://github.com/pingcap/br/blob/9933ebd56bedfa8bd5d06e632741e4ebecafab8b/pkg/lightning/restore/restore.go#L814

To solve this issue, we can save status synchronously. An engine can only be cleaned up after the checkpoint has been successfully saved. Since we don't change the status of the engine very often, I think synchronously saving status checkpoint is acceptable and it can also avoid some unusual situations.

Related unstable integration test: [lightning_checkpoint_engines_order](https://ci.pingcap.net/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/5389/pipeline). 

Save save status checkpoint synchronously.

### Check List 

Tests 

 - Unit test
 - Integration test
 
Code changes

Side effects

Related changes

### Release note

No release note.


pkg/lightning/restore/restore.go
```diff
@@ -911,7 +922,14 @@ func (rc *Controller) saveStatusCheckpoint(tableName string, engineID int32, err
 		metric.RecordEngineCount(statusIfSucceed.MetricName(), err)
 	}
 
-	rc.saveCpCh <- saveCp{tableName: tableName, merger: merger}
+	waitCh := make(chan error, 1)
+	rc.saveCpCh <- saveCp{tableName: tableName, merger: merger, waitCh: waitCh}
+
+	if err = <-waitCh; err != nil {
```
kennytm, line 928: will this cause Lightning unable to exit by Ctrl+C since we're blocked here

gozssky: It might be possible if `checkpoints.DB.Update` is blocked too long.  I will consider adding a ctx parameter to this method later.

pkg/lightning/restore/restore.go
```diff
@@ -923,7 +934,20 @@ func (rc *Controller) saveStatusCheckpoint(tableName string, engineID int32, err
...
 
-	rc.saveCpCh <- saveCp{tableName: tableName, merger: merger}
+	waitCh := make(chan error, 1)
+	rc.saveCpCh <- saveCp{tableName: tableName, merger: merger, waitCh: waitCh}
+
+	var saveCpErr error
+	select {
+	case saveCpErr = <-waitCh:
+	case <-ctx.Done():
+		saveCpErr = ctx.Err()
```
glorv, line 944: I think we can just return `ctx.Err()` here to avoid output an error log with context canceled error since it is useless and annoying

pkg/lightning/restore/restore.go
```diff
@@ -941,10 +966,21 @@ func (rc *Controller) listenCheckpointUpdates() {
...
+			failpoint.Inject("SlowDownCheckpointUpdate", func() {})
+
 			if len(cpd) > 0 {
-				rc.checkpointsDB.Update(cpd)
+				err := rc.checkpointsDB.Update(cpd)
+				for _, w := range ws {
+					w <- err
+				}
+				if err != nil {
+					log.L().Error("save checkpoint failed", zap.Error(err))
```
glorv, line 982: I think we needn't output this error log here, the caller should take responsibility to handle its save cp failed event

glorv (commented):
rest LGTM

pkg/lightning/restore/restore.go
```diff
@@ -1597,6 +1630,17 @@ func (tr *TableRestore) restoreEngines(pCtx context.Context, rc *Controller, cp
 	if indexEngineCp == nil {
 		return errors.Errorf("table %v index engine checkpoint not found", tr.tableName)
 	}
+	// If there is an index engine only, it indicates no data needs to restore.
+	// So we can change status to imported directly and avoid opening engine.
+	if len(cp.Engines) == 1 {
```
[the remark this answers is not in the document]
gozssky: I think it's better to update the checkpoint so that the checkpoint status is consistent.

gozssky:
@glorv @kennytm PTAL again, thanks

PR #1325, opened 2021-07-07
Update range-concurrency default value

glkappe:
modify range-concurrency default value 16 change to 24，10 Gigabit network card, 16 may cause the network card to be relatively idle.
@glorv

### Check List 

Tests 

Code changes

Side effects

Related changes

### Release note

 - No release note


kennytm:
Lightning doesn't read the default values from the template TOML file...

glkappe:
@glorv  please take a look🤭🤭🤭

gozssky:
I think we can configure `range-concurrency` for specific environment rather than changing the default value.

Another approach may be that lightning should detect the environment and automatically set proper value when this field is not set. 🤔

glkappe:
> I think we can configure `range-concurrency` for specific environment rather than changing the default value.
> 
> Another approach may be that lightning should detect the environment and automatically set proper value when this field is not set. 🤔

we hope it is

glorv:
16 is a relatively low setting with 10Gbps network (the most common case) when ingest is not so fast which is caused by severe SST overlap and heavy compaction. So I think it's ok to increase the range-concurrency from 16 to 24. Not sure if there is an easy way to check the environment.

glorv:
@glkappe I'd like to close this PR since it is incomplete and the root cause can't be resolved by only changing the default value. And we will continue to find optimized ways to provide a better default value.

Issue #1322, opened 2021-07-07
lightning: "tidb lightning encountered error:  EOF"


cyliu0:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
Import csv data.

2. What did you expect to see?
Import successfully

3. What did you see instead?
```
/ # /tidb-lightning -d "s3://tpcc/10-warehouses-csv?access-key=minioadmin&secret-access-key=minioadmin&endpoint=http%3a%2f%2fminio.ping
cap.net%3a9000&force-path-style=true" -pd-urls "target-pd.testbed-2h8mp.svc:2379" -tidb-host target-tidb.testbed-2h8mp.svc -tidb-port 4
000 -tidb-user root -backend local -sorted-kv-dir /tmp/sorted-kv-dir -L debug
Verbose debug logs will be written to /tmp/lightning.log.2021-07-07T10.15.31Z

Error: EOF
tidb lightning encountered error:  EOF
/ # cat /tmp/lightning.log.2021-07-07T10.15.31Z
[2021/07/07 10:15:31.884 +00:00] [INFO] [info.go:40] ["Welcome to TiDB-Lightning"] [release-version=v5.0.0-master] [git-hash=df27ac2045fa5f3942bae79a0aa20b58689145bf] [git-branch=master] [go-version=go1.16.4] [utc-build-time="2021-07-06 04:28:09"] [race-enabled=false]
[...]

PR #1331, opened 2021-07-09
lightning: prettify data check error output

gozssky:

Prettify error output for case https://github.com/pingcap/br/issues/1322 and fix a few minor bugs by the way.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### Release note
No release note.


pkg/lightning/restore/check_info.go
```diff
@@ -372,9 +365,19 @@ func (rc *Controller) SchemaIsValid(ctx context.Context, tableInfo *md.MDTableMe
...
 		}
-		err = parser.ReadRow()
-		if err != nil {
+		defer parser.Close()
+
+		if err = parser.ReadRow(); err != nil {
+			if errors.Cause(err) == io.EOF {
+				msg := fmt.Sprintf("file '%s' doesn't have a data row", dataFileMeta.Path)
+				if dataFileMeta.Type == md.SourceTypeCSV {
+					msg += fmt.Sprintf(" (first row is parsed as header: %v)", rc.cfg.Mydumper.CSV.Header)
```
glorv, line 376: I think we should emit this message if `rc.cfg.Mydumper.CSV.Header = false`. BTW, if the source file is empty or only contains the header line, I think it's better to skip this check instead of returning an error. What do you think @3pointer

pkg/lightning/restore/check_info.go
```diff
@@ -294,67 +331,49 @@ func (rc *Controller) SchemaIsValid(ctx context.Context, tableInfo *md.MDTableMe
...
-		colCountFromTiDB := len(info.Core.Columns)
-		core := info.Core
-		defaultCols := make(map[string]struct{})
-		for _, col := range core.Columns {
-			if hasDefault(col) || (info.Core.ContainsAutoRandomBits() && mysql.HasPriKeyFlag(col.Flag)) {
-				// this column has default value or it's auto random id, so we can ignore it
-				defaultCols[col.Name.L] = struct{}{}
-			}
+		if colsFromDataFile == nil && colCountFromDataFile == 0 {
+			log.Info("file doesn't contain columns and rows, skip check it", zap.String("path", dataFileMeta.Path))
```
kennytm, line 374:
```suggestion
			log.Info("file is empty, skip checking against schema validity", zap.String("path", dataFileMeta.Path))
```

kennytm (approved):

glorv (approved):

PR #1332, opened 2021-07-12
*: refine some code and comments according to golangci-lint

JmPotato:

I use golangci-lint 1.41.1 to run:

```shell
golangci-lint run
```

According to the results, I refined some code and comments, such as:

* Inefficient assignment
* Code complexity
* Misspelling
* Deadcode

See above.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

 - No release note

### DIFF
--- a/pkg/lightning/restore/meta_manager.go
+++ b/pkg/lightning/restore/meta_manager.go
@@ -535,6 +535,7 @@ func (m *dbTaskMetaMgr) CheckAndPausePdSchedulers(ctx context.Context) (pdutil.U
 	pauseCtx, cancel := context.WithCancel(ctx)
 	conn, err := m.session.Conn(ctx)
 	if err != nil {
+		cancel()
 		return nil, errors.Trace(err)
 	}
 	defer conn.Close()
@@ -544,6 +545,7 @@ func (m *dbTaskMetaMgr) CheckAndPausePdSchedulers(ctx context.Context) (pdutil.U
 	}
 	err = exec.Exec(ctx, "enable pessimistic transaction", "SET SESSION tidb_txn_mode = 'pessimistic';")
 	if err != nil {
+		cancel()
 		return nil, errors.Annotate(err, "enable pessimistic transaction failed")
 	}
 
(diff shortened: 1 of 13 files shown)


pkg/lightning/restore/meta_manager.go
```diff
@@ -535,6 +535,7 @@ func (m *dbTaskMetaMgr) CheckAndPausePdSchedulers(ctx context.Context) (pdutil.U
 	pauseCtx, cancel := context.WithCancel(ctx)
 	conn, err := m.session.Conn(ctx)
 	if err != nil {
+		cancel()
```
glorv, line 538: perfer call `defer cancel()` afer L535

pkg/restore/client.go
```diff
@@ -181,7 +181,7 @@ func (rc *Client) InitBackupMeta(c context.Context, backupMeta *backuppb.BackupM
 
 		var ddlJobs []*model.Job
 		// ddls is the bytes of json.Marshal
-		ddls, err := reader.ReadDDLs(c)
+		ddls, _ := reader.ReadDDLs(c)
```
glorv, line 184: shall we check insted of ingore of ignore this error? @3pointer

glorv (commented):
rest LGTM

tisonkun:
@JmPotato where does this change set come from? It seems strange a PR with title "make some changes".

JmPotato:
> @JmPotato where does this change set come from? It seems strange a PR with title "make some changes".

Updated the description. PTAL.

pkg/lightning/restore/meta_manager.go
JmPotato: This function returns the cancel function to the caller. So I think `defer` should not be used here.

https://github.com/pingcap/br/blob/df27ac2045fa5f3942bae79a0aa20b58689145bf/pkg/lightning/restore/meta_manager.go#L639-L643

nolouch (commented):
lgtm about the lint.

pkg/restore/client.go
overvenus: We should check the error.

overvenus (commented):
Rest LGTM

JmPotato:
@overvenus @glorv PTAL, thx!

overvenus (approved):
LGTM, thanka!

glorv (approved):

PR #1335, opened 2021-07-12
backup: refine backup log for tracing backup tasks

YuJuncen:
We send backup requests concurrently for better performance, but the log of them got mixed and make tracing the lifecycle of backing up a range harder. Like this:
image
(4 parallel backup tasks were started, and consequent `try backup` logs were mixed and hard to trace which task it was from.) 

Some new utils `ContextWithFields` and `CL(ctx)` were introduced. The former one allows us to bind some fields into the contextual logger, and the latter one retrieves the contextual logger.
During backing up, each range split by `BuildBackupRangeAndSchema` would have a unique `range-sn`, this number would be added to the contextual logger and hence we can trace a backup task trivially with something like `grep`, as below:
image

### Check List 

Tests 

 - Unit test

### Release note

 - Refined the log for better tracing backup tasks.


YuJuncen:
/build 
Why `github.com/pingcap/br/pkg/lightning/restore` failed... 🤔

pkg/backup/client.go
```diff
@@ -840,6 +841,7 @@ func (bc *Client) handleFineGrained(
 // Stop receiving response if respFn returns error.
 func SendBackup(
 	ctx context.Context,
+	// the `storeID` seems only used for logging now, maybe we can remove it then?
```
3pointer, line 844: Do you mean put `storeID` into context🤔

YuJuncen: Yep, if it was only for logging...

YuJuncen:
```
Error: RequestError: send request failed caused by: Put "http://127.0.0.1:24927/mybucket/br_s3/backupmeta": read tcp 127.0.0.1:60432->127.0.0.1:24927: read: connection reset by peer
```

When BR failed before the fail point triggered, `br_s3` may get stuck. (2hrs for this case)🤔

/run-integration-tests

3pointer (approved):

pkg/backup/client.go
```diff
@@ -423,10 +423,12 @@ func (bc *Client) BackupRanges(
 	// we collect all files in a single goroutine to avoid thread safety issues.
 	workerPool := utils.NewWorkerPool(concurrency, "Ranges")
 	eg, ectx := errgroup.WithContext(ctx)
-	for _, r := range ranges {
+	for id, r := range ranges {
+		id := id
```
joccau, line 427: Is it redundant?

YuJuncen: Nope, we need make `id` a local variable so the closure won't capture its change, which is a common trap in Go🤯.

See [this playground](https://play.golang.org/p/OyMjs3Bl4su).
```go
const max = 6

func main() {
	wg := new(sync.WaitGroup)
	wg.Add(max)
	for i := 0; i < max; i++ {
		go func() { fmt.Printf("%d", i); wg.Done() }()
	}
	wg.Wait()
	// output: 656666
}
```

joccau (approved):
LGTM

zwj-coder (approved):

YuJuncen:
/merge

Even I'm wondering whether this need to be rewritten at TiDB.

PR #1336, opened 2021-07-12
lightning: check and restore pd scheduler even if our task failed

glorv:
Fix the bug that in concurrency import mode, if one or more lightning failed before import all tables finished, pd schedulers won't be restored.

This PR adds a new status to mark that one task is exit before finishing. A lightning instance will set its status to one of (checksum_skipping, checksumming, unfinished), other lightinng can check these statuses to make sure whether it is the last running instance. The last instance should recover the pd schedulers configs.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note

 - Fix the bug that in concurrency mode, pd schedulers may not be restored if one or more lightning failed.


pkg/lightning/restore/meta_manager.go
```diff
@@ -525,9 +543,34 @@ func (m *dbTaskMetaMgr) InitTask(ctx context.Context) error {
 		DB:     m.session,
 		Logger: log.L(),
 	}
-	// avoid override existing metadata if the meta is already inserted.
-	stmt := fmt.Sprintf(`INSERT IGNORE INTO %s (task_id, status) values (?, ?)`, m.tableName)
-	err := exec.Exec(ctx, "init task meta", stmt, m.taskID, taskMetaStatusInitial.String())
+
+	err := exec.Transact(ctx, "check and init task status", func(ctx context.Context, tx *sql.Tx) error {
+		// avoid override existing metadata if the meta is already inserted.
+		stmt := fmt.Sprintf(`INSERT IGNORE INTO %s (task_id, status) values (?, ?)`, m.tableName)
```
Little-Wallace, line 549: Why insert be fore select......

pkg/lightning/restore/restore.go
```diff
@@ -1435,6 +1439,7 @@ func (rc *Controller) restoreTables(ctx context.Context) error {
 	// finishSchedulers()
 	// cancelFunc(switchBack)
 	// finishFuncCalled = true
+	taskFinished = true
```
Little-Wallace, line 1442: What is it used for?

pkg/lightning/restore/meta_manager.go
glorv: Ensure there is exact one row represent current lightning. Need the reset the task status if current task's status is `exit_unfinished`

pkg/lightning/restore/restore.go
glorv: If current task is exit before lightning finished (maybe met error or by user terminating), we should clean up the task/table meta tables if all other lightning are finished

Little-Wallace: Oh, I see. `finishSchedulers` will be called before this method ending......  Emmmmmm, it seems to be difficult to understand

pkg/lightning/restore/meta_manager.go
```diff
@@ -483,36 +483,54 @@ const (
 	taskMetaStatusScheduleSet
 	taskMetaStatusSwitchSkipped
 	taskMetaStatusSwitchBack
+	// the task exit before finishing
+	taskMetaStatusExitUnfinished taskMetaStatus = 8
 )
 
+func (m taskMetaStatus) realStatus() taskMetaStatus {
+	return m & (taskMetaStatusExitUnfinished - 1)
```
Little-Wallace, line 491: Why define a method `realstatus` ?

pkg/lightning/restore/meta_manager.go
```diff
@@ -462,7 +462,10 @@ func (m *dbTableMetaMgr) FinishTable(ctx context.Context) error {
 type taskMetaMgr interface {
 	InitTask(ctx context.Context) error
 	CheckAndPausePdSchedulers(ctx context.Context) (pdutil.UndoFunc, error)
-	CheckAndFinishRestore(ctx context.Context) (bool, error)
+	// CheckAndFinishRestore check task meta and return whether to switch cluster to normal state and clean up the metadata
+	// Return values: first boolean indicates whether switch back tidb cluster to normal state (restore schedulers, switch tikv to normal)
+	// the second boolean indicates whether to clean up the metadata in tidb
+	CheckAndFinishRestore(ctx context.Context, finished bool) (bool, bool, error)
```
kennytm, line 468: ya know you could give names to the returned values.

```suggestion
	CheckAndFinishRestore(ctx context.Context, finished bool) (shouldSwitchBack bool, shouldCleanUpMeta bool, err error)
```

Little-Wallace: Agree

Little-Wallace (approved):
LGTM

pkg/lightning/restore/restore.go
```diff
@@ -1435,6 +1440,7 @@ func (rc *Controller) restoreTables(ctx context.Context) error {
 	// finishSchedulers()
 	// cancelFunc(switchBack)
 	// finishFuncCalled = true
```
overvenus, line 1442: Need clean up?

glorv: We originally want to restore pd schedulers and switch back tikv to normal mode after data import finished. Then the cluster can do possible rebalance during checksum and analyze. But In our test, these rebalance will bring non-trivial impact to checksum and analyze. So we need to investigate further to determine whether we can still do this. So I think we can keep these before coming up with a clear conclusion.

pkg/lightning/restore/meta_manager.go
```diff
@@ -525,9 +533,13 @@ func (m *dbTaskMetaMgr) InitTask(ctx context.Context) error {
 		DB:     m.session,
 		Logger: log.L(),
 	}
-	// avoid override existing metadata if the meta is already inserted.
-	stmt := fmt.Sprintf(`INSERT IGNORE INTO %s (task_id, status) values (?, ?)`, m.tableName)
-	err := exec.Exec(ctx, "init task meta", stmt, m.taskID, taskMetaStatusInitial.String())
+
+	err := exec.Transact(ctx, "check and init task status", func(ctx context.Context, tx *sql.Tx) error {
+		// avoid override existing metadata if the meta is already inserted.
+		stmt := fmt.Sprintf(`INSERT INTO %s (task_id, status) values (?, ?) ON DUPLICATE KEY UPDATE state = ?`, m.tableName)
```
kennytm, line 539: the comment feels outdated.

pkg/lightning/restore/restore.go
```diff
@@ -105,6 +105,7 @@ const (
 		task_id BIGINT(20) UNSIGNED NOT NULL,
 		pd_cfgs VARCHAR(2048) NOT NULL DEFAULT '',
 		status  VARCHAR(32) NOT NULL,
+		state   TINYINT(1) NOT NULL DEFAULT 0 COMMENT '0: normal, 1: exited before finish',
```
kennytm, line 108: should the `taskMetaTableName` be changed?

otherwise if we used Lightning before, and this CREATE TABLE IF NOT EXISTS means the `task_meta` with the *old* table structure will be used

kennytm (commented):
rest LGTM

gozssky (commented):
LGTM after we have a consensus on https://github.com/pingcap/br/pull/1336#discussion_r670266022

pkg/lightning/restore/restore.go
glorv: Not sure. But since we don't GA this feature and if lightning exited before finished, the current logic may still not be recover except manually drop the meta schema. We also don't recommend change lightning binary during one import task.

gozssky:
Unstable configlist test will be fixed by https://github.com/pingcap/br/pull/1393.

gozssky (approved):

PR #1343, opened 2021-07-13
pkg/membuf: refine bytesBuf to membuf

gozssky:

There are two almost identical implementations of `bytesBuffer`, which are used by Lightning and BR. Merge them into one for better code readability. 

In addition, this PR sets the capacity for each buffer to avoid accidentally modifying the buffer while using `append` to modify another buffer.

Merge lightning/bytesBuffer to utils.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### Release note

No release note.


pkg/manual/manual.go
```diff
@@ -4,11 +4,62 @@
 
 package manual
```
YuJuncen, line 5: I'm afraid that `pkg/manual` with cgo might make TiDB depend on gcc and the cgo part of this file was moved once at https://github.com/pingcap/br/pull/525.

Also cc @glorv, is remove cgo still acceptable for now? Or we may need to move the cgo part to another package?

gozssky: It seems we don't need `manual`  package anymore. I have removed the entire `manual`  package. @glorv PTAL

pkg/lightning/backend/kv/session.go
```diff
@@ -61,19 +60,18 @@ func (b *bytesBuf) add(v []byte) []byte {
...
 	copy(b.buf[start:], v)
 	b.idx += len(v)
-	return b.buf[start:b.idx]
+	return b.buf[start:b.idx:b.idx]
 }
 
 func newBytesBuf(size int) *bytesBuf {
 	return &bytesBuf{
-		buf: manual.New(size),
+		buf: make([]byte, size),
```
glorv, line 68: CGO is used here to better manage memory, could you do a benchmark to make sure replace CGO with go slice won't increase the peek memory usage

gozssky: CGO is not actually used here because the `manual` dependency is `br/pkg/manual` 😂.

glorv: But the bytesBuffer in pkg/lightning/backend/local depends on pkg/lightning/manual with uses CGO, so I think we'd better do some benchmark or just keep the manual implement.

gozssky: okay, I will do a benchmark later

gozssky: After benchmark lightning local backend without `manual.go`, I found the peak memory increased by about 30%. So I will consider adding `manual.go` back. 

However, BR doesn't need CGO. It's hard to keep a same implementation  of  `byteBuffer`, since there is only one `bytesRecycleChan`. But for this work to continue, I think we can introduce an `Allocator` interface to allocate and free memory. The code looks like:
```go
type Allocator interface {
	Alloc(n int) []byte
	Free([]byte)
}

type stdAllocator struct{}

func (stdAllocator) Alloc(n int) []byte {
	return make([]byte, n)
}

func (stdAllocator) Free(b []byte) {}

type Pool struct {
	ch        chan []byte
	allocator Allocator
}

func NewPool(size int, allocator Allocator) *Pool {
	return &Pool{
		ch:        make(chan []byte, size),
		allocator: allocator,
	}
}

func (p *Pool) Acquire() []byte {
	select {
	case b := <-p.ch:
		return b
	default:
		return p.allocator.Alloc(1 << 20) // 1M
	}
}

func (p *Pool) Release(b []byte) {
	select {
	case p.ch <- b:
	default:
		p.allocator.Free(b)
	}
}

func (p *Pool) NewBuffer() *Buffer {
	return &Buffer{pool: p, bufs: make([][]byte, 0, 128), curBufIdx: -1}
}

var globalPool = NewPool(1024, stdAllocator{})

var NewBuffer = globalPool.NewBuffer
```

What do you think about it? @glorv

glorv: LGTM

gozssky:
@glorv @YuJuncen PTAL agnin, thanks

pkg/lightning/backend/kv/session.go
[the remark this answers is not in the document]
gozssky: The previous `manual` is `"github.com/pingcap/br/pkg/manual"` , which is equivalent to `make`.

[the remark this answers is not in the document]
glorv: I think CGO is a better choice, the previous usage of br's manual should be a mistake.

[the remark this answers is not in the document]
gozssky: Well, I updated the import to `"github.com/pingcap/br/pkg/lightning/manual"`.

glorv (approved):

YuJuncen (approved):

gozssky:
@YuJuncen Can you help merge this PR

PR #1363, opened 2021-07-19
restore: concurrently split

YuJuncen:
Before, when restoring many small tables, the batcher would probably send small batch due to the so called `AutoCommit` feature of the batcher. By this, we can make the split & scatter & restore worker more active.

But frequently send small batches isn't free. The split step is costly and I/O bounded for even small batches. For example, it costs about 3s to splitting 60 ranges, but restore those ranges typically costs only 1s. Then the restore worker get idle at most time. The restore hence has slow down.

BR pipeline-2

Instead of using a single split worker, this PR allow multi restore batches be split concurrently.
We added two **hidden** flags, `--batch-flush-interval` and `--pd-concurrency`, the former for better tuning the behavior of batcher, the latter for tweaking the concurrent split.

### Check List 

Tests

### DIFF
--- a/pkg/restore/pipeline_items.go
+++ b/pkg/restore/pipeline_items.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/pingcap/br/pkg/metautil"
+	"github.com/pingcap/br/pkg/utils"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
@@ -184,6 +185,8 @@ type tikvSender struct {
 	inCh chan<- DrainResult
 
 	wg *sync.WaitGroup
+
+	tableWaiters *sync.Map
 }
 
 func (b *tikvSender) PutSink(sink TableSink) {
@@ -193,6 +196,7 @@ func (b *tikvSender) PutSink(sink TableSink) {
 }
 
 func (b *tikvSender) RestoreBatch(ranges DrainResult) {
+	log.Info("restore batch: waiting ranges", zap.Int("range", len(b.inCh)))
 	b.inCh <- ranges
 }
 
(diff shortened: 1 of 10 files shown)


pkg/restore/pipeline_items.go
```diff
@@ -232,44 +256,84 @@ func (b *tikvSender) splitWorker(ctx context.Context, ranges <-chan DrainResult,
...
+			files := r.result.Files()
+			// There has been a worker in the `RestoreFiles` procedure.
+			// Spawning a raw goroutine won't make too many requests to TiKV.
+			go b.client.RestoreFilesAndThen(ctx, files, r.result.RewriteRules, b.updateCh, func(e error) {
+				if e != nil {
+					b.sink.EmitError(e)
+					return
+				}
+				log.Info("restore batch done", rtree.ZapRanges(r.result.Ranges))
+				r.done()
```
3pointer, line 332: Why need `r.done()` 🤔

3pointer: can we use result in channel to sync the all procedure.

pkg/restore/pipeline_items.go
```diff
@@ -232,44 +256,103 @@ func (b *tikvSender) splitWorker(ctx context.Context, ranges <-chan DrainResult,
...
-				return
-			}
-
-			log.Info("restore batch done", rtree.ZapRanges(result.Ranges))
-			b.sink.EmitTables(result.BlankTablesAfterSend...)
+			restoreWorks.Add(1)
+			files := r.result.Files()
+			// There has been a worker in the `RestoreFiles` procedure.
+			// Spawning a raw goroutine won't make too many requests to TiKV.
+			go b.client.RestoreFilesAndThen(ctx, files, r.result.RewriteRules, b.updateCh, func(e error) {
```
Little-Wallace, line 345: How about use `errgroup.WithContext`

pkg/restore/util.go
```diff
@@ -353,28 +353,44 @@ func truncateTS(key []byte) []byte {
...
+// SplitRanges is the synchronous version of SplitRangesAndThen.
+func SplitRanges(
+	ctx context.Context,
+	client *Client,
+	ranges []rtree.Range,
+	rewriteRules *RewriteRules,
+	updateCh glue.Progress,
+) error {
+	ch := make(chan error)
+	go SplitRangesAndThen(ctx, client, ranges, rewriteRules, updateCh, func(err error) {
```
Little-Wallace, line 390: Why not 
```
var ret error
SplitRangesAndThen(ctx, client, ranges, rewriteRules, updateCh, func(err error) {
     ret = err
})
return ret
```

pkg/restore/pipeline_items.go
```diff
@@ -232,44 +256,103 @@ func (b *tikvSender) splitWorker(ctx context.Context, ranges <-chan DrainResult,
...
+// till all tables provided are no more ‘current restoring’.
+func (b *tikvSender) waitTablesDone(ts []CreatedTable) {
+	for _, t := range ts {
+		wg, ok := b.tableWaiters.LoadAndDelete(t.Table.ID)
+		if !ok {
+			log.Panic("bug! table done before register!",
+				zap.Any("wait-table-map", b.tableWaiters),
+				zap.Stringer("table", t.Table.Name))
 		}
+		wg.(*sync.WaitGroup).Wait()
```
Little-Wallace, line 321: Why not use a single global `WaitGroup`?

YuJuncen: We make a `waitGroup` for each tables, so once some table has been restored fully (i.e. the wait group for it was done), we can send it to checksumming immediately.

pkg/restore/util.go
YuJuncen: Even I don't know why, this ruins the whole procedure.

Thankfully, after using `errgroup`, we don't need this change anymore. 😄

PR #1370, opened 2021-07-20
Lightning: support batch import downgrading

### LINKED ISSUE #1366: Lightning: support downgrading to row-by-row insert when batch insert meets an error

Part of #1365. Usually, Lightning will insert multiple rows at once, however, this will make the error recording and skipping not that easy if it's in a batch mode. We need to support downgrading to row-by-row insert when batch insert meets an error.

Normally, Lightning will import multiple rows like this:

```sql
start transaction;
insert into t1 values (111), (222), (333), (444);
commit;
```

If we want to record and skip any row error that occurs during this insert while not interrupt other normal data rows, we need to make the insert look like this:

```sql
start transaction;
insert into t1 values (111);
insert into t1 values (222);
insert into t1 values (333);
insert into t1 values (444);
commit;
```

Though this may lead to performance degradation, we will only downgrade to row-by-row insert when batch insert meets an error.

### Implementation

https://github.com/pingcap/br/blob/a6f471e8e1d1db0fe4c07f96ff25a42eadd4558f/pkg/lightning/backend/tidb/tidb.go#L366-L384

`(*tidbBackend).WriteRows` will split data into different rows and check if the error is retryable. Retryable errors are often the result of, e.g, network problems, in which case retrying is feasible. However, errors that we need to record and skip are often caused by some fundamental errors, such as the mismatched column type, for which we only need to process further by implementing new code here.

* For the recording, we need to know the position and content of the data from the files for import.
* For the skipping, we need to only skip the row with the error and make sure the others being insert successfully.

Furthermore, metrics and tracking information are needed, such as error counts.

JmPotato:

Close #1366. Support batch import to downgrade to row-by-row mode.

Support downgrading to row-by-row mode once batch inserting meets the error.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

 - No release note.

### DIFF
--- a/pkg/lightning/backend/tidb/tidb.go
+++ b/pkg/lightning/backend/tidb/tidb.go
@@ -86,8 +86,8 @@ type tidbBackend struct {
 	// as for the errors occur before the inserting, e.g, parsing CSV files, KV conversion,
 	// are not being considered yet.
 	// TODO: clarify the scope of the error.
-	maxErrorCount *atomic.Uint64
-	curErrorCount *atomic.Uint64
+	maxErrorCount int /* read only */
+	curErrorCount atomic.Uint64
 }
 
 // NewTiDBBackend creates a new TiDB backend using the given database.
@@ -104,8 +104,7 @@ func NewTiDBBackend(db *sql.DB, onDuplicate string, maxErrorCount int) backend.B
 	return backend.MakeBackend(&tidbBackend{
 		db:            db,
 		onDuplicate:   onDuplicate,
-		maxErrorCount: atomic.NewUint64(uint64(maxErrorCount)),
-		curErrorCount: atomic.NewUint64(0),
+		maxErrorCount: maxErrorCount,
 	})
 }
 
@@ -407,7 +406,7 @@ rowLoop:
 					continue retryLoop
 				}
 				// If batch is false and the error is not nil, it means we reach the max error count in the non-batch mode.
-				return errors.Annotatef(err, "[%s] write rows reach max error count %d", tableName, be.maxErrorCount.Load())
+				return errors.Annotatef(err, "[%s] write rows reach max error count %d", tableName, be.maxErrorCount)
 			}
 		}
 	}
@@ -415,23 +414,17 @@ rowLoop:
 }
 
 type stmtTask struct {
-	rowIdxStart int
-	rowIdxEnd   int
-	rows        tidbRows
-	stmt        string
+	rows tidbRows
+	stmt string
 }
 
 // If batch is true, we will insert multiple rows like this:
-//   start transaction;
 //   insert into t1 values (111), (222), (333), (444);
-//   commit;
 // If batch is false, we will insert multiple rows like this:
-//   start transaction;
 //   insert into t1 values (111);
 //   insert into t1 values (222);
 //   insert into t1 values (333);
 //   insert into t1 values (444);
-//   commit;
 // See more details in br#1366: https://github.com/pingcap/br/issues/1366
 // WriteRowsToDB will return the index of error row and its corresponding error if any error occurs.
 func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, columnNames []string, r kv.Rows, batch bool) error {
@@ -452,14 +445,14 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
 			}
 			insertStmt.WriteString(string(row))
 		}
-		stmtTasks[0] = stmtTask{0, len(rows), rows, insertStmt.String()}
+		stmtTasks[0] = stmtTask{rows, insertStmt.String()}
 	} else {
 		stmtTasks = make([]stmtTask, 0, len(rows))
-		for i, row := range rows {
+		for _, row := range rows {
 			var finalInsertStmt strings.Builder
 			finalInsertStmt.WriteString(insertStmt.String())
 			finalInsertStmt.WriteString(string(row))
-			stmtTasks = append(stmtTasks, stmtTask{i, i, tidbRows{row}, finalInsertStmt.String()})
+			stmtTasks = append(stmtTasks, stmtTask{[]tidbRow{row}, finalInsertStmt.String()})
 		}
 	}
 	return be.execStmts(ctx, stmtTasks, batch)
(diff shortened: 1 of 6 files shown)


JmPotato:
@kennytm a friendly ping😁

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -411,27 +492,50 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
...
-	// complication arise from data length overflow of BIT and BINARY columns
+// Only used for TestWriteRowsErrorSkip
+var mockErrorCount = 0
 
-	for i, row := range rows {
-		if i != 0 {
-			insertStmt.WriteByte(',')
+func (be *tidbBackend) execStmts(ctx context.Context, stmtTasks []stmtTask, rows tidbRows) (int, error) {
+	// Retry will be done externally, so we're not going to retry here.
+	tx, err := be.db.BeginTx(ctx, &sql.TxOptions{})
```
glorv, line 503: Since some if the stmtTask can fail, I think there is no good to exec all the statements in one tx.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -411,27 +492,50 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
...
+			if mockErrorCount == 0 || mockErrorCount == 2 || mockErrorCount == 5 {
+				err = stderrors.New("mock non-retryable error")
+			}
+			mockErrorCount++
+		})
+
+		if err != nil {
+			if !common.IsContextCanceledError(err) {
+				log.L().Error("execute statement failed inside the transaction", zap.String("stmt", redact.String(stmt)),
+					zap.Array("rows", rows), zap.Error(err))
```
Little-Wallace, line 524: Do we need to record every row even if only one statement task failed? When exec without batch, every statement will only include one row.

glorv: BTW, we need to redact the `rows` since it's sensitive data

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -411,27 +492,50 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
...
+			}
+			if rollbackErr := tx.Rollback(); rollbackErr != nil {
+				panic(fmt.Sprintf("execute statement failed: %v, unable to rollback: %v\n", err, rollbackErr))
+			}
+			return stmtTask.rowIdx, errors.Trace(err)
```
Little-Wallace, line 529: If we return an error, the error logs in this method will be record many times. I think we shall collect the fail rows, and record them in error log in another place.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -364,31 +378,99 @@ func (be *tidbBackend) ImportEngine(context.Context, uuid.UUID) error {
...
+				}
+				// If batch is false, it means we find the row where the non-retryable error located,
+				// now we should record and skip it if necessary.
+				be.curErrorCount++
+				if be.curErrorCount >= be.maxErrorCount {
+					return errors.Annotatef(err, "[%s] write rows reach max error count %d", tableName, be.maxErrorCount)
+				}
+				// TODO: record the error.
+				// Reset the retry count before we skip the error and continue inserting the rest of rows.
+				i = 0
```
Little-Wallace, line 409: Does it mean that we will retry maxErrorCount times?

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -364,31 +378,99 @@ func (be *tidbBackend) ImportEngine(context.Context, uuid.UUID) error {
...
-	var err error
-outside:
+rowLoop:
 	for _, r := range rows.SplitIntoChunks(be.MaxChunkSize()) {
+		batch := true
+		rowsToSkip := make(map[int]struct{})
+	retryLoop:
 		for i := 0; i < writeRowsMaxRetryTimes; i++ {
-			err = be.WriteRowsToDB(ctx, tableName, columnNames, r)
+			errRowIdx, err := be.WriteRowsToDB(ctx, tableName, columnNames, r, batch, rowsToSkip)
```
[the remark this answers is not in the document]
JmPotato: Please check the `TestWriteRowsErrorSkip` test case in this PR, I think it can help understand the current error count and retry strategy. Maybe it's not quite like what you think here.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -364,31 +378,99 @@ func (be *tidbBackend) ImportEngine(context.Context, uuid.UUID) error {
...
 	if len(rows) == 0 {
-		return nil
+		return 0, nil
 	}
 
+	insertStmt := be.prepareStmt(tableName, columnNames)
+	// Note: we are not going to do interpolation (prepared statements) to avoid
+	// complication arise from data length overflow of BIT and BINARY columns
+	var stmtTasks []stmtTask
+	if batch {
```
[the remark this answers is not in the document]
JmPotato: I think the bool is still needed here because the rows we pass are the same and we can only tell whether it's a batch mode by this flag according to the current implementation.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -78,20 +80,33 @@ type tidbEncoder struct {
 type tidbBackend struct {
 	db          *sql.DB
 	onDuplicate string
+
+	// Max tolerable error count, once it's exceeded, stop the import.
+	// Currently, this will only count the error happens inside the TiDB Backend,
+	// as for the errors occur before the inserting, e.g, parsing CSV files, KV conversion,
+	// are not being considered yet.
+	// TODO: clarify the scope of the error.
+	maxErrorCount *atomic.Uint64
```
glorv, line 89: maxErrorCount is read-only, so it doesn't need to be automic. I must made a mistake in the last review comment.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -78,20 +80,33 @@ type tidbEncoder struct {
...
+	curErrorCount *atomic.Uint64
```
glorv, line 90:
```suggestion
	curErrorCount atomic.Uint64
```
No need to introduce another level of pointer

pkg/lightning/backend/tidb/tidb.go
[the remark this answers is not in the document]
glorv: If we convert the rows to tidbRows in the caller, we can also take slice of the rows to avoid introduce startIndex and endIndex

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -411,27 +479,81 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
...
+					log.L().Error("begin a transaction failed before executing statement",
+						zap.Int("row_index_start", stmtTask.rowIdxStart), zap.Int("row_index_end", stmtTask.rowIdxEnd),
+						zap.Array("rows", stmtTask.rows), zap.String("stmt", redact.String(stmt)), zap.Error(err))
+				}
+				goto errCheck
+			}
+			_, err = tx.ExecContext(txnCtx, stmt)
+
+			failpoint.Inject("mockNonRetryableError", func() {
+				if batch || stmtTask.rowIdxStart == 1 || stmtTask.rowIdxStart == 3 {
```
glorv, line 508: Shall we depend on gomock to return this kind of error instead of depending on this kind of some what weird ways?

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -411,27 +479,81 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
...
-	}
+func (be *tidbBackend) execStmts(ctx context.Context, stmtTasks []stmtTask, batch bool) error {
+stmtTaskLoop:
+	for _, stmtTask := range stmtTasks {
+	retryLoop:
+		for i := 0; i < writeRowsMaxRetryTimes; i++ {
+			stmt := stmtTask.stmt
+			txnCtx, cancel := context.WithCancel(ctx)
+			defer cancel()
+			tx, err := be.db.BeginTx(txnCtx, &sql.TxOptions{})
```
glorv, line 496: I still not understand why we need to depend on a transaction to exec these statements? If batch is true, then there should be only one statement, so tx is useless. If there are multi statements and some of them are expect to fail, then seems exec the statement one by one until meeting the first error seems to be a better choice, in this way wo can also avoid retry all the alredy succeeded statements again.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -364,31 +378,79 @@ func (be *tidbBackend) ImportEngine(context.Context, uuid.UUID) error {
...
+				if batch {
+					i = 0
+					batch = false
+					continue retryLoop
+				}
+				// If batch is false and the error is not nil, it means we reach the max error count in the non-batch mode.
+				return errors.Annotatef(err, "[%s] write rows reach max error count %d", tableName, be.maxErrorCount)
 			}
 		}
-		return errors.Annotatef(err, "[%s] write rows reach max retry %d and still failed", tableName, writeRowsMaxRetryTimes)
```
glorv, line 381: Need to keep this line if WriteRowsToDB return retryable error more than writeRowsMaxRetryTimes times

pkg/lightning/backend/tidb/tidb.go
JmPotato: I haven't any better idea to do this yet, but some comments are added. PTAL.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -365,30 +381,80 @@ func (be *tidbBackend) ImportEngine(context.Context, uuid.UUID) error {
...
 	for _, r := range rows.SplitIntoChunks(be.MaxChunkSize()) {
+		batch := true
+	retryLoop:
 		for i := 0; i < writeRowsMaxRetryTimes; i++ {
-			err = be.WriteRowsToDB(ctx, tableName, columnNames, r)
+			err = be.WriteRowsToDB(ctx, tableName, columnNames, r, batch)
 			switch {
 			case err == nil:
-				continue outside
+				continue rowLoop
```
Little-Wallace, line 392: Why not use `break`

JmPotato: Using `break` here will trigger the line 403.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -411,27 +482,54 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
...
+				// Retry the non-batch insert here.
+				if common.IsRetryableError(err) {
+					continue
+				}
+				// TODO: record the error.
+				// Check the error count.
+				be.curErrorCount.Inc()
+				if be.curErrorCount.Load() >= uint64(maxErrorCount) {
+					return errors.Trace(err)
+				}
```
kennytm, line 523: i mean just remove the entire error counting logic (only keep the TODO comment). the max-error implementation will not use a localized error count.

JmPotato: Do you mean just keep the downgrading logic in this PR🤔?

kennytm: yes, just keep the PR focusing on downgrading.

JmPotato: Done. PTAL.

JmPotato:
@kennytm @glorv @Little-Wallace PTAL again, thx!

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -411,27 +469,45 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
...
+				if common.IsRetryableError(err) {
+					continue
+				}
+				// TODO: count, record and skip the error.
+				// Just return if any error occurs for now.
+				return errors.Trace(err)
+			}
+			// No error, contine the next stmtTask.
+			break
+		}
```
glorv, line 505: Need to return a error here if max retry times is exceeded.

glorv:
Does the `max-erorr-count` still needed?

pkg/lightning/backend/tidb/tidb.go
JmPotato: I add a check in the line 504 to make sure the logic won't reach here.

```golang
if common.IsRetryableError(err) && i != writeRowsMaxRetryTimes-1 {
```

JmPotato:
> Does the `max-erorr-count` still needed? Don't see it in this PR

Please take a look at the @kennytm's comments, he thinks we should focus this PR on the logic of downgrading.

Little-Wallace (approved):
LGTM

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -373,30 +374,88 @@ func (be *tidbBackend) ImportEngine(context.Context, uuid.UUID) error {
...
+func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, columnNames []string, r kv.Rows) error {
+	rows := r.(tidbRows)
+	insertStmt := be.checkAndBuildStmt(rows, tableName, columnNames)
+	if insertStmt == nil {
+		return nil
+	}
+	stmtTasks := make([]stmtTask, 0, len(rows))
+	for _, row := range rows {
+		var finalInsertStmt strings.Builder
+		finalInsertStmt.WriteString(insertStmt.String())
```
kennytm, line 451: move the `insertStmt.String()` call outside of the loop

JmPotato:
Migrate to https://github.com/pingcap/tidb/pull/27008. cc @kennytm @Little-Wallace @glorv

Issue #1233, opened 2021-06-16
Lightning import always report "the number of cluster read_flows more than expected"


XuJianxu:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
Use lightning to import the data, and do not skip the pre-check
tidb-lightning -config=lightning.toml

2. What did you expect to see?
If no heavy workload in tidb, all pre-check will pass.

3. What did you see instead?
Performance pre-check fail even the tidb cluster was new deployed.
![image](https://user-images.githubusercontent.com/23207893/122215126-74f15780-cedd-11eb-87ec-fc5687faed0e.png)

4. What version of BR and TiDB/TiKV/PD are you using?

tidb-lightning -V5.1.0-20210615
tidb-server -V5.1.0-20210615
tikv-server -V5.1.0-20210615
pd-server -V5.1.0-20210615

5. Operation logs
   - Please upload `br.log` for BR if possible
   - Please upload `tidb-lightning.log` for TiDB-Lightning if possible
   - Please upload `tikv-importer.log` from TiKV-Importer if possible
   - Other interesting logs

6. Configuration of the cluster and the task
   - `tidb-lightning.toml` for TiDB-Lightning if possible
   - `tikv-importer.toml` for TiKV-Importer if possible
   - `topology.yml` if deployed by TiUP

7. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus if possible

glorv:
closed by #1247

cyliu0:
![Lark20210714-173200](https://user-images.githubusercontent.com/35356271/125599598-5343b73e-6831-46a5-8bc4-1d88811ac1fb.png)
still got this precheck failed. @3pointer

cyliu0:
The above is a similar one, write flow, not exactly the same.

3pointer:
@cyliu0 does the cluster has other workload when import starts.

cyliu0:
> @cyliu0 does the cluster has other workload when import starts.

No workload. Before the import, I dropped the same data which is not very large, only 10 warehouses tpcc data. This is the only operation that might lead to this.

3pointer:
Ok, then I think the writeflow and readflow is not a proper check item. I think we may find another way to detect workload.

fengou1:
Still study the lightning code, need more time to understand read flow logic.

Leavrth:
[...]

PR #1373, opened 2021-07-20
skip/remove check cluster has workload (isOnline)

fengou1:
Resolve issue 1233:  https://github.com/pingcap/br/issues/1233

skip/remove check cluster has workload (isOnline) 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note

 -


pkg/lightning/restore/restore_test.go
```diff
@@ -1550,67 +1550,67 @@ func (s *restoreSchemaSuite) TestRestoreSchemaContextCancel(c *C) {
...
-		c.Assert(err, IsNil)
-
-		c.Assert(template.FailedCount(Warn), Equals, ca.expectWarnCount)
-		c.Assert(template.FailedCount(Critical), Equals, ca.expectErrorCount)
-		c.Assert(template.Success(), Equals, ca.expectResult)
-		c.Assert(strings.ReplaceAll(template.Output(), "\n", ""), Matches, ca.expectMsg)
-
-		server.Close()
-	}
+	// cases := []struct {
```
glorv, line 1553: I prefer to lower case this function name and add a common to minimal the affacted lines

pkg/lightning/restore/check_info.go
```diff
@@ -84,23 +83,25 @@ func (rc *Controller) ClusterIsOnline(ctx context.Context) error {
...
-			passed = false
-			regionStr, err := json.Marshal(region)
-			if err != nil {
-				return errors.Trace(err)
-			}
-			message = fmt.Sprintf("The write flow on cluster are more than expection, %s", string(regionStr))
-			return nil
-		}
-	}
+	// disabled the following code since it blocked some QA tests, it surpose to investigate a
```
glorv, line 86: How about skip this function where it is called

pkg/lightning/restore/restore_test.go
fengou1: directly remove this function call,  When multiple lightning instance work on the cluster, check cluster workload is meaningless

pkg/lightning/restore/check_info.go
fengou1: skip the function by remove it, since multiple lightning instance work on the cluster, check cluster workload is meaningless

fengou1:
Currently the CheckClusterIsOnline is to check if the cluster has business workload. We remove this check because:
1. multiple lightning be introduced, there is no way to check if cluster has other workload.
2. when lightning start to work, most of time, we assume the cluster is stay idle and clean.
3. Currently, there is no interface in pd or TiKV to check the workload.

Hot Region is also check the written or read bytes/keys during period of time, it can not resolve this issue.

glorv (approved):

gozssky (approved):

3pointer (approved):

PR #1377, opened 2021-07-21
restore: less split during restoring 

### LINKED ISSUE #1374: Too many empty region after restoration of many small tables
1. What did you do?
Restore a workload with 190G, 6000 tables, according to the backup meta, there are about 42K files backed up.

Restore done, and the region count should be less than 42K (or we nearly must get empty regions), and the number of empty region should be a reasonable value.

3. What did you see instead?
See figures below.

Figure 1. too many unhealthy regions
image
Figure 2. too many regions
image

Nightly (2020-7-20)

5. Operation logs

(The log is too large to be updated... Too many write conflicts with crating tables concurrently...)

YuJuncen:
Partially fix #1374

```
before：
--|-------t1 data-------|-----|---t2 data-------|
after: 
----------t1 data-------|---------t2 data-------|

Legends:
'|' the split point
'-' the key space
```

Also, `downloadSST` would find rewrite rules by **file** instead of **region start key** for now.

### Check List 

Tests 

 The same case as #1374, the region count reduced about 1/2.
image

### Release note

 - BR would split less regions to reduce the number of empty region after restoration.


3pointer:
LGTM, but I think we need more tests on this fix.  @fubinzh

YuJuncen:
A Note: This optimization assumes all files only contain one table or one index. If this precondition violated, a data lost may happen. 

When we have data keys and index keys in the same file(for example, `t1_i1_d` and `t1_r1`): in the past, we should split at the `[t1_i1, t1_r]` and import this file twice both at `[t1_i1, t1_r)` and `[t1_r, ...)`, which sounds inefficient but did the right thing. With this change, things get wrong: only the part of `[t1_i1, t1_r)` could be imported, and the `t1_r1` part wouldn't be rewritten, or even worse, an unordered SST file would be ingested.

This defect wouldn't affect the current BR because we create backup files for each indices and data separately. But we did need to change the way we validate files.

YuJuncen:
close it due to https://github.com/pingcap/tidb/pull/27240

PR #1387, opened 2021-07-27
lightning: auto configure for parallel import

Little-Wallace:
Users have encountered many problems in the process of using Lightning to perform parallel import. Many configurations are too complicated and difficult to understand, which increases the difficulty of users.

This PR will adjust some configure by sample the source data.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note

 - No releasee note


pkg/lightning/backend/local/local.go
```diff
@@ -29,6 +29,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/pingcap/br/pkg/pdutil"
```
kennytm, line 32: move this to line 64

pkg/lightning/backend/local/local.go
```diff
@@ -1381,6 +1383,27 @@ func (local *local) WriteToTiKV(
...
+			store, err := local.pdCtl.GetStoreInfo(ctx, peer.StoreId)
+			if err != nil {
+				e = err
+				continue
+			}
+			if store.Status.Capacity > 0 {
+				// The available disk percent of TiKV
+				ratio := store.Status.Available * 100 / store.Status.Capacity
+				if ratio < 10 {
+					return nil, Range{}, rangeStats{}, errors.Errorf("The available disk of TiKV only left %d, and capacity is %d", store.Status.Available, store.Status.Capacity)
```
kennytm, line 1398: 1. include the TiKV address or store ID or any identifying information into the error message
2. include the unit. i guess it is "bytes" in both cases
3. i think measuring the absolute amount of available space is better than a percentage. if the TiKV disk is very small, 10% avail is probably not enough; and if the TiKV is very large, 10% avail would be excessive.

pkg/lightning/checkpoints/checkpoints.go
```diff
@@ -595,6 +597,10 @@ func (*NullCheckpointsDB) Get(_ context.Context, _ string) (*TableCheckpoint, er
 	}, nil
 }
 
+func (cpdb *NullCheckpointsDB) IsEmpty(_ context.Context) (bool, error) {
+	return false, nil
```
kennytm, line 601:
```suggestion
	return true, nil
```

😅 right?

pkg/lightning/checkpoints/checkpoints.go
```diff
@@ -717,6 +723,19 @@ func (cpdb *MySQLCheckpointsDB) Close() error {
 	return errors.Trace(cpdb.db.Close())
 }
 
+func (cpdb *MySQLCheckpointsDB) IsEmpty(ctx context.Context) (bool, error) {
+	s := common.SQLWithRetry{
+		DB:     cpdb.db,
+		Logger: log.With(zap.String("CheckEmpty", "IsEmpty")),
```
kennytm, line 729: ??

pkg/lightning/config/config.go
```diff
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"time"
 
+	pd "github.com/tikv/pd/client"
+
```
kennytm, line 30:
```suggestion
```

pkg/lightning/config/config.go
```diff
@@ -152,7 +158,6 @@ func (cfg *Config) ToTLS() (*common.TLS, error) {
 
 type Lightning struct {
 	TableConcurrency  int    `toml:"table-concurrency" json:"table-concurrency"`
-	IndexConcurrency  int    `toml:"index-concurrency" json:"index-concurrency"`
```
kennytm, line 155: removing this breaks existing `tidb-lightning.toml` which defined `index-concurrency` (triggers "config file contained unknown configuration options" error).

pkg/lightning/restore/check_info.go
```diff
@@ -14,20 +14,31 @@
...
 
+	"github.com/pingcap/tidb/table/tables"
+
+	"github.com/pingcap/br/pkg/lightning/backend/kv"
+	"github.com/pingcap/br/pkg/lightning/log"
+	verify "github.com/pingcap/br/pkg/lightning/verification"
+
+	"github.com/pingcap/failpoint"
+
+	"github.com/pingcap/br/pkg/lightning/mydump"
```
kennytm, line 33: why can't you sort the import into 3 groups like all existing files?

pkg/lightning/restore/check_info.go
```diff
@@ -82,27 +95,17 @@ func (rc *Controller) ClusterResource(ctx context.Context) error {
 	}
 	totalAvailable := typeutil.ByteSize(0)
 	for _, store := range result.Stores {
-		totalAvailable += store.Status.Available
+		totalAvailable += store.Status.Capacity
```
kennytm, line 98: better rename `totalAvailable` to `totalCapacity`? the message in line 111 may be inacurrate too.

pkg/lightning/restore/check_info.go
```diff
@@ -238,7 +259,7 @@ func (rc *Controller) CheckpointIsValid(ctx context.Context, tableInfo *md.MDTab
 	tableCheckPoint, err := rc.checkpointsDB.Get(ctx, uniqueName)
 	if err != nil {
 		// there is no checkpoint
-		log.Debug("no checkpoint detected", zap.String("table", uniqueName))
+		log.L().Debug("no checkpoint detected", zap.String("table", uniqueName))
```
kennytm, line 262: ???

pkg/lightning/restore/check_info.go
Little-Wallace: In acutally, I want to check the available rather than capacity. But when one lightning instance call this function,there may already import some data to TiKV cluster.

pkg/lightning/restore/check_info.go
Little-Wallace: The origin code use another `log` dependency ,which is usually used by `BR` rather than lightning

pkg/lightning/restore/meta_manager.go
```diff
@@ -520,17 +524,81 @@ type storedCfgs struct {
...
+			return errors.Annotate(err, "fetch task meta failed")
+		}
+		var taskID int64
+		for rows.Next() {
+			if err = rows.Scan(&taskID); err != nil {
+				rows.Close()
+				return errors.Trace(err)
+			}
+			if taskID == m.taskID {
+				hasStarted = true
```
gozssky, line 558: I'm not sure if `Transact` will retry the closure. If a retry happens, we actually need to reset `hasStarted` at first.

pkg/lightning/restore/meta_manager.go
```diff
@@ -520,17 +524,81 @@ type storedCfgs struct {
...
+		var (
+			taskID      int64
+			sourceBytes int64
+		)
+		for rows.Next() {
+			if err = rows.Scan(&taskID, &sourceBytes); err != nil {
+				rows.Close()
+				return errors.Trace(err)
+			}
+			source += sourceBytes
```
gozssky, line 594: ditto

pkg/lightning/restore/table_restore.go
```diff
@@ -158,28 +159,36 @@ func (t *TableRestore) RebaseChunkRowIDs(cp *checkpoints.TableCheckpoint, rowIDB
 //
 // The argument `columns` _must_ be in lower case.
 func (tr *TableRestore) initializeColumns(columns []string, ccp *checkpoints.ChunkCheckpoint) error {
+	colPerm, err := createColumnPermutation(columns, tr.ignoreColumns, tr.tableInfo.Core)
+	if err != nil {
+		return err
+	} else {
+		ccp.ColumnPermutation = colPerm
+	}
```
gozssky, line 167:
```suggestion
	}
	ccp.ColumnPermutation = colPerm
```

pkg/lightning/checkpoints/checkpoints.go
Little-Wallace: This method has been removed

pkg/lightning/restore/check_info.go
```diff
@@ -439,3 +469,126 @@ func (rc *Controller) SchemaIsValid(ctx context.Context, tableInfo *md.MDTableMe
...
+		}
+	}
+
+	if rowSize > 0 && kvSize > rowSize {
+		tableMeta.IndexRatio = float64(kvSize) / float64(rowSize)
+	} else {
+		tableMeta.IndexRatio = 1.0
+	}
+	log.L().Info("Sample source data", zap.String("table", tableMeta.Name), zap.Float64("IndexRatio", tableMeta.IndexRatio), zap.Bool("IsSourceOrder", tableMeta.IsRowOrdered))
+	parser.Close()
```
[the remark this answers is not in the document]
Little-Wallace: OK

pkg/lightning/config/config.go
glorv: For the current logic, remove index-concurrency may bring a side-effect. Since we start import index engine only after all the data engine are imported. I'm afraid after this change, more table's index engine will finally start import at the same time than before, which may hurt the performance.

pkg/lightning/restore/check_info.go
gozssky: I wrote a script that sorts all imports into 3 groups. 
```bash
#!/bin/bash

# Remove blank lines of imports.
find . -type f -name "*.go"  | grep -v pb.go | xargs  sed -i ":a;N;\$!ba;s/\"\n\n\t\"/\"\n\t\"/g"
# Sort all imports into 3 groups.
make static # ignore the output errors
```
@kennytm Do you think it is necessary to check it in CI?

pkg/lightning/config/config.go
Little-Wallace: But most of them will be blocked by TableConcurrency

pkg/lightning/restore/check_info.go
kennytm: @gozssky yeah we could put it in `make static` (together with `gofumports`)

btw what happened to our Go style guide 🤣 https://github.com/pingcap/style-guide/tree/master/docs/go

pkg/lightning/restore/check_info.go
```diff
@@ -14,28 +14,35 @@
...
+	"bytes"
 	"context"
 	"fmt"
 	"io"
 	"path/filepath"
 	"reflect"
 	"strings"
 
+	"github.com/pingcap/br/pkg/lightning/config"
+
```
gozssky, line 26:
```suggestion

```

pkg/lightning/backend/local/local.go
Little-Wallace: Done

pkg/lightning/config/config.go
Little-Wallace: I have revert it.

gozssky:
@Little-Wallace You can use *`bash`* to run the follow command to sort all imports to 3 groups
```bash
git diff origin/master --name-only | grep -E "*.go" | xargs  sed -i ":a;N;\$!ba;s/\"\n\n\t\"/\"\n\t\"/g" && make static &>/dev/null
```

pkg/lightning/config/config.go
```diff
@@ -67,9 +69,12 @@ const (
 	ErrorOnDup = "error"
 
 	defaultDistSQLScanConcurrency     = 15
+	DistSQLScanConcurrencyPerStore    = 4
```
kennytm, line 72: this can be private?

pkg/lightning/backend/local/local.go
kennytm: @Little-Wallace item 1 is still not fixed.

pkg/lightning/restore/restore.go
```diff
@@ -1798,26 +1760,58 @@ func (rc *Controller) isLocalBackend() bool {
 // 3. Lightning configuration
 // before restore tables start.
 func (rc *Controller) preCheckRequirements(ctx context.Context) error {
-	if !rc.cfg.App.CheckRequirements {
-		log.L().Info("skip pre check due to user requirement")
-		return nil
-	}
```
kennytm, line 1804: removing this means all the pre-check are now always run, but all failures are just hidden away. why is this changed 🤔

pkg/lightning/restore/restore.go
```diff
@@ -1798,26 +1760,58 @@ func (rc *Controller) isLocalBackend() bool {
...
 	if rc.isLocalBackend() {
-		if err := rc.LocalResource(ctx); err != nil {
+		source, err := rc.EstimateSourceData(ctx)
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		// disable some pd schedulers
+		pdController, err := pdutil.NewPdController(ctx, rc.cfg.TiDB.PdAddr,
+			rc.tls.TLSConfig(), rc.tls.ToPDSecurityOption())
```
kennytm, line 1783: wrong comment? 😅 no schedulers are disabled here, the comment should be kept after line 1206.

kennytm:
rest LGTM

pkg/lightning/restore/restore.go
Little-Wallace: Because some tests shall be  always checked.

pkg/lightning/config/config.go
Little-Wallace: Let me take a look. Maybe

pkg/lightning/restore/restore.go
Little-Wallace: Wrong comment, I will remove it.

kennytm (approved):

gozssky (approved):

PR #1392, opened 2021-07-27
Lightning: Changed data-source-dir to use StorageBackend to prevent leaking secret access key to JSON output

kennytm:

When using Lightning to load from S3, the non-environmental access key and secret key have to be provided through URL parameters in `mydumper.data-source-dir`. This is stored as a plain string.

Lightning also prints the config as JSON before starting a task for debugging. 

These two together means the access key and secret key will be leaked to the log.

The type of `mydumper.data-source-dir` is changed from a string to a custom type (`*backuppb.StorageBackend`) which scrubs all extra parameters when serializing as JSON.

The change introduces some additional effects needing refactoring:

* There was a check whether `mydumper.data-source-dir` exists before running. This overlaps with the `StoragePermissions` pre-check for S3. These, along with GCS, are combined into the existing  `CheckPermissions` check.
* The `SkipCheckPath` field for external storage is thus entirely ignored (it is already set to `true` on Dumpling and TiCDC so it is safe to ignore). *The BR flag `--skip-check-path` is unaffected.*
* (More in review comments below.)

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note

 - Lightning will stop printing the explicit S3 secret access key from the config file into the log. (It is still discouraged to put the secret access key in the config file.)

### DIFF
--- a/pkg/lightning/config/config.go
+++ b/pkg/lightning/config/config.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
-	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -248,7 +247,7 @@ type MydumperRuntime struct {
 	ReadBlockSize    ByteSize         `toml:"read-block-size" json:"read-block-size"`
 	BatchSize        ByteSize         `toml:"batch-size" json:"batch-size"`
 	BatchImportRatio float64          `toml:"batch-import-ratio" json:"batch-import-ratio"`
-	SourceDir        string           `toml:"data-source-dir" json:"data-source-dir"`
+	SourceDir        SourceDir        `toml:"data-source-dir" json:"data-source-dir"`
 	CharacterSet     string           `toml:"character-set" json:"character-set"`
 	CSV              CSVConfig        `toml:"csv" json:"csv"`
 	MaxRegionSize    ByteSize         `toml:"max-region-size" json:"max-region-size"`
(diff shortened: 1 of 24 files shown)


pkg/lightning/config/config.go
```diff
@@ -762,54 +765,6 @@ func (cfg *Config) CheckAndAdjustTiDBPort(ctx context.Context, mustHaveInternalC
 	return nil
 }
 
-func (cfg *Config) CheckAndAdjustFilePath() error {
```
kennytm, line 765: this check-and-adjust does several things:

1. if the data-source-dir is a path, we `common.IsDirExists` it. this is going to be replaced by the ListObjects permission check.
2. turn the path into a URL. this comes for free by `storage.ParseBackend`.
3. ban unsupported schemes. currently all schemes are supported (including GCS), so the check is useless now.

pkg/lightning/lightning.go
```diff
@@ -259,6 +259,15 @@ func (l *Lightning) run(taskCtx context.Context, taskCfg *config.Config, g glue.
 		}
 	}()
 
+	storagePermissions := []storage.Permission{storage.ListObjects, storage.GetObject}
+	if !taskCfg.App.CheckRequirements {
+		storagePermissions = nil
+	}
+	s, err := taskCfg.Mydumper.SourceDir.NewStorage(ctx, storagePermissions)
```
kennytm, line 266: this replaces the `StoragePermission` pre-check

pkg/lightning/restore/check_info.go
```diff
@@ -253,11 +220,7 @@ func (rc *Controller) CheckpointIsValid(ctx context.Context, tableInfo *md.MDTab
 			chunk := eng.Chunks[0]
 			permFromCheckpoint = chunk.ColumnPermutation
 			columns = chunk.Chunk.Columns
-			if filepath.Dir(chunk.FileMeta.Path) != rc.cfg.Mydumper.SourceDir {
-				message := fmt.Sprintf("chunk checkpoints path is not equal to config"+
-					"checkpoint is %s, config source dir is %s", chunk.FileMeta.Path, rc.cfg.Mydumper.SourceDir)
-				msgs = append(msgs, message)
-			}
```
kennytm, line 260: i don't know how this check can ever be valid, when SourceDir can be non-local

Issue #1382, opened 2021-07-26
confusing error info when the data source dir doesn't contain a valid table schema sql file


cyliu0:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
import without table schema SQL file

2. What did you expect to see?
An error about missing schema files

3. What did you see instead?
```
# /tidb-lightning "-sorted-kv-dir" "/tmp/sorted-kv-dir" "-d" "s3://nfs/jianxu/parquet/import?access-key=minioadmin&se
cret-access-key=minioadmin&endpoint=http%3a%2f%2fminio.pingcap.net%3a9000&force-path-style=true" "-pd-urls" "source-tid
b-pd.cy-brie-testbed-zw6rt.svc:2379" "-tidb-host" "source-tidb-tidb.cy-brie-testbed-zw6rt.svc" "-tidb-port" "4000" "-ti
db-user" "root" "-tidb-password" "" "-backend" "local"
/ # cat /tmp/lightning.log.2021-07-26T08.27.19Z
[2021/07/26 08:27:19.968 +00:00] [INFO] [info.go:40] ["Welcome to TiDB-Lightning"] [release-version=v5.0.0-master] [git-hash=fb65c986afa7cd8470852662dd4785aa5aa09a9b] [git-branch=heads/refs/tags/master-nightly] [go-version=go1.16.4] [utc-build-time="2021-07-25 09:48:22"] [race-enabled=false]
[...]

PR #1396, opened 2021-07-29
lightning: print a clear error when schema file not found

gozssky:
 solve https://github.com/pingcap/br/issues/1382

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### Release note
No release note.

### DIFF
--- a/pkg/lightning/mydump/loader.go
+++ b/pkg/lightning/mydump/loader.go
@@ -54,6 +54,9 @@ type SourceFileMeta struct {
 }
 
 func (m *MDTableMeta) GetSchema(ctx context.Context, store storage.ExternalStorage) (string, error) {
+	if m.SchemaFile.FileMeta.Path == "" {
+		return "", errors.Errorf("schema file for `%s`.`%s` not found", m.DB, m.Name)
+	}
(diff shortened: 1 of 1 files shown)


dveeden:
Another similar fix is in #1397

pkg/lightning/mydump/loader.go
```diff
@@ -54,6 +54,9 @@ type SourceFileMeta struct {
 }
 
 func (m *MDTableMeta) GetSchema(ctx context.Context, store storage.ExternalStorage) (string, error) {
+	if m.SchemaFile.FileMeta.Path == "" {
+		return "", errors.Errorf("schema file for `%s`.`%s` not found", m.DB, m.Name)
```
dveeden, line 58: This is an improvement. However with `-no-schema` I don't expect it to read the schema file.

It looks like it wants to read the schema file if the table doesn't exist even with `-no-schema`. Here the problem is not that the schema file doesn't exist, the problem is that the table is missing.

gozssky: It sounds more reasonable. I'm going to close this PR and help merge your fix.

PR #1397, opened 2021-07-29
lightning: Return error for empty path

dveeden:
If `Path` is empty this leads to lightning trying to open the basedir,
which then results in an error like `Error: read /home/ubuntu/csv: is a
directory` which is misleading.

Issue: https://github.com/pingcap/br/issues/1394

### Check List 

Tests 

### Release note

 - Fix the error message when trying to load CSV data in tables that don't exist.

### DIFF
--- a/pkg/lightning/restore/restore.go
+++ b/pkg/lightning/restore/restore.go
@@ -513,6 +513,8 @@ func (worker *restoreSchemaWorker) makeJobs(
 				// we already has this table in TiDB.
 				// we should skip ddl job and let SchemaValid check.
 				continue
+			} else if tblMeta.SchemaFile.FileMeta.Path == "" {
+				return errors.Errorf("table `%s`.`%s` schema not found", dbMeta.Name, tblMeta.Name)
 			}
 			sql, err := tblMeta.GetSchema(worker.ctx, worker.store)
 			if sql != "" {
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 # for the web interface
 web/node_modules/
 web/dist/
+.vscode/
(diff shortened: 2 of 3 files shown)


dveeden:
1. Would be good to add a test for this, but I don't know what the best way is to do this.
2. The `if` could also consider `cfg.Mydumper.NoSchema` in various places.

gozssky:
Thanks for your PR, but it seems this is duplicate with https://github.com/pingcap/br/pull/1396.

gozssky:
Unstable integration test  [lightning_checkpoint](https://ci.pingcap.net/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/6255/pipeline/82).

pkg/lightning/restore/restore.go
```diff
@@ -513,6 +513,8 @@ func (worker *restoreSchemaWorker) makeJobs(
 				// we already has this table in TiDB.
 				// we should skip ddl job and let SchemaValid check.
 				continue
+			} else if tblMeta.SchemaFile.FileMeta.Path == "" {
+				return errors.Errorf("table '%s.%s' schema not found", dbMeta.Name, tblMeta.Name)
```
gozssky, line 517:
```suggestion
				return errors.Errorf("table `%s`.`%s` schema not found", dbMeta.Name, tblMeta.Name)
```

gozssky (approved):

kennytm (approved):

kennytm:
actually maybe add a unit test?

dveeden:
> actually maybe add a unit test?

@kennytm Yes, that's what I said before: https://github.com/pingcap/br/pull/1397#issuecomment-888891459

I've spend some more time on figuring out the unit tests for lightning and I think I have a reasonable test now.

kennytm (approved):

PR #1401, opened 2021-08-03
lightning: fix disk-quota for local backend

glorv:
Disk-quota may lead to data loss with local backend.

Do a flush before unsafe import for local backend.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### Release note

 - No release note

### DIFF
--- a/pkg/lightning/restore/restore.go
+++ b/pkg/lightning/restore/restore.go
@@ -1739,7 +1739,10 @@ func (rc *Controller) enforceDiskQuota(ctx context.Context) {
 			task := logger.Begin(zap.WarnLevel, "importing large engines for disk quota")
 			var importErr error
 			for _, engine := range largeEngines {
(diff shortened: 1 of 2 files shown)


pkg/lightning/restore/restore.go
```diff
@@ -1739,7 +1739,10 @@ func (rc *Controller) enforceDiskQuota(ctx context.Context) {
 			task := logger.Begin(zap.WarnLevel, "importing large engines for disk quota")
 			var importErr error
 			for _, engine := range largeEngines {
-				if err := rc.backend.UnsafeImportAndReset(ctx, engine); err != nil {
+				var err error
+				if err = rc.backend.FlushEngine(ctx, engine); err != nil {
```
gozssky, line 1743: Why we need to flush engine again after `FlushAll` engines?

PR #1413, opened 2021-08-04
lightning: fix the bug that calculate unfinished ranges may miss some range

### LINKED ISSUE #1403: tidb lightning encountered error "error="checksum mismatched remote vs local

1. What did you do?
Import sql data

checksum sucessfully

3. What did you see instead?
image

tidb-V4.0.12

5. Operation logs
   - Please upload `br.log` for BR if possible
   - Please upload `tidb-lightning.log` for TiDB-Lightning if possible
   - Please upload `tikv-importer.log` from TiKV-Importer if possible
   - Other interesting logs

6. Configuration of the cluster and the task
   - `tidb-lightning.toml` for TiDB-Lightning if possible
   - `tikv-importer.toml` for TiKV-Importer if possible
   - `topology.yml` if deployed by TiUP

7. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus if possible

glorv:
Fix the bug that lightning local backend may encounter data loss.

Close #1403 

- Avoid change the global ranges when calculate unfinished ranges
- Allow `writeAndIngestByRanges` to fail fast to avoid waiting some much time when met unrecoverable failure.

### Check List 

Tests 

 - Unit test
 - Integration test
   - Use old and new cold to run with a test dataset about 10GB with 50GB disk capacity for sort-kv-dir. Before data loss always occur, and disappeared after this change. 
 - No code

Code changes

Side effects

Related changes

### Release note

 - Fix the bug that lightning local backend may met checksum mismatch error due to data loss.


pkg/lightning/backend/local/local.go
```diff
@@ -1878,13 +1878,20 @@ func (local *local) writeAndIngestByRanges(ctx context.Context, engineFile *File
...
+	for i, r := range ranges {
 		startKey := r.start
 		endKey := r.end
 		w := local.rangeConcurrency.Apply()
+		// if meet error here, skip try more here to allow fail fast.
+		if metErr.Load() {
+			wg.Add(i - len(ranges))
+			local.rangeConcurrency.Recycle(w)
+			break
+		}
```
gozssky, line 1894: How about add `wg.Add(1)` before each goroutine start and remove `wg.Add(len(ranges))` and `wg.Add(i - len(ranges))`?

gozssky:
rest LGTM

pkg/lightning/backend/local/local.go
```diff
@@ -1878,13 +1878,20 @@ func (local *local) writeAndIngestByRanges(ctx context.Context, engineFile *File
...
 
 	wg.Add(len(ranges))
 
-	for _, r := range ranges {
+	for i, r := range ranges {
 		startKey := r.start
 		endKey := r.end
 		w := local.rangeConcurrency.Apply()
+		// if meet error here, skip try more here to allow fail fast.
+		if metErr.Load() {
```
kennytm, line 1890: this feels like something that can be handled by `errgroup` 🤔

glorv: There are some difference, here we don't want to terminate the other running tasks if one task fails

pkg/lightning/backend/local/local.go
```diff
@@ -2151,11 +2161,15 @@ func sortAndMergeRanges(ranges []Range) []Range {
...
-func filterOverlapRange(ranges []Range, finishedRanges []Range) []Range {
+func filterOverlapRange(rawRanges []Range, finishedRanges []Range) []Range {
 	if len(finishedRanges) == 0 {
-		return ranges
+		return rawRanges
 	}
 
+	// copy the ranges because we may change its values
+	ranges := make([]Range, 0, len(rawRanges))
+	ranges = append(ranges, rawRanges...)
```
lance6716, line 2171: maybe can use built-in copy function

pkg/lightning/backend/local/local.go
Little-Wallace: Agree

pkg/lightning/backend/local/local.go
```diff
@@ -2152,41 +2160,52 @@ func sortAndMergeRanges(ranges []Range) []Range {
 }
 
 func filterOverlapRange(ranges []Range, finishedRanges []Range) []Range {
-	if len(finishedRanges) == 0 {
+	if len(ranges) == 0 || len(finishedRanges) == 0 {
```
gozssky, line 2163: The current implementation should be correct. But it took me a long time to understand the logic 😂. 
I think these logic may  be simplified. I wrote a demo like this(not tested😂):
```go
func filterOverlapRange(ranges []Range, finishedRanges []Range) []Range {
	var result []Range
	for _, r := range ranges {
		start, end := r.Start, r.End
		for bytes.Compare(start, end) < 0 && len(finishedRanges) > 0 && bytes.Compare(finishedRanges[0].Start, end) < 0 {
			fr := finishedRanges[0]
			// No overlap, skip it.
			if bytes.Compare(fr.End, start) <= 0 {
				finishedRanges = finishedRanges[1:]
				continue
			}
			if bytes.Compare(start, fr.Start) < 0 {
				result = append(result, Range{start, fr.Start})
			}
			start = fr.End
			finishedRanges = finishedRanges[1:]
		}
		if bytes.Compare(start, end) < 0 {
			result = append(result, Range{start, end})
		}
	}
	return result
}
```

Little-Wallace: +1

pkg/lightning/backend/local/local.go
```diff
@@ -2152,43 +2160,31 @@ func sortAndMergeRanges(ranges []Range) []Range {
...
 			}
-		} else if bytes.Compare(ranges[rIdx].end, finishedRanges[fIdx].end) > 0 {
-			ranges[rIdx].start = finishedRanges[fIdx].end
-			fIdx++
-		} else {
-			rIdx++
+			if bytes.Compare(fr.end, start) > 0 {
+				start = fr.end
+			}
+			if bytes.Compare(fr.end, end) > 0 {
```
gozssky, line 2179:
```suggestion
			if bytes.Compare(fr.end, end) >= 0 {
```

pkg/lightning/backend/local/local.go
```diff
@@ -2152,43 +2160,31 @@ func sortAndMergeRanges(ranges []Range) []Range {
...
+				break
+			}
+			finishedRanges = finishedRanges[1:]
+		}
+		if bytes.Compare(start, end) < 0 {
+			result = append(result, Range{start: start, end: r.end})
```
gozssky, line 2185:
```suggestion
			result = append(result, Range{start: start, end: end})
```

gozssky (approved):

pkg/lightning/backend/local/local.go
glorv: There is no difference between > and >= in this context

gozssky: Yes. For `>=`, we can exit the loop early when `fr.end == end`.

3pointer (approved):

PR #1414, opened 2021-08-05
lightning: better estimate of the max-open-files

glorv:
Make better estimate of the max-open-files by also take import routines into account.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note

 - No release note


pkg/lightning/backend/local/local_unix.go
```diff
@@ -39,8 +39,8 @@ func GetSystemRLimit() (Rlim_t, error) {
 // In Local-backend, we need to read and write a lot of L0 SST files, so we need
 // to check system max open files limit.
 func VerifyRLimit(estimateMaxFiles Rlim_t) error {
-	if estimateMaxFiles < minRLimit {
-		estimateMaxFiles = minRLimit
+	if estimateMaxFiles > maxRLimit {
+		estimateMaxFiles = maxRLimit
```
kennytm, line 43: the "1024" introduced in pingcap/tidb-lightning#501 is no longer useful?

glorv: Yes, `minRLimit` is to avoid the minlimit calculated too small, but it seems to be not very useful

kennytm (approved):

kennytm:
merged into pingcap/tidb#27597 already.

PR #1429, opened 2021-09-01
[cherry-pick] restore: split & scatter regions concurrently(tidb#27034)

YuJuncen:
cherry-picking tidb#27034

=== 

Port of https://github.com/pingcap/br/pull/1363

Before, when restoring many small tables, the batcher would probably send small batch due to the so called `AutoCommit` feature of the batcher. By this, we can make the split & scatter & restore worker more active.

But frequently send small batches isn't free. The split step is costly and I/O bounded for even small batches. For example, it costs about 3s to splitting 60 ranges, but restore those ranges typically costs only 1s. Then the restore worker get idle at most time. The restore hence has slowed down.

BR pipeline-2

Instead of using a single split worker, this PR allow multi restore batches be split concurrently.
We added two **hidden** flags, `--batch-flush-interval` and `--pd-concurrency`, the former for better tuning the behavior of batcher, the latter for tweaking the concurrent split.
Also, more logs were added so the create table speed, download, ingest time cost can be observed via log.

### Check List 

Tests 

 - Integration test
 A internal test shows, in a 190GB, 6000 tables workload, this PR can speed up the restoration: the original version takes over **2 hours**  for restoring, and this version takes about **30mins** for restoring. The latter is nearly equal to the time cost of creating tables(see figure below).

image

### Release note

```release-note
Restore many small tables would be faster now.
```

### DIFF
--- a/pkg/logutil/logging.go
+++ b/pkg/logutil/logging.go
@@ -115,7 +115,7 @@ func (region zapMarshalRegionMarshaler) MarshalLogObject(enc zapcore.ObjectEncod
 	for _, peer := range region.GetPeers() {
 		peers = append(peers, peer.String())
 	}
-	enc.AddUint64("ID", region.Id)
+	enc.AddUint64("ID", region.GetId())
 	enc.AddString("startKey", redact.Key(region.GetStartKey()))
 	enc.AddString("endKey", redact.Key(region.GetEndKey()))
 	enc.AddString("epoch", region.GetRegionEpoch().String())
--- a/pkg/restore/import.go
+++ b/pkg/restore/import.go
@@ -261,6 +261,8 @@ func (importer *FileImporter) Import(
 							logutil.Region(info.Region),
 							logutil.Key("startKey", startKey),
 							logutil.Key("endKey", endKey),
+							logutil.Key("file-simple-start", files[0].StartKey),
+							logutil.Key("file-simple-end", files[0].EndKey),
 							logutil.ShortError(e))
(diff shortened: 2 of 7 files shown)


kennytm (approved):

YuJuncen:
pkg/restore/pipeline_items.go:319:27: b.tableWaiters.LoadAndDelete undefined (type *sync.Map has no field or method LoadAndDelete)
OOPS.

kennytm:
LoadAndDelete was added in 1.15 😿

YuJuncen:
```diff
@@ -316,12 +316,13 @@ func (b *tikvSender) registerTableIsRestoring(ts []CreatedTable) func() {
 // till all tables provided are no more ‘current restoring’.
 func (b *tikvSender) waitTablesDone(ts []CreatedTable) {
        for _, t := range ts {
-               wg, ok := b.tableWaiters.LoadAndDelete(t.Table.ID)
+               wg, ok := b.tableWaiters.Load(t.Table.ID)
                if !ok {
                        log.Panic("bug! table done before register!",
                                zap.Any("wait-table-map", b.tableWaiters),
                                zap.Stringer("table", t.Table.Name))
                }
+               b.tableWaiters.Delete(t.Table.ID)
                wg.(*sync.WaitGroup).Wait()
        }
```

Well, anyway, @kennytm  PTAL

kennytm (approved):

zwj-coder (approved):

PR #1480, opened 2022-12-13
br: Ignore ddl jobs with empty query or blacklist type when exec restore

WangLe1321:
Signed-off-by: WangLe1321 <wangle1321@163.com>

Issue Number: #33322

cherry-pick https://github.com/pingcap/tidb/pull/33384

Problem Summary: Sometimes, query of ddl job can be empty, which will cause error when execute incremental restore.

During backup, when we find ddl job with action types about TiFlash or lock/unlock table, we will set the query empty. So we add an action type blacklist, ddl job with action type in blacklist will be ignored. We also ignore ddl job with empty query in case other unexpected situation.

This means that we will ignore ddl jobs about TiFlash or lock/unlock table and ddl jobs with empty query, so they won't cause error by mistake.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note

```release-note
Fix a bug that BR incremental restore return error by mistake caused by ddl job with empty query.
```

### DIFF
--- a/pkg/restore/db.go
+++ b/pkg/restore/db.go
@@ -24,6 +24,20 @@ type DB struct {
 	se glue.Session
 }
 
+type UniqueTableName struct {
+	DB    string
+	Table string
+}
+
+type DDLJobFilterRule func(ddlJob *model.Job) bool
+
+var incrementalRestoreActionBlockList = map[model.ActionType]struct{}{
+	model.ActionSetTiFlashReplica:          {},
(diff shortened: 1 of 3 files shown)


joccau:
cherry-pick https://github.com/pingcap/tidb/pull/33384

joccau (approved):
LGTM

YuJuncen (approved):

